PR #38, opened 2022-12-09
add support for middleware in src dir

speier:
Minor change to support middleware in `src` directory: https://nextjs.org/docs/advanced-features/src-directory

Otherwise the build is failing with:
```
ERROR: Could not map all functions to an entry in the manifest
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -369,9 +369,9 @@ const transform = async ({
   const middlewareEntries = Object.values(middlewareManifest.middleware);
   const functionsEntries = Object.values(middlewareManifest.functions);
   for (const [name, filepath] of functionsMap) {
-    if (name === "middleware" && middlewareEntries.length > 0) {
+    if (middlewareEntries.length > 0 && (name === "middleware" || name === "src/middleware")) {
       for (const entry of middlewareEntries) {
-        if ("middleware" === entry?.name) {
+        if (entry?.name === "middleware" || entry?.name === "src/middleware") {
           hydratedMiddleware.set(name, { matchers: entry.matchers, filepath });
         }
       }
--- a/.changeset/smooth-nails-smash.md
+++ b/.changeset/smooth-nails-smash.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+add support for middleware in src dir


mrluc:
This has been sitting for quite a while; @GregBrimble what's your opinion on this one?

I think NextJS would consider this a [bug](https://github.com/vercel/next.js/pull/37428) and they say [this](https://nextjs.org/docs/advanced-features/src-directory) about the src directory:

> The src directory is very common in many apps and Next.js supports it by default.

src/index.ts
```diff
@@ -369,9 +369,9 @@ const transform = async ({
   const middlewareEntries = Object.values(middlewareManifest.middleware);
   const functionsEntries = Object.values(middlewareManifest.functions);
   for (const [name, filepath] of functionsMap) {
-    if (name === "middleware" && middlewareEntries.length > 0) {
+    if (middlewareEntries.length > 0 && (name === "middleware" || name === "src/middleware")) {
       for (const entry of middlewareEntries) {
-        if ("middleware" === entry?.name) {
+        if (entry?.name === "middleware" || entry?.name === "src/middleware") {
```
dario-piotrowicz, line 374: this feels slightly clunky, could you maybe create a utility function? 🙏 

for examples:
```ts
function isMiddlewareDir(dirName: string): boolean {
  return /^(?:src\/)?middleware$/.test(dirName);
}
```

so that then the code can be updated to something like:
```ts
    if (middlewareEntries.length > 0 && isMiddlewareDir(name)) {
      for (const entry of middlewareEntries) {
        if (entry?.name && isMiddlewareDir(entry.name)) {
```

what do you think? 🙂

dario-piotrowicz (approved):

dario-piotrowicz:
@speier thanks a lot for the PR 😃, sorry it took a while for it to get reviewed 😢🙇 , looks good to me, I just left a suggestion on how I think the code could be improved please have a look 🙂 

Also could you please create a changeset for your change? 🙏  (by running `npx changeset`)

dario-piotrowicz:
@mrluc you're right, sorry for keeping the PR stale, Greg's been rather busy and hasn't really had much capacity to review PRs here lately, we're now making an effort to reinvigorate the project and keeping it more active 💪

src/index.ts
speier: sure, a utility function would make sense. i just didn't wanted to introduce bigger changes initially, hence i choose to simply add it in place..

dario-piotrowicz: ah I see, thanks for trying to reduce the change 🙂, I'm pretty sure different people will have different opinions on this (especially since it's a small change) but yeah to me a utility function here helps in making the code (slightly) more dry and more readable, so definitely a plus from my point of view 😃 

(I'm am planning to also slit the index.ts file in multiple files at some point so adding utilities should be nicer in the future since there would be more structure to the code)

GregBrimble, line 374:
```suggestion
        if (["middleware", "src/middleware"].includes(entry?.name)) {
```

GregBrimble (approved):
Left a suggestion, but take-it-or-leave-it :)

src/index.ts
speier: i would do a refactoring separated from this pr. when you start splitting index.ts into smaller chunks anyway you would have a better overview on which functions needs to be extracted and where.

speier:
> @speier thanks a lot for the PR 😃, sorry it took a while for it to get reviewed 😢🙇 , looks good to me, I just left a suggestion on how I think the code could be improved please have a look 🙂
> 
> Also could you please create a changeset for your change? 🙏 (by running `npx changeset`)

running `npx changeset` gives an error. maybe it's introduced after the fork i was working on?

dario-piotrowicz:
> > @speier thanks a lot for the PR 😃, sorry it took a while for it to get reviewed 😢🙇 , looks good to me, I just left a suggestion on how I think the code could be improved please have a look 🙂
> > Also could you please create a changeset for your change? 🙏 (by running `npx changeset`)
> 
> running `npx changeset` gives an error. maybe it's introduced after the fork i was working on?

@speier I've added the changeset for you, I hope you don't mind 🙂 🙏

dario-piotrowicz:
@speier I was about testing middlewares locally before merging your PR, but unfortunately I could not get them to work not even on /middleware.ts, something on the vercel side could be breaking things 😓 

I assume the middlewares work locally for you? if so could you share what version of next and the vercel cli you're using?

dario-piotrowicz:
Thanks a bunch @speier, checked locally and it is all good 🙂:+1:

Issue #42, opened 2022-12-16
"ERROR: Could not map all functions to an entry in the manifest." when adding basePath to next.config.js


kendall-wahnschaffe:
### Out of the box, everything works
✅ Works as expected
Following steps 1 - 4 from the (README)[https://github.com/cloudflare/next-on-pages#quick-start]

✅ Works as expected
`npx @cloudflare/next-on-pages --experimental-minify`

✅ Deploys as expected
`npx wrangler pages publish .vercel/output/static/`

### However, with basePath
Add `basePath` to `next.config.js`

```js
const nextConfig = {
  basePath: "/base-path",
  experimental: {
    runtime: "experimental-edge",
  },
  reactStrictMode: true,
};
```
❌ Does not work.
`npx @cloudflare/next-on-pages --experimental-minify`

Build completes, but logs the following ERROR:

```ERROR: Could not map all functions to an entry in the manifest.```

Furthermore, if I proceed to deploy the application, I get a browser HTTP ERROR 404 at `/base-path` and the application's 404 page at `/base-path/`

hanford:
This happens to me as well. I rely on `basePath` pretty heavily in order to have many Next.js applications on the same hostname.

I've created a repro over here: https://github.com/hanford/next-on-pages-42 and can also be messaged on the Cloudflare Discord as @hanford if any additional information is helpful

hanford:
I believe https://github.com/cloudflare/next-on-pages/pull/44 should add proper support for `basePath` on `next-on-pages`. I've got the changes in the PR available in `.patch` that can be applied using something like [patch-package](https://www.npmjs.com/package/patch-package)

The patch file is here: https://gist.github.com/hanford/9308f35b5d12a3c225a0045cf6bff3c8

(unrelated, also in the patch, I've disabled the `prepInstall` function because I've been using Pages direct upload and `next-on-pages` doesn't yet support many package managers (https://github.com/cloudflare/next-on-pages/pull/11), if desired I can create a new patch without that change..

rationalthug:
Will this fix be accepted in the near future? I utilize basePath extensively as well and am using Pages' GitHub integration, which causes the build to fail with this same error.

PR #44, opened 2022-12-22
Implement support for `next.config#basePath`

hanford:
The [basepath](https://nextjs.org/docs/api-reference/next.config.js/basepath) feature in Next.js allows applications to be hosted specific paths on the same domain like the following:

example.com/user/account <-- Next.js App 1
example.com/store/checkout <-- Next.js App 2

---

Currently `@cloudflare/next-on-pages` errors when using the basepath option, this PR aims at adding support by reading the basePath value from the build output, `.next/routes-manifest.json`, and then inlines the value into the Worker. 

Upon receiving a request, the worker will strip the `basePath` value from the pathname, and then run the matching logic with the stripped pathname. This appears to play nicely with everything else inside of Next's `base-server#handleRequest`, though I'm happy to move things around if there is a better way..?

Also happy to chat real time in the Cloudflare Discord, with my username being @hanford. 

---

closes https://github.com/cloudflare/next-on-pages/issues/42

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -133,7 +133,52 @@ const transform = async ({
     exit(1);
   }
 
-  const functionsDir = resolve(".vercel/output/functions");
+  interface RoutesManifest {
+    version: 3;
+    pages404: boolean;
+    basePath: string;
+    // TODO type these?
+    redirects: unknown;
+    headers: unknown;
+    dynamicRoutes: unknown;
+    staticRoutes: unknown;
+    dataRoutes: unknown;
+    rsc: unknown;
+    rewrites: unknown;
+  }
+
+  let routesManifest: RoutesManifest;
+
+  try {
+    routesManifest = JSON.parse(
+      await readFile(".next/routes-manifest.json", "utf8")
+    );
+  } catch {
+    console.error(
+      "⚡️ ERROR: Could not read ./next/routes-manifest.json files"
+    );
+    console.error(
+      "⚡️ Please report this at https://github.com/cloudflare/next-on-pages/issues."
+    );
+    exit(1);
+  }
+
+  if (routesManifest.version !== 3) {
+    console.error(
+      `⚡️ ERROR: Unknown functions manifest version. Expected 3 but found ${routesManifest.version}.`
+    );
+    console.error(
+      "⚡️ Please report this at https://github.com/cloudflare/next-on-pages/issues."
+    );
+    exit(1);
+  }
+
+  const basePath = routesManifest.basePath ?? "";
+  if (basePath !== "") {
+    console.log("⚡️ Using basePath ", basePath);
+  }
+
+  const functionsDir = resolve(".vercel/output/functions" + basePath);
   let functionsExist = false;
   try {
     await stat(functionsDir);
@@ -455,6 +500,7 @@ const transform = async ({
     platform: "neutral",
     define: {
       __CONFIG__: JSON.stringify(config),
+      __BASE_PATH__: JSON.stringify(basePath),
     },
     outfile: ".vercel/output/static/_worker.js",
   });
--- a/src/index.ts
+++ b/src/index.ts
@@ -133,17 +133,37 @@ const transform = async ({
     exit(1);
   }
 
+  // RoutesManifest.version and RoutesManifest.basePath are the only fields accessed
   interface RoutesManifest {
     version: 3;
-    pages404: boolean;
     basePath: string;
-    // TODO type these?
-    redirects: unknown;
+    pages404: boolean;
+    redirects: {
+      source: string;
+      destination: string;
+      basePath: boolean | undefined;
+      internal: boolean;
+      statusCode: number;
+      regex: string;
+    }[];
+    dynamicRoutes: {
+      page: string;
+      regex: string;
+      routeKeys: any; // object of dynamic parameters;
+      namedRegex: string;
+    };
+    staticRoutes: {
+      page: string;
+      regex: string;
+      routeKeys: any; // object of dynamic parameters;
+      namedRegex: string;
+    }[];
+    rsc: {
+      header: string;
+      varyHeader: string;
+    };
     headers: unknown;
-    dynamicRoutes: unknown;
-    staticRoutes: unknown;
     dataRoutes: unknown;
-    rsc: unknown;
     rewrites: unknown;
   }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -170,7 +170,7 @@ const transform = async ({
     console.log("⚡️ Using basePath ", basePath);
   }
 
-  const functionsDir = resolve(".vercel/output/functions" + basePath);
+  const functionsDir = resolve(`.vercel/output/functions${basePath}`);
   let functionsExist = false;
   try {
     await stat(functionsDir);
--- a/src/index.ts
+++ b/src/index.ts
@@ -137,34 +137,6 @@ const transform = async ({
   interface RoutesManifest {
     version: 3;
     basePath: string;
-    pages404: boolean;
-    redirects: {
-      source: string;
-      destination: string;
-      basePath: boolean | undefined;
-      internal: boolean;
-      statusCode: number;
-      regex: string;
-    }[];
-    dynamicRoutes: {
-      page: string;
-      regex: string;
-      routeKeys: any; // object of dynamic parameters;
-      namedRegex: string;
-    };
-    staticRoutes: {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -160,7 +160,52 @@ const transform = async ({
     exit(1);
   }
 
-  const functionsDir = resolve(".vercel/output/functions");
+  interface RoutesManifest {
+    version: 3;
+    pages404: boolean;
+    basePath: string;
+    // TODO type these?
+    redirects: unknown;
```
hanford, line 168: I can probably add some better types here...

hanford:
This approach is working well for me in some private repos. I have an example application using `basePath` over [here](https://github.com/hanford/next-on-pages-42), and am utilizing these changes in patch over [here](https://github.com/hanford/next-on-pages-42/tree/wip).

src/index.ts
```diff
@@ -160,7 +160,72 @@ const transform = async ({
...
+      routeKeys: any; // object of dynamic parameters;
+      namedRegex: string;
+    }[];
+    rsc: {
+      header: string;
+      varyHeader: string;
+    };
+    headers: unknown;
+    dataRoutes: unknown;
+    rewrites: unknown;
```
hanford, line 194: These types could use some work if requested, though may not be necessary as they're not used in this package at all...

dario-piotrowicz:
@hanford I rebased your branch just so that the prerelease job could kick in, I hope you don't mind! 🙏 

by the way, could you please create a changeset? (via `npx changeset`) 🙏

src/index.ts
```diff
@@ -160,7 +160,72 @@ const transform = async ({
     exit(1);
   }
 
-  const functionsDir = resolve(".vercel/output/functions");
+  // RoutesManifest.version and RoutesManifest.basePath are the only fields accessed
```
dario-piotrowicz, line 163: I am not a big fan of including this if it's not really needed, could you instead just set the version and basepath?

also the `3` is used here and also below, I think that setting up a constant and using that instead would make the code more robust and clear.

Basically I am thinking of something along these lines:
Screenshot 2023-03-10 at 21 13 55

What do you think? 🙂 

(PS: I am not sure if I got the right url there but you get the gist 🙂)

src/index.ts
```diff
@@ -160,7 +160,72 @@ const transform = async ({
...
+    );
+    exit(1);
+  }
+
+  const basePath = routesManifest.basePath ?? "";
+  if (basePath !== "") {
+    console.log("⚡️ Using basePath ", basePath);
+  }
+
+  const functionsDir = resolve(".vercel/output/functions" + basePath);
```
dario-piotrowicz, line 201:
```suggestion
  const functionsDir = resolve(`.vercel/output/functions${basePath}`);
```

src/index.ts
```diff
@@ -482,6 +547,7 @@ const transform = async ({
     platform: "neutral",
     define: {
       __CONFIG__: JSON.stringify(config),
+      __BASE_PATH__: JSON.stringify(basePath),
```
dario-piotrowicz, line 523: is the `JSON.stringify` needed?

templates/_worker.js/index.ts
```diff
@@ -147,9 +149,19 @@ export default {
...
+              // `/docs/` -> `/`
+              // `/docs/_next/static/main.js` -> `/_next/static/main.js`
+              pathname.replace(__BASE_PATH__, "/").replace("//", "/")
+            : pathname;
+
           if (
-            pathname.match(regexp) ||
-            `${pathname}/page`.replace("//page", "/page").match(regexp)
+            nextPathname.match(regexp) ||
+            `${nextPathname}/page`.replace("//page", "/page").match(regexp)
```
dario-piotrowicz, line 164: is this replace still needed since you're squashing `//` into `/` already?

templates/_worker.js/index.ts
```diff
@@ -147,9 +149,19 @@ export default {
...
         if (matcher.regexp) {
           const regexp = new RegExp(matcher?.regexp);
+          const nextPathname = __BASE_PATH__
+            ? // Remove basePath from URL, also squish `//` into `/`
+              // If the baseUrl is set to "/docs" the following will happen:
+              // `/docs/profile/settings` -> `/profile/settings`
+              // `/docs` -> `/`
+              // `/docs/` -> `/`
+              // `/docs/_next/static/main.js` -> `/_next/static/main.js`
+              pathname.replace(__BASE_PATH__, "/").replace("//", "/")
```
dario-piotrowicz, line 159: what do you think? shall we also check that __BASE_PATH__ is at the start of the string? (I think it wouldn't really make a difference but I figured I would just ask 🙂)

src/index.ts
hanford: I'm not a 100% certain. Happy to change it if ESBuild won't inline the string literal as a literal

dario-piotrowicz (approved):
@hanford This is awesome! thank you so very much! 😃 

Also your [repro](https://github.com/hanford/next-on-pages-42) really made reviewing this very straightforward! You rock! 💪 

(PR: sorry it took quite a while for the PR to get reviewed)

src/index.ts
dario-piotrowicz: I must say that I don't really get it... 😕 

but it looks like that in other places we also run `JSON.stringify` on a string:
Screenshot 2023-03-10 at 21 42 44

So I think that this is fine (and consistent) and we can revisit this later 🙂👍

templates/_worker.js/index.ts
hanford: Happy change the logic to "If the pathname startsWith basePath". I think it'll be a tiny bit safer

templates/_worker.js/index.ts
hanford: I think this is still needed in the case that bathPath isn't used

hanford:
@dario-piotrowicz appreciate the 👀, let me now if you'd like me to change/update anything else 🥂

templates/_worker.js/index.ts
dario-piotrowicz: ah yeah sorry, it is, my bad 😅

src/index.ts
james-elicx: > but it looks like that in other places we also run `JSON.stringify` on a string

Just to add, the reason for the JSON.stringify on chunkFilePath is so that the raw value is `"'...'"` instead of `"..."` for the AST. You can see this demonstrated if you put in some code that uses require on astexplorer.net.

dario-piotrowicz:
> @dario-piotrowicz appreciate the 👀, let me now if you'd like me to change/update anything else 🥂

All good 😃 (you did leave the magic number `3` in, it's no big deal if you prefer it that way 🙂)

I'm just checking with the team and should hopefully be able to merge the PR soon 😄💪 
(we're in a phase where a few new people like me have been added to the repo, so we agreed to tread lightly and checking with Greg, the repo's creator, before merges for the time being 🙂)

src/index.ts
dario-piotrowicz: ah yeah I see, sorry yeah it slipped my mind that JSON.stringify adds quotes! thanks! 🙂

hanford:
Awesome, no rush. I figure keeping the `routesManifest.version` in there may be useful incase they (Next) totally overhaul the structure of the routesManifest or something

dario-piotrowicz:
> Awesome, no rush. I figure keeping the routesManifest.version in there may be useful incase they (Next) totally overhaul the structure of the routesManifest or something. Happy to remove that (and the conditional) if you feel strongly!

nono, it's all good thanks 🙂👍 

Greg said he'll have one last double check here early next week and then we can merge! 😃

hanford:
That's interesting, as it is today, you can't create `.vercel/out` without a `.next` directory... this library is also already reading from the `.next`

Maybe I could tweak this PR to read from the `next.config.js`.. but that's not a json file which leads to different complexities.. (Can't just require it because the file is optional).

As far as I'm aware what is in this PR is actually the cleanest approach.

dario-piotrowicz:
> I'm not able to use this library at all because my Next.js app's are too big (even with the minification fixes), so I'm a lot less inclined to go back to drawing board on this PR.

I'm sorry to hear that 😓, yeah that's a known limitation that we ought to look into at some point (feel free to open an issue for it if you want, it would be good to track that)

hanford:
Can you check with the team as to whether or not reading the `next.config.js` is a reasonable fix?

PR #50, opened 2023-01-03
Avoid invoking worker for static files

Sebmaster:
The files in _next/static are guaranteed to be static, so we can avoid invoking the worker for these.

This is a very, very minimal solution for #5 which should (eventually) be extended to walking the static assets to pre-register them as well.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,6 +24,39 @@ type LooseNode = Node & {
   value: any;
 };
 
+const generateHeaders = async () => {
+  await writeFile(
+    ".vercel/output/static/_headers",
+    `\n/_next/static/*\n  Cache-Control: public,max-age=31536000,immutable\n`,
+    {
+      // in case someone configured redirects already, append to the end
+      flag: "a",
+    }
+  );
+};
+
+const generateRoutes = async () => {
+  try {
+    await writeFile(
+      ".vercel/output/static/_routes.json",
+      JSON.stringify({
+        version: 1,
+        include: ["/*"],
+        exclude: ["/_next/static/*"],
+      }),
+      { flag: "ax" } // don't generate file if it's already manually maintained
+    );
+  } catch (e) {
+    if (e.code != "EEXIST") {
+      throw e;
+    }
+  }
+};
+
+const generateMetadataFromConfig = async () => {
+  await Promise.all([generateHeaders(), generateRoutes()]);
+};
+
 const prepVercel = async () => {
   try {
     await stat(".vercel/project.json");
@@ -528,7 +561,10 @@ const main = async ({
     await buildVercel();
   }
 
-  await transform({ experimentalMinify });
+  await Promise.all([
+    transform({ experimentalMinify }),
+    generateMetadataFromConfig(),
+  ]);
 };
 
 (async () => {


src/index.ts
```diff
@@ -24,6 +24,39 @@ type LooseNode = Node & {
...
+      { flag: "ax" } // don't generate file if it's already manually maintained
+    );
+  } catch (e) {
+    if (e.code != "EEXIST") {
+      throw e;
+    }
+  }
+};
+
+const generateMetadataFromConfig = async () => {
```
dario-piotrowicz, line 56: I'm not sure why its `FromConfig` since it just generated the data without any configuration?

src/index.ts
```diff
@@ -528,7 +561,10 @@ const main = async ({
     await buildVercel();
   }
 
-  await transform({ experimentalMinify });
+  await Promise.all([
+    transform({ experimentalMinify }),
+    generateMetadataFromConfig(),
```
dario-piotrowicz, line 566: Cloud you move the `generateMetadataFromConfig` function (alongside the other function) closer to here? (for example above `main`)

I think it would make things clearer to define the function(/s) closer to where it is used.

(PS: we need to split this file into multiple files at some point but that's a slightly lesser priority so for now at least if you could move the functions that's be great 🙏 )

dario-piotrowicz:
@Sebmaster thanks a lot for the PR, it looks good I'm going to also double check it locally 🙂👍

Could you please create a changeset? (`npx changeset`) 🙏

dario-piotrowicz (approved):

src/index.ts
```diff
@@ -24,6 +24,39 @@ type LooseNode = Node & {
   value: any;
 };
 
+const generateHeaders = async () => {
+  await writeFile(
+    ".vercel/output/static/_headers",
+    `\n/_next/static/*\n  Cache-Control: public,max-age=31536000,immutable\n`,
```
GregBrimble, line 30: Two things on this `_headers` generation:

1. `.vercel/output/config.json` should include a route rule which covers this. For example, in a build I just produced, I find the following:

    ```json
    {
      "routes": [
        {
          "src": "/_next/static/(?:[^/]+/pages|pages|chunks|runtime|css|image|media|3xPbVjttysUBv9VjNExZE)/.+",
          "headers": {
            "cache-control": "public,max-age=31536000,immutable"
          },
          "continue": true,
          "important": true
        }
      ]
    }
    ```

    It would be great to pull from here and just automatically apply all the headers for all the rules that this gives us. Perhaps this is best tackled in a separate PR? I won't personally block on this though, so we all want to get the current implementation through on this current PR, it's okay with me, but I'd at least like to create an issue to track the follow-up of implementing this from `.vercel/output/config.json` instead.

1. [SvelteKit also produce a `_headers` file](https://github.com/sveltejs/kit/blob/c45c35e7b40837cd1bea3fc26cfb831478e03c00/packages/adapter-cloudflare/index.js#L133). They add a nice comment wrapping the autogenerated rules. It would be nice to do the same here for any human that comes along and inspects it.

src/index.ts
```diff
@@ -24,6 +24,39 @@ type LooseNode = Node & {
...
+    }
+  );
+};
+
+const generateRoutes = async () => {
+  try {
+    await writeFile(
+      ".vercel/output/static/_routes.json",
+      JSON.stringify({
+        version: 1,
```
GregBrimble, line 43: We also accept a `description` property, which [Wrangler](https://github.com/cloudflare/workers-sdk/blob/f61de1dd1b09cb45fefb6e2ace2e615f6a87ef4b/packages/wrangler/src/pages/constants.ts#L18) and [SvelteKit](https://github.com/sveltejs/kit/blob/c45c35e7b40837cd1bea3fc26cfb831478e03c00/packages/adapter-cloudflare/index.js#L126) add a small note to describe who built this file. We could add something like "Built with @cloudflare/next-on-pages@0.0.3."

src/index.ts
```diff
@@ -24,6 +24,39 @@ type LooseNode = Node & {
...
+        include: ["/*"],
+        exclude: ["/_next/static/*"],
```
GregBrimble, line 45: [Contrary to the comment above about `_headers`](https://github.com/cloudflare/next-on-pages/pull/50/files#r1134719235), I think it's fine to exclude everything after `/_next/static/` here :)

src/index.ts
GregBrimble: I wouldn't `await Promise.all()` these. There's no serious performance benefit, they're not conceptually similar, and it would result in intermingled messages if either of these functions log something. Better to just `await` each, one-after-the-other.

GregBrimble (approved):

src/index.ts
Sebmaster: > It would be great to pull from here and just automatically apply all the headers for all the rules that this gives us.

Yeah I was a little wary of that since they limit their paths much more aggressively and I didn't want to expand (potentially) unsafe headers to arbitrary paths. Can change that though if we assume it's safe enough.

src/index.ts
Sebmaster: Ah, that function used to actually use the config, but I dropped it since it wasn't really necessary.

src/index.ts
Sebmaster: Added.

src/index.ts
Sebmaster: Both done as well.

src/index.ts
dario-piotrowicz: @Sebmaster I think it should be safe enough and that does sound like the best way of implementing this

So I think we can either merge this as is and open an issue/followup PR to apply the improvement or change this before merging

If it's quick enough let's do it here otherwise let's open an issue about it

what do you think? could you have a look and let me know which is? 🙂 🙏

PR #51, opened 2023-01-04
Set `x-matched-path` fixing dynamic routing

### LINKED ISSUE #32: Dynamic routing doesn't work.
When I set up dynamic routing and deploy it on cloudflare pages, page fails with 500 error.

my code:

image

hanford:
This PR ensures `x-matched-path` is set, so `Next.js#handleRequest` is able to recognize and render dynamic routes.

It appears Next.js uses this header [pretty extensively](https://github.com/vercel/next.js/blob/cafbe2a0ade6375681224ff79799ececdbd821f2/packages/next/src/server/base-server.ts#L534) in it's base-server, to determine whether or not we're in an edge runtime. Without this header, Next.js doesn't match the URL segments resulting in a "Internal Server Error" page.

I'm happy to change any part of this PR, but think this might be one of the "cleaner" solutions I've been able to come up with.

---
I've also created a reproduction of this bug, along with a patched version of `next-on-pages` that proves that this solution works over here: https://github.com/hanford/next-on-pages-32

closes #32

### DIFF
--- a/templates/_worker.js/index.ts
+++ b/templates/_worker.js/index.ts
@@ -140,7 +140,8 @@ export default {
       }
     }
 
-    for (const { matchers, entrypoint } of Object.values(__FUNCTIONS__)) {
+    for (const [key, value] of Object.entries(__FUNCTIONS__)) {
+      const { matchers, entrypoint } = value;
       let found = false;
       for (const matcher of matchers) {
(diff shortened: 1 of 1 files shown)


hanford:
Hmmmm.. going to close this until I can get it working 100%

PR #55, opened 2023-01-13
doc: update readme, cleanup

shinebayar-g:
- Removed Compatibility Flags, they're enabled by default now.
- Reword Local testing => Local development?
- ~Looks like it's not necessary to use `experimental-edge` flag with NextJS 13 with default Pages setup. Still required with App directory though.~ Match router example with official NextJS docs.
Edit: Nah I'm getting this msg with it. Reverted.
Edit2: Difference between `experimental-edge` and `edge` is asked [here](https://github.com/vercel/next.js/discussions/44871#discussioncomment-4683660)
   ```
   error - Page /src/pages/api/hello provided runtime 'edge', the edge runtime for rendering is currently experimental. Use 
   runtime 'experimental-edge' instead.
   ```
- Mention NodeJS 18 is not supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,42 +17,25 @@ Reference:
 
 1. Configure the project to use the Edge Runtime:
 
-   1. Replace `pages/api/hello.ts` with the following:
+   Replace `pages/api/hello.ts` with the following:
 
-      ```typescript
-      // Next.js Edge API Routes: https://nextjs.org/docs/api-routes/edge-api-routes
-      import type { NextRequest } from 'next/server';
+   ```typescript
+   // Next.js Edge API Routes: https://nextjs.org/docs/api-routes/edge-api-routes
+   import type { NextRequest } from 'next/server';
 
-      export const config = {
-      	runtime: 'experimental-edge',
-      };
+   export const config = {
+   	runtime: 'experimental-edge',
+   };
 
-      export default async function handler(
-      	req: NextRequest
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,42 +17,29 @@ Reference:
 
 1. Configure the project to use the Edge Runtime:
```
dario-piotrowicz, line 18: You've removed the runtime config in the next.config.js and just exporting the config in the hello.ts file, doesn't that make the config apply only when you try to use the hello endpoint? (and the rest of the app still uses the default node runtime)

dario-piotrowicz: I think this is fine as the vp of dx at Vercel suggested that setting the runtime globally is not recommened (https://github.com/cloudflare/cloudflare-docs/pull/7097#issue-1508782427)

dario-piotrowicz (approved):
Thanks a bunch @shinebayar-g :smiley: 

(We also probably need to update `experimental-edge` to just `edge` but I'm planning to improve the readme very soon so I can just do that as part of it :slightly_smiling_face::+1: )

README.md
shinebayar-g: Hey @dario-piotrowicz thanks for the link. Nice catch.

shinebayar-g:
Thanks for the review @dario-piotrowicz. Will be waiting for your README updates. They're super helpful.

Issue #1, opened 2022-10-20
Environment variables and bindings on `env`


GregBrimble:
Next.js has it's own API design for [`getServerSideProps()`](https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#context-parameter) and the Edge API Routes and Middleware take only the `NextRequest` object as its single parameter. This means that passing Workers' `env` object which contains environment variables and bindings isn't immediately obvious.

One idea is to 'pollute' the global space and access things by request. This has the advantage of being _similar_ to how it works today, but might be too far into the uncanny valley.

For example:

```ts
import type { NextRequest } from 'next/server'

export const config = {
  runtime: 'experimental-edge'
}

export default async function(req: NextRequest) {
  const value = process.env[req].MY_ENV_VAR
}
```

Another might be to use the `@cloudflare/next-on-pages` package as a runtime library which is different enough to be intentional, but still fundamentally works the same way:

```ts
import type { NextRequest } from 'next/server'
import { env } from '@cloudflare/next-on-pages'

export const config = {
  runtime: 'experimental-edge'
}

export default async function(req: NextRequest) {
  const value = env(req, 'MY_ENV_VAR')
}
```

What do y'all think? Any preferences? Any other ideas?

bluebeel:
Why not extend NextRequest?
Next ignores the properties added to the object. So we can add what we want to the Request object. Then we just have to provide a type `CfNextRequest = NextRequest & { env: Record<string, unknown> }` via the `@cloudflare/next-on-pages` lib.

```js
export default {
  async fetch(request, env, context) {
    const { pathname } = new URL(request.url);
    const routes = routesMatcher({ request }, __CONFIG__.routes);

   // link env to request object
   request.env = env;

    for (const route of routes) {
      if ("middlewarePath" in route && route.middlewarePath in __MIDDLEWARE__) {
        return await __MIDDLEWARE__[route.middlewarePath].entrypoint.default(
          request,
          context
        );
      }
    }

    for (const { matchers, entrypoint } of Object.values(__FUNCTIONS__)) {
      let found = false;
      for (const matcher of matchers) {
        if (matcher.regexp) {
          if (pathname.match(new RegExp(matcher?.regexp))) {
            found = true;
            break;
          }
        }
      }

      if (found) {
        return entrypoint.default(request, context);
      }
    }

    return env.ASSETS.fetch(request);
  },
} as ExportedHandler<{ ASSETS: Fetcher }>;```

and we have
[...]

PR #58, opened 2023-01-24
spread env into process.env

dario-piotrowicz:
in the _worker.js template take the provided env and spread its values into the global process.env so that bindings can be accessed in the Next.js application

### DIFF
--- a/.changeset/young-coins-appear.md
+++ b/.changeset/young-coins-appear.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": minor
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/young-coins-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": minor
+---
+
+spread env into process.env so that bindings can be accessed in Next.js app
```
GregBrimble, line 5:
```suggestion
Environment variables and bindings are now available on `process.env` from within your SSR'd pages/API handlers!
```

GregBrimble:
Fixes #1

Issue #76, opened 2023-02-21
Use AsyncLocalStorage to support newer versions of `next`


GregBrimble:
`next` has started to rely on AsyncLocalStorage being available in the global scope. We need to add the requirement for the new `nodejs_compat` Compatibility Flag and add this to our Worker template.

GregBrimble:
While we're here, we can also swap out the hack of setting `globalThis.process.env` to the incoming `env` object from the fetch handler and instead use AsyncLocalStorage to do this correctly scoped to the current request.

AndrewIngram:
Should setting the `nodejs_compat` flag in my pages project be enough to make this work? Because I've just added it and it didn't seem to solve the `"Error: Invariant: AsyncLocalStorage accessed in runtime where it is not available"` error I see in the logs.

dario-piotrowicz:
This should have been solved by #115 🙂

PR #80, opened 2023-02-27
Update README with valid instructions on 26 Feb 2023

AaronFriel:
The Next template and version should be pinned to a working version.

The compatibility flags documentation is out of date and causes an error when setting them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,15 @@ Reference:
 
 ## Quick Start
 
-1. `npx create-next-app@latest my-app`
+1. `npx create-next-app@13.1.5 my-app`
 
    Note that if you elect to use eslint, there are a couple of places you need to add return types to make the default template pass the pre-build checks.
 
 1. `cd` into the new directory (e.g. `cd my-app`)
+
+1. `npm install --save next@13.1.5`
+
+   This will pin the version of Next to the latest validated version.
 
 1. `npm install -D @cloudflare/next-on-pages vercel`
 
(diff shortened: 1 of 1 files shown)


dario-piotrowicz (approved):

README.md
```diff
@@ -63,9 +67,7 @@ Reference:
    | Build command          | `npx @cloudflare/next-on-pages --experimental-minify` |
    | Build output directory | `.vercel/output/static`                               |
 
-1. Add a `NODE_VERSION` environment variable set to `14` or greater.
-
-1. In the Pages project **Settings** > **Functions** > **Compatibility Flags**, add the `transformstream_enable_standard_constructor` and `streams_enable_constructors` flags. These will not be necessary once they graduate to be on by default on 2022-11-30's compatibility date.
+1. Add a `NODE_VERSION` environment variable set to `16` or greater.
```
dario-piotrowicz, line 70: Nextjs app directory requires `v16.8.0` (https://beta.nextjs.org/docs/installation#system-requirements)

Should we maybe say 
```
`16.8.0` or greater
```
?

:thinking:

dario-piotrowicz: (PS: node 18 is not yet supported on pages and 17 is getting deprecated, so maybe we want to just be more specific here?)

README.md
```diff
@@ -7,12 +7,16 @@ Reference:
 
 ## Quick Start
 
-1. `npx create-next-app@latest my-app`
+1. `npx create-next-app@13.1.5 my-app`
```
dario-piotrowicz, line 10: Could you put `13.0.0` here and below?

I did try using `13.1.5` but it seems like Next using app directory doesn't work because they started relying on the AsyncLocalStorage (which we don't yet support: https://github.com/cloudflare/next-on-pages/issues/76), so I think for now it's safer to just do `13.0.0`

README.md
dario-piotrowicz: This has already been updated

dario-piotrowicz:
Thanks a lot for the PR @AaronFriel :pray: 

The node version and local command parts of the README have already been updated in https://github.com/cloudflare/next-on-pages/pull/55

Regarding the nextjs pinning I think we need to check things up and decided on what version to apply (the one you're suggesting for example doesn't work if the developer chooses to use the app dir)

(and maybe after https://github.com/cloudflare/next-on-pages/pull/115 we can even bump it up to 13.2.4)

I'm going to look into improving the readme generally in the next few days so I'm closing this PR just not to keep it lingering without merging it

I hope you don't mind :pray: 

please feel free to reopen it or create a new one regarding the version pinning if you feel strongly about it :slightly_smiling_face:

PR #86, opened 2023-03-07
fix: windows file paths not matching middleware manifest

james-elicx:
On Windows, the file paths may use `\\` instead of `/`.

This would result in the error `Could not map all functions to an entry in the manifest.` being thrown.

When we are running through the middleware manifest entries and checking that our function entries exist in it, we are comparing an entry with `/` in it to our path with `\`  in it.

This change replaces the `\\` with `/` so they are in the same format as the middleware manifest when adding to the functions map.

### DIFF
--- a/.changeset/ninety-queens-pump.md
+++ b/.changeset/ninety-queens-pump.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+Fix Windows file paths not matching entries in the middleware manifest.
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@ import { watch } from "chokidar";
 import pLimit from "p-limit";
 import acorn, { parse, Node } from "acorn";
 import { generate } from "astring";
+import { normalizePath } from './utils';
 
 type LooseNode = Node & {
   expression?: LooseNode;
@@ -307,8 +308,8 @@ const transform = async ({
 
           // `\\` needs to be replaced with `/` for Windows so that it can match the path correctly in the middleware manifest.
           functionsMap.set(
-            relative(functionsDir, filepath).slice(0, -".func".length).replace(/\\/g, "/"),
-            newFilePath.replace(/\\/g, "/")
+            normalizePath(relative(functionsDir, filepath).slice(0, -".func".length)),
+            normalizePath(newFilePath)
           );
         } else if (isDirectory) {
           await walk(filepath);
--- a/src/index.ts
+++ b/src/index.ts
@@ -305,9 +305,10 @@ const transform = async ({
           await mkdir(dirname(newFilePath), { recursive: true });
           await writeFile(newFilePath, contents);
 
+          // `\\` needs to be replaced with `\` for Windows so that it can match the path correctly in the middleware manifest.
           functionsMap.set(
-            relative(functionsDir, filepath).slice(0, -".func".length),
-            newFilePath
+            relative(functionsDir, filepath).slice(0, -".func".length).replace(/\\/g, "/"),
+            newFilePath.replace(/\\/g, "/")
           );
         } else if (isDirectory) {
           await walk(filepath);
--- a/src/index.ts
+++ b/src/index.ts
@@ -305,7 +305,7 @@ const transform = async ({
           await mkdir(dirname(newFilePath), { recursive: true });
           await writeFile(newFilePath, contents);
 
-          // `\\` needs to be replaced with `\` for Windows so that it can match the path correctly in the middleware manifest.
+          // `\\` needs to be replaced with `/` for Windows so that it can match the path correctly in the middleware manifest.
           functionsMap.set(
             relative(functionsDir, filepath).slice(0, -".func".length).replace(/\\/g, "/"),
             newFilePath.replace(/\\/g, "/")
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,7 @@ import { watch } from "chokidar";
 import pLimit from "p-limit";
 import acorn, { parse, Node } from "acorn";
 import { generate } from "astring";
-import { normalizePath } from './utils';
+import { normalizePath } from "./utils";
 
 type LooseNode = Node & {
   expression?: LooseNode;
--- a/src/index.ts
+++ b/src/index.ts
@@ -306,7 +306,6 @@ const transform = async ({
           await mkdir(dirname(newFilePath), { recursive: true });
           await writeFile(newFilePath, contents);
 
-          // `\\` needs to be replaced with `/` for Windows so that it can match the path correctly in the middleware manifest.
           functionsMap.set(
             normalizePath(relative(functionsDir, filepath).slice(0, -".func".length)),
             normalizePath(newFilePath)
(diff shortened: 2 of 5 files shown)


dario-piotrowicz:
thanks a lot for the fix @james-elicx 😃 

I wonder if it would be better to use this package instead: https://www.npmjs.com/package/slash

By looking at the code it looks identical to what you're doing (except it handles `\\?\` but that doesn't really concern us, does it?) so there isn't any real benefit code-wise but by using that I think we could just wrap the paths in `slash(...)`, avoid any comment and the code would be self documenting anyways

<details>
<summary>tsdoc</summary>

The function there has also a nice tsdoc so if someone were unsure just hovering over it would tell them you all they need to know anyways:
Screenshot 2023-03-08 at 19 01 58


</details>

james-elicx:
> thanks a lot for the fix @james-elicx 😃
> 
> I wonder if it would be better to use this package instead: https://www.npmjs.com/package/slash
> 
> By looking at the code it looks identical to what you're doing (except it handles `\\?\` but that doesn't really concern us, does it?) so there isn't any real benefit code-wise but by using that I think we could just wrap the paths in `slash(...)`, avoid any comment and the code would be self documenting anyways
> 
> tsdoc

If we use that library and someone is using an extended path, it will just return it and nothing will change in the path name we use to match in the manifest. When they aren't using an extended path, it does the same replace I did.

If we use that library on the newFilePath part too and someone is using extended paths, it will continue to use their original path, but it will fail to build because of the use of backslashes `\`.

image

So, in this case, using that library vs doing this will make no difference. If someone is using extended-length paths on windows, they will receive errors when building because of the backslashes. If they use extended-length paths and replace them with forward slashes, it will probably also error based on the note in the readme of that repository.

So, at the end of the day, if it's an extended-length path, it will error either way and make zero difference.

I have no idea if `tmpdir()` would ever return an extended-length path. I assume not. Either way, it makes no difference to the end result in my view.

tl;dr: using that package is adding a dependency to do the exact same thing. based on their readme, one can assume forward slashes in extended length paths will result in an error, but when building, backwards slashes result in error, so the edge case that package checks will result in error regardless of what we do. `tmpdir()` might never get an extended length path, not sure, but I doubt it would.

james-elicx:
> but by using that I think we could just wrap the paths in slash(...), avoid any comment and the code would be self documenting anyways

I could extract the replacement to a `normalizePath` function if you would prefer?

dario-piotrowicz:
> tl;dr: using that package is adding a dependency to do the exact same thing
yeah I totally get that but it's a minimal one (without any sub-dependencies) so I figured it wouldn't be that bad 🤷 

> I could extract the replacement to a normalizePath function if you would prefer?
yeah I would prefer that if you'd be ok with it 🙂, I think I would really like it if you could create a utils folder and have the function there, with some tsdoc config like the slash one to make it clearer what it's doing, would you do that? 🙏 

(@GregBrimble does that sound ok to you? 🙂)

___

anyways @james-elicx regarding the extended paths, sorry I don't use windows so I am not really familiar with that, is that something that happens often? and doesn't your change break it? (since it replaces the backlashes there?)

james-elicx:
> yeah I would prefer that if you'd be ok with it 🙂, I think I would really like it if you could create a utils folder and have the function there, with some tsdoc config like the slash one to make it clearer what it's doing, would you do that? 🙏
> 
> (@GregBrimble does that sound ok to you? 🙂)
> 
> anyways @james-elicx regarding the extended paths, sorry I don't use windows so I am not really familiar with that, is that something that happens often? and doesn't your change break it? (since it replaces the backlashes there?)

Well, today is the first time I have ever heard of them, so we're kind of in the same boat.

From what I can find on Google, the `\\?` prefix is something that is used when you exceed the `MAX_PATH` limit on Windows, which according to [this](https://learn.microsoft.com/en-gb/windows/win32/fileio/maximum-file-path-limitation) is 260 characters. While tmpdir 

The error from esbuild when you use `\` paths is when your path name incidentally happens to start a unicode escape sequence. So, as my username on Windows is `user`, `\u` is causing this and turning into a big error. So this is why we need to convert the file's path name to forwards slashes, so it doesn't choose to be edgy and interpret it as something it's not intended to be when processing the path. I believe that's the reason it errors.

If someone is using a pathname that is longer than `MAX_PATH` as their `tmpdir()` location on Windows, and there incidentally happens to be a backslash followed by a `u`, I'm not sure if they'd receive the error, but it's an assumption. Changing the path name to have forward slashes also will give them an error because of `MAX_PATH`.

It is worth noting that the default temp directory on Windows is `C:/Users/${username}/AppData/Local/Temp`, so we aren't actually going to ever reach a path name with 260 characters. Now, I have no idea if esbuild will process path names that long, if nodejs will convert them between the two, or what will happen.

The default Windows temporary directory is not at all close to 260 characters, so realistically no one should ever have a path name in that format. It would virtually never occur.

So I guess after this wall of text, the question is what do you want to do? I mean, we can check for it and return it as is if you want to. From the sounds of things it would be an edge case we would probably never encounter.

If you want me to include that, I'll just put it into the function, but please, I pray, don't add a dependency for something simple that can be done with a one-liner.

And I guess today we all learned something new about extended paths on Windows :D

p.s. sorry for the essay, I just kinda was typing as I was reading and it ended up *quite* long.

james-elicx:
Added a utility function for it, as well as a couple of tests.

I would really appreciate it if you guys would consider adding eslint and prettier configs to the root of the repository. That way, we can ensure that contributions follow a similar style and format.

dario-piotrowicz:
> Added a utility function for it, as well as a couple of tests.

Thanks, it looks good 🙂 

Although please let me have some time to play with it and experiment, I'd like to understand better if the solution is correct in regards of extended paths. Are you sure that not doing anything with them is correct? Have you checked how Nextjs' output for extended paths look?

> I would really appreciate it if you guys would consider adding eslint and prettier configs to the root of the repository. That way, we can ensure that contributions follow a similar style and format.

Yeah we definitely should have eslint and prettier, I'll look into it thanks 🙂

james-elicx:
> Although please let me have some time to play with it and experiment, I'd like to understand better if the solution is correct in regards of extended paths. Are you sure that not doing anything with them is correct? Have you checked how Nextjs' output for extended paths look?

Not doing anything with them is the only option it would seem. If we change them, Windows will not be able to recognise them as extended length paths, so just not changing them is probably the only option.

Personally, I don't really view this as being much of an issue. No one will realistically have a path name that long as their temporary directory.

Note that even though we are doing nothing with extended length paths, the part of the string that gets matched in the manifest is still being modified because that is only the file name, not the full path, so that part will always function correctly.

It's just a matter of whether the full path includes a character that causes esbuild to throw an error or not. And when we reach extended length paths territory, we have no options for fixing if that happens.

I couldn't really find anything with regards to how nextjs or vercel deal with them.

dario-piotrowicz (approved):

dario-piotrowicz:
Thanks a bunch @james-elicx the code looks good and it is working as expected on mac, we couldn't test it on windows but the code itself does make sense so we're happy to go with it 🙂 (we'll need to find a way to do windows specific tests)

PR #87, opened 2023-03-07
fix: add dynamic route params to request url

### LINKED ISSUE #25: AppDir Page path params undefined
Hey, 

I am trying to mimic the demo of vercel, to be specific the [ISR with path variables](https://github.com/vercel/app-playground/blob/main/app/isr/%5Bid%5D/page.tsx). However, it appears that the params parameter is always undefined, despite it working with `next build && next start`. 

`app/something/[id]/page.tsx`: 
```
export default async function Page({
  params,
}: {
  params?: any;
  children?: React.ReactNode;
}) {
  console.log(params);
  const data = await fetchData(params.id);
  return (
    <>
      <PageComponent {...data} />
    </>
  );
}
```

Am I missing something? Are path parameters supported?

### LINKED ISSUE #35: Incoming request URL is not updated to include dynamic route params
Next.js attaches dynamic route params as query params. In normal routes, developers usually grab these values from the [`params` object](https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#context-parameter):

```ts
// pages/blog/[slug].ts

export async function getServerSideProps({ params }) {
  // params.slug = 'a-blog-slug'
}
```

But in Edge API routes, there is no such mechanism. In order to grab these values, you have to look at the incoming request URL:

```ts
// pages/api/blog/[slug].ts

export default function handler(request) {
  const url = new URL(request.url)
  url.searchParams.get('slug')
  // should be 'a-blog-slug' but is currently null
}
```

We should pass these handling functions the finalized URL which includes any dynamic route params in the query string.

james-elicx:
On incoming requests to pages with dynamic route parameters, those params need to be extracted from the URL and added to the search parameters so that they can be consumed by the page.

The matcher for the route actually has the regular expression that we need to use to extract the parameters, so we should check if there are any groups when trying to match the pathname to the expression. Then, we can turn those groups into URL search parameters.

Because `request.url` is readonly, we need to create a new URL and append the search parameters to it. After this, we have to create a new request that will then get passed to the entrypoint, so that it has the correct URL with the search params added to it.

I have tested this locally using vercel/app-playground, with single nested dynamic route parameters and double nested dynamic route parameters, inside route groups and out, i.e. `route-groups\(shop)\[categorySlug]\[subCategorySlug]`, and it fixes all the issues.

fixes #35
fixes #25


GregBrimble:
This looks great, thank you!! We'll test it out shortly before merging! :)

Would you mind adding a changeset? Just run `npx changeset` in the root of the repo and select a "patch" change before adding a brief description of the fix you've done here. That'll generate a markdown fine which you can add to this PR and that'll help us release your changes.

james-elicx:
> This looks great, thank you!! We'll test it out shortly before merging! :)
> 
> Would you mind adding a changeset? Just run `npx changeset` in the root of the repo and select a "patch" change before adding a brief description of the fix you've done here. That'll generate a markdown fine which you can add to this PR and that'll help us release your changes.

Added a short and simple summary :)

First time using changeset! A lot simpler than I expected.

rfire12:
Can we get this merged?

madaskig:
Hi! #32 has the most interesting discussion related to this issue IMO but it was closed prematurely :( Contributors over there (@hanford and @viantirreau) correctly identified a missing header used by next `x-matched-path` for matching routes with dynamic params. I'm just wondering if that would be the preferred solution?  What do you think? I'm linking my comment with a working solution [here](https://github.com/cloudflare/next-on-pages/issues/32#issuecomment-1453685867).

james-elicx:
> Hi! #32 has the most interesting discussion related to this issue IMO but it was closed prematurely :( Contributors over there (@hanford and @viantirreau) correctly identified a missing header used by next `x-matched-path` for matching routes with dynamic params. I'm just wondering if that would be the preferred solution? What do you think? I'm linking my comment with a working solution [here](https://github.com/cloudflare/next-on-pages/issues/32#issuecomment-1453685867).

Hello @madaskig, I have tried copy-pasting your patch into my local next-on-pages and it does not fix it.

The only documentation that I've been able to find about `x-matched-path` states that it is a response header. The [page](https://vercel.com/docs/concepts/edge-network/headers) for Vercel's edge network headers does not list it as one that Vercel attaches to requests. And from my testing with vercel/app-playground, it doesn't appear to have any impact on what happens.

image

rfire12:
> > Hi! #32 has the most interesting discussion related to this issue IMO but it was closed prematurely :( Contributors over there (@hanford and @viantirreau) correctly identified a missing header used by next `x-matched-path` for matching routes with dynamic params. I'm just wondering if that would be the preferred solution? What do you think? I'm linking my comment with a working solution [here](https://github.com/cloudflare/next-on-pages/issues/32#issuecomment-1453685867).
> 
> Hello @madaskig, I have tried copy-pasting your patch into my local next-on-pages and it does not fix it.
> 
> The only documentation that I've been able to find about `x-matched-path` states that it is a response header. The [page](https://vercel.com/docs/concepts/edge-network/headers) for Vercel's edge network headers does not list it as one that Vercel attaches to requests. And from my testing with vercel/app-playground, it doesn't appear to have any impact on what happens.
> 
> image
> 
> Hanford also wrote in the issue you linked that it did not work for him either, which is why he closed his PR for it.

It didn't work for me either. Only @james-elicx solution has worked so far.

This issue is a big blocker.

cc:
@GregBrimble @dario-piotrowicz

dario-piotrowicz:
I think the PR's change look good 🙂, I'll test it soon, I also want to check the `x-matched-path` header, I strongly suspect that the current PR's solution is the way to go but I want to double check the header before merging just to understand the situation better (and be sure we're applying the right fix).

@james-elicx could you please rebase? 🙏

james-elicx:
> I think the PR's change look good 🙂, I'll test it soon, I also want to check the `x-matched-path` header, I strongly suspect that the current PR's solution is the way to go but I want to double check the header before merging just to understand the situation better (and be sure we're applying the right fix).
> 
> @james-elicx could you please rebase? 🙏

I have just rebased it. Never done a rebase before, but I believe I did it correctly.

> Apologies, the patch I linked to was made for v0.3.0. Here is an up-to-date branch: https://github.com/madaskig/next-on-pages/tree/add-x-matched-path-header
> 
> In any case I don't want to delay the merging of this PR and having a fix for this issue asap would great!

I have just tried copy-pasting from the new version you linked. Setting the `x-matched-path` header on the request still doesn't fix anything for me. Maybe you are using the Next.js patch hanford posted on that issue? Anyway, the documentation points to it being a reserved response header and not one that Vercel adds to requests.

templates/_worker.js/index.ts
```diff
@@ -159,10 +159,21 @@ export default {
...
 
+					const nextPathnameMatcher = nextPathname.match(regexp);
+
 					if (
-						nextPathname.match(regexp) ||
+						nextPathnameMatcher ||
 						`${nextPathname}/page`.replace('//page', '/page').match(regexp)
 					) {
+						if (nextPathnameMatcher?.groups) {
+							const params = new URLSearchParams(nextPathnameMatcher.groups);
```
dario-piotrowicz, line 169: you could just have
```ts
const params = Object.entries(nextPathnameMatcher.groups);
```
which to me looks much simpler/cleaner.

is using a `URLSearchParams` necessary for some reason?

dario-piotrowicz:
@james-elicx I rebased your PR as it had a new merge conflict I wanted to double check it and merge it, I hope you don't mind :pray: 

(PS: but I haven't merged as I have some questions :sweat:)

dario-piotrowicz:
I've tried a dynamic route with an encoded param, this is my component:
```tsx
import { useRouter } from "next/router";

export const config = { runtime: "experimental-edge" };

export async function getServerSideProps({
  params,
}: {
  params: { routeName: string };
}) {
  return {
    props: {
      routeNameFromServiceSideProps: params.routeName,
    },
  };
}

export default function RouteADynamic({
  routeNameFromServiceSideProps,
}: {
  routeNameFromServiceSideProps: string;
}) {
  const router = useRouter();
  const { routeName } = router.query;

  return (
    <>
      <h1>Route A Dynamic :</h1>
      <h2>routeName: {routeName}</h2>
      <h2>routeNameFromServiceSideProps: {routeNameFromServiceSideProps}</h2>
    </>
  );
}
```

 this is how it looks using next dev:
Screenshot at 2023-03-18 17-04-57

And this is how it looks using next-on-pages:
Screenshot at 2023-03-18 17-05-51

(mh... not sure why the styling looks different :thinking:)

It seems like we're missing a decoding step.

I think applying [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) to the param values should be enough to fix the issue for the loader at least.

What do you think @james-elicx, would you like to give it a go?

Since for some reason we even have the client side value non decoded (I haven't looked into it, I'm not too sure why :confused:) I would be ok merging this without the decoding and open an issue to fix the decoding in both client side and server side code, alternatively we can fix the server side decoding since we're already touching it (and looks like a small change to me) and create an issue just to fix the client side part, up to you I am fine with either :slightly_smiling_face:

dario-piotrowicz:
Regarding the `x-matched-path` header, I think that the solution in this PR is valid, also it seems like that header isn't even properly working on edge functions? (based on this [comment](https://github.com/vercel/next.js/blob/1467be886a94ac2e66a0442bd6bec1dc53da6608/packages/next/src/server/base-server.ts#L1945)).

Since the solution here relies on regexes that we already rely upon it does seem like the better/more robust one to me.

So I'd just merge this (and we can always anyways revert and invest time and look more into `x-matched-path` if needed).

I'm happy to merge this as soon as the above comments are addressed :slightly_smiling_face:

EdenCoder (approved):

templates/_worker.js/index.ts
james-elicx: Doesn't appear to have any difference. Will use `Object.entries` instead.

james-elicx:
> I think applying [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) to the param values should be enough to fix the issue for the loader at least.
> 
> What do you think @james-elicx, would you like to give it a go?
> 
> Since for some reason we even have the client side value non decoded (I haven't looked into it, I'm not too sure why 😕) I would be ok merging this without the decoding and open an issue to fix the decoding in both client side and server side code, alternatively we can fix the server side decoding since we're already touching it (and looks like a small change to me) and create an issue just to fix the client side part, up to you I am fine with either 🙂

I have tried `decodeURIComponent` on both the value being passed as the search param and the stringified URL but both give me the encoded version on the client when using next-on-pages. Not sure if there's a quick fix that will work with this solution.

I have been doing some research into the Vercel build output config file and it turns out that we aren't processing it correctly when it comes to the different routing phases. From what I've been able to gather, dynamic params, rewrites, and redirects will all be handled if we handle the build output config correctly.

The documentation isn't the best about how the different routing phases work, but I'm going to play around with it and see if I can come up with a proper routing solution using that config over the next few days. I'll open a PR once I have something working that walks through the logic behind how I'm processing each part.

I am comfortable with letting this PR get merged as a temporary fix until we process the build output config as intended.

Apologies for the delay - my uni's internet has been offline for the past 24 hours 🙄. For what it's worth, I'm always happy to chat on Discord if you would like an instant messaging option instead of the slow GitHub back and forth.

james-elicx:
> This PR is set to fix #25, but it doesn't seem to be fixing it for me, running an app dir application I get an internal server error which seems to indicate that we don't get the params in the component.
> 
> The error comes from this line: https://github.com/vercel/next.js/blob/6b09bc86a89b620616d84e5161c9e5eed873a10f/packages/next/src/server/app-render/index.tsx#L235 which throws since `pathParams` is undefined.
> 
> I'm fine if this change doesn't fix the app dir params passing, I just wouldn't want to prematurely close #25 if it is not actually getting fixed here.
> 
> Could you double check? am I missing something or is this not actually fixing #25?

I have just tested with a fresh clone of this branch, then following the reproduction steps that I previously wrote to you in the minification PR (minus no. 3 as it is not relevant). All my PRs have been done using a version of vercel/app-playground that you can find on my repos, so this fixes it for app dir for me :)

I would also add that this solution is very very very similar to how Vercel handles it with their routing phases. The `rewrite` routing phase uses regular expressions to rewrite the URL from `/.../[param]` to `/.../[paramName]?paramName=param`. So the only reasons I can think of it not working are either not being in this branch, or using a version of Next.js that has a bug with the edge runtime.

<details>
<summary>Images of it working.</summary>

image

image

image
</details>

dario-piotrowicz:
Thanks a lot @james-elicx 😄 , sorry about all the back and forth 🙇

james-elicx:
> @james-elicx no yeah you're right, the Next.js version I was testing with (13.0.0) might have been the problem, I bumped it up and now the params seem to work fine!
> 
> sorry about that 😓

Haha, I thought that might be the case. The earlier versions were insanity for me when I was first trying it after release 😂

dario-piotrowicz:
@shayanb just release 0.5.0 🙂:+1:  (https://www.npmjs.com/package/@cloudflare/next-on-pages/v/0.5.0)

shayanb:
@dario-piotrowicz perfect, thanks for the ping 🙏

vykintasmak:
@dario-piotrowicz I think the issue still persists to some extent. I have a working example:
If you'll go to liquidjpeg.xyz and enter 0xE34f580ed6D3D8ad7719B4f6f36b2CcC61A8cEE8 address to input field you will be redirected to liquidjpeg.xyz/wallets/0xE34f580ed6D3D8ad7719B4f6f36b2CcC61A8cEE8 as expected

But if you will go directly to this same URL you will get 404

[Heres a build log from CF pages](https://github.com/cloudflare/next-on-pages/files/11160156/liquidjpeg.848cafdd-0371-441f-8b12-762dfa71634a.log) and it shows 0.6.0 is used to build. Let me know if you need anything so it would be possible to debug it

shayanb:
Same, we had a similar use case as 👆, but at the end had to move to Vercel as it worked there with no issues. 
Tried many approaches to get it to work on CF Pages, but needed to launch so moved there a day before our launch :)

dario-piotrowicz:
Thank you very much for the comment @vykintasmak

Currently @james-elicx is putting a significant effort in improving the routing (see: https://github.com/cloudflare/next-on-pages/issues/129) I don't think there is too much point in looking into this right now since things are going to change quite a bit in respect to routing and your issue might very likely get resolved by that.

What I would suggest to do is wait for James to change the routing and see if the issue still persist, if it does we can the look into why and address it.

If you want I would suggest you to open a [new issue](https://github.com/cloudflare/next-on-pages/issues/new) so that we keep track of this and don't forget to check if it gets fixed or not, feel free to tag me there.

If you can please also share your app's code since that is always extremely helpful (but since you haven't done so already I'd imagine that maybe you can't)

<details>
<summary>if you can't share your code</summary>

If you can't share the app's code what I would anyway already find very useful is for example if you could outline that your page does and a snippet of your code (there you can rename variables, replace private code with `...`, etc...).

It's not as useful as the real working code of course but still much better than nothing :smiley: 
</details>

dario-piotrowicz:
> Same, we had a similar use case as point_up_2, but at the end had to move to Vercel as it worked there with no issues.
Tried many approaches to get it to work on CF Pages, but needed to launch so moved there a day before our launch :)

:cry: 

Hopefully you'll give next-on-pages another shot, we've been trying quite hard to make things better as of late :pray:

shayanb:
> Hopefully you'll give next-on-pages another shot, we've been trying quite hard to make things better as of late :pray: 

For sure, I'll keep an eye on this thread 🤞 
Great job overall, we were almost there on next-on-pages but you know, sh*t happens :) there'll be a lot more apps in the near future.

PR #88, opened 2023-03-07
fix: minification

### LINKED ISSUE #47: Massively inflated bundle size with pages
Following the installation guide in this repo, my initial `_worker.js` output is 370KB.

If I then simply duplicate `pages/index.ts` to `pages/index2.ts`, and rebuild, I get over 750KB. And then if I do the same thing with `index3.ts`, I get over 1MiB. This is adding around 300KB for _every single page_. 

Any real-world site is going to rapidly exceed Workers limits, even with the increase to 5MiB. We've seen multiple people in Discord report hitting this limit with sites that had as little as 7 pages!

Here are some numbers for the generated `_worker.js` when using the experimental minify option, and when not:

## using `--experimental-minify`

1 page site: 370KB output
2 page site: 750KB output
3 page site: 1MiB output

## without `--experimental-minify`

1 page site: 333KB output
2 page site: 622KB output
3 page site: 911KB output

james-elicx:
### Bullet point version:

- Fix which property is checked for `webpackChunk_N_E`
- Use regex to hack around a bug caused by our webpack minification. It doesn't appear to cause issues.
- Set esbuild's `minify` option to be enabled when the user uses `--experimental-minify`
 
### Statistics:

Build is 3289854 bytes
  - change (regular): -19566933 bytes.
  - change (esbuild): -12028637 bytes.
  - change (experim): -2272043 bytes.
  - change (exp+esb): +0 bytes.

### Long version:

The primary reason minification was failing was due to checking the wrong property for `webpackChunk_N_E`. This was actually stored under `name` instead of `value`. Below is an example of the property object:
```
Node3 {
  type: 'Identifier',
  start: 57372,
  end: 57388,
  name: 'webpackChunk_N_E'
}
```

I did actually end up coming across a bug that only occurs if you use `--experimental-minify`, meaning it doesn't occur when there is no minification, and it doesn't occur if you only use esbuild's minification. It is rather bizarre. The below part of one of the minified webpack chunks results in an error due to the value being undefined. I have traced this back to [next/server/web/adapter.ts](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L237) in the next.js repo.

```
return { response: x2, waitUntil: Promise.all(h2[s]) };
```

I've thought and thought about this but I cannot figure out why it is occurring. Well, I can, because I did console.log on each value, but the point is that it *shouldn't* happen, but for some reason, it does when we minify the webpack chunks.

image

At the moment, my workaround for this bug is using regex to replace the instances of it in each relevant chunk to have a nullish coalescing operator followed by an empty array. This means that instead of erroring out and breaking because it received undefined, it gets an empty array.

From the testing in vercel/app-playground that I did locally when trying this fix, it doesn't seem like it is causing any noticeable issues. I've done a lot of playing around and have encountered no issues caused by it.

As minify is experimental like the flag states, I feel that this is an acceptable hack around for the bug until a better solution can be identified, if any.

The final change I have made is to toggle esbuild's `minify` option when the user uses `--experimental-minify`

I do think that there is probably going to be more room for optimization with other parts. For example, the metadata for static routes is written in the final _worker.js 38 times for vercel/app-playground. But those would be optimizations to explore in another PR.

fixes #47

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -490,6 +490,7 @@ const transform = async ({
       __CONFIG__: JSON.stringify(config),
     },
     outfile: ".vercel/output/static/_worker.js",
+    minify: experimentalMinify,
   });
 
   console.log("⚡️ Generated '.vercel/output/static/_worker.js'.");
--- a/src/index.ts
+++ b/src/index.ts
@@ -241,7 +241,7 @@ const transform = async ({
                   expression.callee?.type === "MemberExpression" &&
                   expression.callee.object?.type === "AssignmentExpression" &&
                   expression.callee.object.left?.object?.name === "self" &&
-                  expression.callee.object.left.property?.value ===
+                  expression.callee.object.left.property?.name ===
                     "webpackChunk_N_E" &&
                   expression.arguments?.[0]?.elements?.[1]?.type ===
                     "ObjectExpression"
--- a/src/index.ts
+++ b/src/index.ts
@@ -228,6 +228,12 @@ const transform = async ({
           );
 
           if (experimentalMinify) {
+            // TODO: Investigate alternatives or a real fix for this. Maybe a PR is needed in Next.js?
+            // This is a hack to fix an issue with --experimental-minify where compiled next/server/web/adapter.ts throws:
+            // `Unhandled Promise Rejection: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
+            // Source code: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L237
+            contents = contents.replace(/(?:(return{response:\w+,waitUntil:Promise\.all\(\w+\[\w+\])(\)}))/gm, "$1??[]$2");
+
             const parsedContents = parse(contents, {
               ecmaVersion: "latest",
               sourceType: "module",
--- a/src/index.ts
+++ b/src/index.ts
@@ -228,12 +228,6 @@ const transform = async ({
           );
 
           if (experimentalMinify) {
-            // TODO: Investigate alternatives or a real fix for this. Maybe a PR is needed in Next.js?
-            // This is a hack to fix an issue with --experimental-minify where compiled next/server/web/adapter.ts throws:
-            // `Unhandled Promise Rejection: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
-            // Source code: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L237
-            contents = contents.replace(/(?:(return{response:\w+,waitUntil:Promise\.all\(\w+\[\w+\])(\)}))/gm, "$1??[]$2");
-
             const parsedContents = parse(contents, {
               ecmaVersion: "latest",
               sourceType: "module",
--- a/.changeset/twenty-spoons-search.md
+++ b/.changeset/twenty-spoons-search.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+Fix Webpack minification and toggle esbuild minification.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "@cloudflare/next-on-pages",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@cloudflare/next-on-pages",
-      "version": "0.2.0",
+      "version": "0.3.0",
       "dependencies": {
         "acorn": "^8.8.0",
-        "astring": "^1.8.3",
+        "astring": "^1.8.4",
         "chokidar": "^3.5.3",
         "cookie": "^0.5.0",
         "esbuild": "^0.15.3"
@@ -1678,9 +1678,9 @@
       }
     },
     "node_modules/astring": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/astring/-/astring-1.8.3.tgz",
-      "integrity": "sha512-sRpyiNrx2dEYIMmUXprS8nlpRg2Drs8m9ElX9vVEXaCB4XEAJhKfs7IcX0IwShjuOAjLR6wzIrgoptz1n19i1A==",
+      "version": "1.8.4",
+      "resolved": "https://registry.npmjs.org/astring/-/astring-1.8.4.tgz",
+      "integrity": "sha512-97a+l2LBU3Op3bBQEff79i/E4jMD2ZLFD8rHx9B6mXyB2uQwhJQYfiDqUwtfjF4QA1F2qs//N6Cw8LetMbQjcw==",
       "bin": {
         "astring": "bin/astring"
       }
@@ -7430,9 +7430,9 @@
       "dev": true
     },
     "astring": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/astring/-/astring-1.8.3.tgz",
-      "integrity": "sha512-sRpyiNrx2dEYIMmUXprS8nlpRg2Drs8m9ElX9vVEXaCB4XEAJhKfs7IcX0IwShjuOAjLR6wzIrgoptz1n19i1A=="
+      "version": "1.8.4",
+      "resolved": "https://registry.npmjs.org/astring/-/astring-1.8.4.tgz",
+      "integrity": "sha512-97a+l2LBU3Op3bBQEff79i/E4jMD2ZLFD8rHx9B6mXyB2uQwhJQYfiDqUwtfjF4QA1F2qs//N6Cw8LetMbQjcw=="
     },
     "async-sema": {
       "version": "3.1.1",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   ],
   "dependencies": {
     "acorn": "^8.8.0",
-    "astring": "^1.8.3",
+    "astring": "^1.8.4",
     "chokidar": "^3.5.3",
     "cookie": "^0.5.0",
     "esbuild": "^0.15.3"


mrluc:
@james-elicx cool, this seems super promising for me locally. That regex and fix looks sketchy though 😆.

| Before | After |
|-|-|
| 5.2 MB | 793 KB |

I also tested adding 10 mostly-identical pages, and as you mention, the size does creep up more than a lot of folks would probably like -- the fellow with 600 pages/*.tsx from the linked issue is out of luck still, I'd imagine -- it's tremendously better; bundle size goes up a few tens of K per pages probably.

Gonna test deploying with it, see if that solves the dreaded 'CPU time limit' error.

Which is funny, because I think it's actually a bundle size error? We're on a paid plan, and 5MB worker size was *just* crossed -- so I think this might be an inaccurate error, due either to Pages or 'next-on-pages'.

james-elicx:
> @james-elicx cool, this seems super promising for me locally. That regex and fix looks sketchy though 😆.

It might not look the prettiest but all the regex is doing really is just falling back to a blank array when the value is null so that the promise.all has something to resolve instead of throwing an error.

I agree that it's not ideal, but it's the only workaround that I can think of at the moment. Still stumped as to why the bug occurs, and still cannot figure out if it actually is affecting anything.

More optimizations are definitely possible. I have just managed to reduce my final build size by another -1,465,810 bytes, taking it to 1,797,877 by minifying the manifests.

I am now doing 22,830,243 -> 1,797,877 bytes.

I think it will be best to wait until my file paths PR is merged so that I can open a new PR for additional optimizations and using the `utils` folder that I created in that PR. What do you think @dario-piotrowicz?

dario-piotrowicz:
> > @james-elicx cool, this seems super promising for me locally. That regex and fix looks sketchy though 😆.
> 
> It might not look the prettiest but all the regex is doing really is just falling back to a blank array when the value is null so that the promise.all has something to resolve instead of throwing an error.
> 
> I agree that it's not ideal, but it's the only workaround that I can think of at the moment. Still stumped as to why the bug occurs, and still cannot figure out if it actually is affecting anything.
> 
> More optimizations are definitely possible. I have just managed to reduce my final build size by another -1,465,810 bytes, taking it to 1,797,877 by minifying the manifests.
> 
> I am now doing 22,830,243 -> 1,797,877 bytes.
> 
> I think it will be best to wait until my file paths PR is merged so that I can open a new PR for additional optimizations and using the `utils` folder that I created in that PR. What do you think @dario-piotrowicz?

yeah awesome stuff @james-elicx 😄 

___

regarding utils if you want to start working on stuff I would suggest you to just create a new branch and re-create that folder with the index file then it will be a single very easy to fix merge conflict (anyways sorry about the later reply there, I still need to play a bit with that, also too bad I don't have a windows machine to test that on 😓)

___

Regarding the stuff here it looks like it's making a huge improvement, thanks a lot!
of course the regex solution is very very hacky... I'll review it more (and mull over it) properly tomorrow 🙂👍 

If you could share maybe an example app in which you encountered that undefined value that would really be helpful 🙏

james-elicx:
> regarding utils if you want to start working on stuff I would suggest you to just create a new branch and re-create that folder with the index file then it will be a single very easy to fix merge conflict (anyways sorry about the later reply there, I still need to play a bit with that, also too bad I don't have a windows machine to test that on 😓)
> 
> Regarding the stuff here it looks like it's making a huge improvement, thanks a lot! of course the regex solution is very very hacky... I'll review it more (and mull over it) properly tomorrow 🙂👍
> 
> If you could share maybe an example app in which you encountered that undefined value that would really be helpful 🙏

Sure thing. Here is the slightly modified clone of vercel/app-playground that I've been doing all my testing with - https://github.com/james-elicx/app-playground-edge

dario-piotrowicz:
> > regarding utils if you want to start working on stuff I would suggest you to just create a new branch and re-create that folder with the index file then it will be a single very easy to fix merge conflict (anyways sorry about the later reply there, I still need to play a bit with that, also too bad I don't have a windows machine to test that on 😓)
> > Regarding the stuff here it looks like it's making a huge improvement, thanks a lot! of course the regex solution is very very hacky... I'll review it more (and mull over it) properly tomorrow 🙂👍
> > If you could share maybe an example app in which you encountered that undefined value that would really be helpful 🙏
> 
> Sure thing. Here is the slightly modified clone of vercel/app-playground that I've been doing all my testing with - https://github.com/james-elicx/app-playground-edge

Awesome!!! thank you so very much!!! 😍

src/index.ts
```diff
@@ -241,7 +247,7 @@ const transform = async ({
                   expression.callee?.type === "MemberExpression" &&
                   expression.callee.object?.type === "AssignmentExpression" &&
                   expression.callee.object.left?.object?.name === "self" &&
-                  expression.callee.object.left.property?.value ===
+                  expression.callee.object.left.property?.name ===
```
dario-piotrowicz, line 250: (mostly note to self 😅)

I've checked this and I believe it's correct, the expression is:
`(self.webpackChunk_N_E = self.webpackChunk_N_E || []).push([[...], {...`

so `webpackChunk_N_E`'s indeed a prop name there

<details>

Screenshot 2023-03-10 at 13 59 35


</details>

dario-piotrowicz:
Hi @james-elicx sorry to bother you, I couldn't get the `cannot read property Symbol(Symbol.iterator)` issue and suggestion on how to try to reproduce it?

dario-piotrowicz:
@james-elicx thanks a bunch for the reproduction steps, I did manage to reproduce the issue 🙂 

Sorry for reviewing the PR taking a bunch of time, I'm trying to understand the issue 😕

To be honest I feel like that might not be the right solution and we're sort of hiding/patching the real issue instead of fixing it (but anyways I'm still looking into it, I might change my mind 😅)

For example, instead of adding `?? []` shouldn't we instantiate that field? can not having that field defined break other places? etc...

Any more context you could provide would be appreciated btw 😅  

Also, I think that the bugfix is the only questionable thing in this PR, so if you want you could externalize it in a separate PR and we can merge the rest just to fix #47 and deal with the bug separately, whatever works best for you 🙂

james-elicx:
> To be honest I feel like that might not be the right solution and we're sort of hiding/patching the real issue instead of fixing it (but anyways I'm still looking into it, I might change my mind 😅)
> 
> For example, instead of adding `?? []` shouldn't we instantiate that field? can not having that field defined break other places? etc...
> 
> Any more context you could provide would be appreciated btw 😅
> 
> Also, I think that the bugfix is the only questionable thing in this PR, so if you want you could externalize it in a separate PR and we can merge the rest just to fix #47 and deal with the bug separately, whatever works best for you 🙂

I also feel the same way about it. It doesn't seem like an ideal solution and I don't understand why the value would be undefined only when we minify the Webpack chunks. From my playing around, it didn't *appear* to impact anything, but I also haven't tried using middleware.

For some reason, it would appear that the waitUntil array is under the property `s` instead of the symbol(?). That's what appears to be happening when you look at the console.log I included in my first post. Looking at the [source](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/fetch-event.ts#L9) code for NextFetchEvent, it should use the same symbol that is used to access it in the function.

So that's why I ended up doing `?? []`. Nothing about it being undefined makes sense to me. It shouldn't be. It should be defined when the a new instance of `NextFetchEvent` is created on [line 121](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L121).

I shall remove it from this PR and create a new PR for it after this is merged, per your suggestion, although I fear that running a minified app won't work properly without it, at least it won't for me. Pasting what was cut out below so I have easy access to it.

<details>
<summary>Cut out part.</summary>

```ts
// TODO: Investigate alternatives or a real fix for this. Maybe a PR is needed in Next.js?
// This is a hack to fix an issue with --experimental-minify where compiled next/server/web/adapter.ts throws:
// `Unhandled Promise Rejection: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
// Source code: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L237
contents = contents.replace(/(?:(return{response:\w+,waitUntil:Promise\.all\(\w+\[\w+\])(\)}))/gm, "$1??[]$2");
```
</details>

dario-piotrowicz (approved):

dario-piotrowicz:
@james-elicx Thanks a lot for removing the fix 🙂 (and for the PR in general of course! ❤️ )

As we said if you want feel free to open a new PR with that if you want we can continue the conversation there and hopefully get to the bottom of it 🙂 (or if you want to leave it for a bit and just create an issue what also works too 🙂)

Regarding `NextFetchEvent` yeah I saw that too, it's been quite hard to see what's going on since we get the code minified, I am afraid that our experimenalMinify can be messing with the code and that for some reason we are missing that field? do you think that could be the case? anyways yeah digging more into the experimentalMinify code it's what I next want to try 🙂

james-elicx:
> @james-elicx Thanks a lot for removing the fix 🙂 (and for the PR in general of course! ❤️ )
> 
> As we said if you want feel free to open a new PR with that if you want we can continue the conversation there and hopefully get to the bottom of it 🙂 (or if you want to leave it for a bit and just create an issue what also works too 🙂)
> 
> Regarding `NextFetchEvent` yeah I saw that too, it's been quite hard to see what's going on since we get the code minified, I am afraid that our experimenalMinify can be messing with the code and that for some reason we are missing that field? do you think that could be the case? anyways yeah digging more into the experimentalMinify code it's what I next want to try 🙂

@dario-piotrowicz Did you try disabling the esbuild minification to make debugging easier? It can help make it much more readable and followable. But yeah, it's something to do with minifying the Webpack chunks. If you try my repro steps without the experimental-minify flag, it runs fine, but with it, it has the error. It's strange. Opened #95 for it.

Added a changeset too.

Assuming this gets merged, would you be okay with me making a PR afterwards that refactors the minification logic out into the utility directory? I did something along those lines when I was playing around locally and would be happy to open a PR for it.

dario-piotrowicz:
Thanks a lot @james-elicx (I did forget to ask you about the changeset 😓 ❤️ 👍)

Regarding the minification logic, splitting the index.ts file is something I've got on my list of things to do, but we're trying to merge some PRs first so that we don't disrupt collaborators too much. So we'll get to it soon anyways so I would suggest you to maybe not do it at least for now.

I think as we split the index.ts file we'll want to add unit tests for the various utilities as well so that should also be part of the refactoring I think.

If you want I can keep you in the loop regarding that, but it may be something that may just come out of the global index refactoring/splitting anyways.

james-elicx:
> Regarding the minification logic, splitting the index.ts file is something I've got on my list of things to do, but we're trying to merge some PRs first so that we don't disrupt collaborators too much. So we'll get to it soon anyways so I would suggest you to maybe not do it at least for now.
> 
> I think as we split the index.ts file we'll want to add unit tests for the various utilities as well so that should also be part of the refactoring I think.
> 
> If you want I can keep you in the loop regarding that, but it may be something that may just come out of the global index refactoring/splitting anyways.

That's understandable. Well, I'd be more than happy to help contribute to some of the efforts if you do decide to open up parts of that process.

dario-piotrowicz:
@james-elicx Awesome, thanks you very much for the offer it's very much appreciated, sure we'll see how it goes and try open it up, in that case we'll ping you 🙂👍

james-elicx:
Note for reviewers: the `astring` bump is necessary to fix the bug that we dove into in #95 and was touched on above.

PR #89, opened 2023-03-08
fix: external fetch requests

### LINKED ISSUE #36: With Next.js13 app-directory, Fetch API cause error when optional argument set
Here is my `app/page.tsx`code.(`experimental-edge` is enabled)

```javascript
export default async function Page() {
  // OK
  const json = await fetch("https://jsonplaceholder.typicode.com/todos/1").then(
    (res) => res.json()
  );

  // Error
  const json = await fetch("https://jsonplaceholder.typicode.com/todos/1", {
    method: "GET",
  }).then((res) => res.json());

  return (
    <>
      <p>{json.userId}</p>
      <p>{json.title}</p>
    </>
  );
}
```

image

james-elicx:
The Cloudflare workers runtime does not implement certain properties, for instance `mode` and `credentials`. When [generating](https://github.com/vercel/next.js/blob/canary/packages/next/src/compiled/react/cjs/react.shared-subset.development.js#L191) a cache key for request deduping, the React Fetch 'monkey-patch' implementation tries to read some of these properties.

Therefore, the simple solution is to replace these specific properties with `null` when we build the _worker.js script as they would have no effect on the cache keys anyway.

https://github.com/cloudflare/workerd/blob/main/src/workerd/api/http.h#L518
https://github.com/cloudflare/miniflare/blob/master/packages/core/src/standards/http.ts#L501

fixes #36

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -227,6 +227,13 @@ const transform = async ({
             "$1true$3"
           );
 
+          // The workers runtime does not implement certain properties like `mode` or `credentials`.
+          // Due to this, we need to replace them with null so that request deduping cache key generation will work.
+          contents = contents.replace(
+            /(?:(JSON\.stringify\(\[\w+\.method\S+,)\w+\.mode(,\S+,)\w+\.credentials(,\S+,)\w+\.integrity(\]\)))/gm,
+            "$1null$2null$3null$4"
+          );
+
           if (experimentalMinify) {
             const parsedContents = parse(contents, {
               ecmaVersion: "latest",
--- a/src/index.ts
+++ b/src/index.ts
@@ -227,8 +227,10 @@ const transform = async ({
             "$1true$3"
           );
 
+          // TODO: Investigate alternatives or a real fix. This hack is rather brittle.
           // The workers runtime does not implement certain properties like `mode` or `credentials`.
           // Due to this, we need to replace them with null so that request deduping cache key generation will work.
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -227,6 +227,13 @@ const transform = async ({
             "$1true$3"
           );
 
+          // The workers runtime does not implement certain properties like `mode` or `credentials`.
+          // Due to this, we need to replace them with null so that request deduping cache key generation will work.
+          contents = contents.replace(
+            /(?:(JSON\.stringify\(\[\w+\.method\S+,)\w+\.mode(,\S+,)\w+\.credentials(,\S+,)\w+\.integrity(\]\)))/gm,
+            "$1null$2null$3null$4"
+          );
+
```
dario-piotrowicz, line 236: This is quite brittle, could you add a TODO to fix this like the `contents.replace` above?

Also I think it would be valuable if you could add a link to the code you're patching (https://github.com/vercel/next.js/blob/cbd3c2983ab8730594d7a9633f56a39c1e5b920f/packages/next/src/compiled/react/cjs/react.shared-subset.development.js#L198)

dario-piotrowicz (approved):
Thanks a bunch @james-elicx for the fix 😄

I've tested it locally and it does seem to fix https://github.com/cloudflare/next-on-pages/issues/36 🙂

james-elicx:
Added a todo, source code link, and changeset :)

PR #34, opened 2022-11-27
support detect and use corresponding manager and fix windows related error

LOLBRUHNICE:
This PR is created to achieve the following things: 
1. Use the package manager that was used in the project to build
> + yarn v2+ without `nodeLinker` set to `node-modules` isn't supported since vercel build pipeline doesn't support Plug'n'Play
2. fix `spawn npm ENOENT` error occured when running on windows
3. normalize path when running on windows (back slash to forward slash) so esbuild can identify the path properly
> slash normalization changes:
> + [L544-L547](https://github.com/LOLBRUHNICE/next-on-pages/blob/main/src/index.ts#L544-L547)
> + [L557](https://github.com/LOLBRUHNICE/next-on-pages/blob/main/src/index.ts#L557)
> + [L571-574](https://github.com/LOLBRUHNICE/next-on-pages/blob/main/src/index.ts#L571-L574)
> + [L575-578](https://github.com/LOLBRUHNICE/next-on-pages/blob/main/src/index.ts#L575-L578)

Fixes #52

PR #90, opened 2023-03-09
Add eslint and prettier

dario-piotrowicz:


Cherry:
Can we get hard tabs please as per https://github.com/cloudflare/workers-sdk/issues/1298? 😄 The configs at https://github.com/cloudflare/next-on-pages/pull/34 should be good.

.prettierrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    singleQuote: true,
+    arrowParens: 'avoid'
```
Cherry, line 3:
```suggestion
    arrowParens: 'avoid',
    useTabs: true
```
😄

dario-piotrowicz:
@Cherry I know you already looked at the prettier file, but I just wanted to double check with you 🙂 

it's almost exactly the same as the workers-sdk one: https://github.com/cloudflare/workers-sdk/blob/main/.prettierrc
it has but it adds `arrowParens: 'avoid',` and has `"singleQuote"` set to `true` instead of `false`, these are my personal preferences, let me know if we can keep them as is, or if you think we should 100% match workers-sdk I am fine to change these as well 🙂:+1:

Cherry (approved):
I think `singleQuote` and `arrowParens` are completely personal preference, so if you prefer one over the other, sounds good to me! 🧡

jculvey (approved):

PR #95, opened 2023-03-13
fix: strange minification bug

james-elicx:
This is a follow-up to #88. When using the `--experimental-minify` flag, there is a strange bug after the Webpack chunks are minified. 

```
Unhandled Promise Rejection: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

The cause of this appears to be the `waitUntil` value on the `FetchEvent` [class](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/fetch-event.ts#L9) through `NextFetchEvent` not being defined, despite a new instance of the class being [created](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L121).

The spot in the Next.js source code where this error is through is located [here](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L237). It does not make sense. When we don't minify the Webpack chunks, it is fine, but when we minify them, the error is thrown while running the minified app.

At the moment, I've committed a hack around it which should be replaced with something better in the future. It adds a nullish coalescing operator follow by a blank array so that the promise has an empty array to resolve, instead of trying to resolve undefined.

```ts
return { response: x2, waitUntil: Promise.all(h2[s]) };
```

image

<details>
<summary>Reproduction Steps</summary>

1. Run `npx vercel build` in the clone of [app-playground-edge ](https://github.com/james-elicx/app-playground-edge).
2. Copy `.next` and `.vercel` to the directory with the clone of this branch.
3. Commented out the `contents = contents.replace(.....)` that this PR added.
4. Added `AsyncLocalStorage` import and assigned it to `globalThis` in `globals.js`.
5. Added `node:async_hooks` as external to esbuild.
6. Ran `npm run build && node ./bin/index.js --skip-build --experimental-minify` to build next-on-pages and compile the `_worker.js` script.
7. Ran `cd .vercel/output/static && npx wrangler pages dev . --compatibility-date=2023-03-07 --compatibility-flag=nodejs_compat --bundle`
8. Pressed `b` to open the web page.
9. The error was thrown in the console I ran wrangler from.
</details>

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -228,6 +228,12 @@ const transform = async ({
           );
 
           if (experimentalMinify) {
+            // TODO: Investigate alternatives or a real fix for this. Maybe a PR is needed in Next.js?
+            // This is a hack to fix an issue with --experimental-minify where compiled next/server/web/adapter.ts throws:
+            // `Unhandled Promise Rejection: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
+            // Source code: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/adapter.ts#L237
+            contents = contents.replace(/(?:(return{response:\w+,waitUntil:Promise\.all\(\w+\[\w+\])(\)}))/gm, "$1??[]$2");
+
             const parsedContents = parse(contents, {
               ecmaVersion: "latest",
               sourceType: "module",


dario-piotrowicz:
@james-elicx just to continue the conversation from the other PR...

this also looks broken to me, what do you think?
Screenshot 2023-03-13 at 21 40 16

Compared to the nextjs source:
Screenshot 2023-03-13 at 21 40 45

(the initializations in the constructor do not match the symbols)

james-elicx:
> @james-elicx just to continue the conversation from the other PR...
> 
> (the initializations in the constructor do not match the symbols)
> 
> Those strings not be strings but just the symbols right?

Yeah, that would explain why it's a string instead of the symbol in the console.log. Funnily enough, I looked at that exact code and completely skipped over the fact that it was setting the string instead of the symbol 😂

Nice spot. Even though that seems to be the cause of the problem, the fact that it was compiled like that is nothing short of bizarre. It really shouldn't have been. It's very unusual.

Looking at the code from the Vercel build output, `reviews.func/index.js` has the following:

image

The Webpack chunk that we extract in minification and write to a file has the following:

image

It appears that the `generate` function from `astring` is interpreting the `[c]=[]` as `c=[]` as if it were a string, which doesn't make sense.

james-elicx:
Putting the raw code from the build output into astexplorer.net, we can see that it treats it as `"c"` as well.

image

dario-piotrowicz:
> It appears that the generate function from astring is interpreting the [c]=[] as c=[] as if it were a string, which doesn't make sense. Or is the AST parser. I didn't check that part. Derp.
🤔

dario-piotrowicz:
I __think__ that acorn and astring may be working fine (based on this simplistic check: https://stackblitz.com/edit/typescript-gqntn3?file=index.ts) also this only happens when we use our experimentalMinify so it does feel like we must be messing something up there, no?

james-elicx:
Solution identified. Upgrading `astring` to 1.8.4... 😂

I copied the source code locally to play around with, but I realised it was working fine with their latest source code. Checked the `astring` versions. We're on 1.8.3 and the latest is 1.8.4. Upgraded. It works fine.

dario-piotrowicz:
> Solution identified. Upgrading `astring` to 1.8.4... 😂
> 
> I copied the source code locally to play around with, but I realised it was working fine with their latest source code. Checked the `astring` versions. We're on 1.8.3 and the latest is 1.8.4. Upgraded. It works fine.

tell me you're joking! 😭 🤣

james-elicx:
> > Solution identified. Upgrading `astring` to 1.8.4... 😂
> > I copied the source code locally to play around with, but I realised it was working fine with their latest source code. Checked the `astring` versions. We're on 1.8.3 and the latest is 1.8.4. Upgraded. It works fine.
> 
> tell me you're joking! 😭 🤣

Sadly not... All that time to just upgrade the depending hahaha. At least we found a solution now! 😂😂

Want me to upgrade it in this PR or would you rather do that at the same time as prettier etc.?

dario-piotrowicz:
I just checked `generate` is indeed only used when we have `experimentalMinify` on, so that explains that 😓 

sorry for the back and forth, since it's just bumping astring I think you can just do it in the other PR, so that we can land all the minify fixes there after all, would that work for you? 🙂 

(PS: regarding prettier thanks for the consideration, I'll just solve the lock files conflicts when I get to merge that 🙂👍 )

james-elicx:
> I just checked `generate` is indeed only used when we have `experimentalMinify` on, so that explains that 😓
> 
> sorry for the back and forth, since it's just bumping astring I think you can just do it in the other PR, so that we can land all the minify fixes there after all, would that work for you? 🙂
> 
> (PS: regarding prettier thanks for the consideration, I'll just solve the lock files conflicts when I get to merge that 🙂👍 )

That was a fun dive into learning a bit more about how ASTs work. Sounds good, I'll add it to that one :)

PR #100, opened 2023-03-14
add prepare script so that package can be installed through git

### LINKED ISSUE #49: "npm run build" should be run on "prepare" event to allow direct github installs
Currently `package.json` seems to run `npm run build` as part of the `publish` hook. Instead, that part should be run in the `prepare` hook, to support cases where this repository is installed as a dependency directly from GitHub and not via a published packages.

See: https://docs.npmjs.com/cli/v9/using-npm/scripts#prepare-and-prepublish

dario-piotrowicz:
resolves #49


jculvey (approved):

dario-piotrowicz:
Merged the PR as it is doing what it is supposed to, if we'll have some issues with it we can always revert it

PR #106, opened 2023-03-17
feat: support route handlers

### LINKED ISSUE #105: App Dir Route Handlers aren't correctly handled
When a user created a Next app in app dir mode and they create come [route handlers](https://beta.nextjs.org/docs/routing/route-handlers) they end up with something like:
Screenshot 2023-03-16 at 23 08 21

We need to handle this situation which we currently are not.

<details>

As per the example above.

`functionsMap` is populated with the just `api/hello`

Screenshot 2023-03-16 at 23 10 07

Then in our check we have two issues:
 - the entry name is `app/api/hello/route` while we only have `api/hello` in the functionsMap (no `/route`)
 - the check assumes that we're always inside a `pages` directory which is not the case with api routes in app dir mode  

Screenshot 2023-03-16 at 23 12 49

</details>

james-elicx:
- Created a new utility function for matching manifest function entries with the file paths we obtain.
- Support route handlers.
- Added tests for the utility function.

fixes #105

### DIFF
--- a/src/utils/transform/manifest.ts
+++ b/src/utils/transform/manifest.ts
@@ -1,0 +1,17 @@
+/**
+ * Check if a function file name matches an entry in the manifest.
+ *
+ * @param entry Manifest entry name.
+ * @param file Function file name.
+ * @returns Whether the function file name matches the manifest entry name.
+ */
+export const matchFunctionEntry = (entry: string, file: string) => {
+  // app directory
+  if (entry.startsWith('app/')) {
+    const type = entry.endsWith('/route') ? '/route' : '/page';
+    return `app${file !== 'index' ? `/${file}` : ''}${type}` === entry;
+  }
+
+  // pages directory
+  return entry.startsWith('pages/') && `pages/${file}` === entry;
+};
--- a/src/utils/transform/manifest.ts
+++ b/src/utils/transform/manifest.ts
@@ -1,17 +1,19 @@
 /**
  * Check if a function file name matches an entry in the manifest.
  *
- * @param entry Manifest entry name.
- * @param file Function file name.
+ * @param entryName Manifest entry name.
+ * @param fileName Function file name.
  * @returns Whether the function file name matches the manifest entry name.
  */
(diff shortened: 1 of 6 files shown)


src/utils/transform/manifest.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Check if a function file name matches an entry in the manifest.
+ *
+ * @param entry Manifest entry name.
+ * @param file Function file name.
+ * @returns Whether the function file name matches the manifest entry name.
+ */
+export const matchFunctionEntry = (entry: string, file: string) => {
```
dario-piotrowicz, line 8: nit, maybe it would be clearer to include name in the params names?
```suggestion
export const matchFunctionEntry = (entryName: string, fileName: string) => {
```

tests/manifest.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  { name: 'app dir - index page', entry: 'app/page', file: 'index' },
+  { name: 'app dir - route handler', entry: 'app/test/route', file: 'test' },
+  { name: 'pages dir - page', entry: 'pages/api/hello', file: 'api/hello' },
+];
+
+suite('manifest utils', () => {
+  describe('match function entries', () => {
+    test.each(manifestEntries)('$name', ({ entry, file }) => {
+      expect(matchFunctionEntry(entry, file)).toEqual(true);
+    });
```
dario-piotrowicz, line 17: In testing frameworks I find these `test.each` utilities awkward (with then the fake string interpolation), I personally prefer to just do it manually instead:
```suggestion
    manifestEntries.forEach(({ name, entry, file }) => {
      test(name, () => {
        expect(matchFunctionEntry(entry, file)).toEqual(true);
      })
    })
```

looks much cleaner and clear to me (by a glance you can immediately understand what's going on)

do you feel strongly that `test.each` is better?

dario-piotrowicz: awesome! thanks you! 🙏 ❤️

PR #127, opened 2023-03-22
Index refactoring + adding unit test

dario-piotrowicz:


tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -0,0 +1,61 @@
...
+		expect(functionsMap.size).toEqual(expectedFunctions.length);
+		expectedFunctions.forEach(fn => {
+			expect(functionsMap.has(fn)).toBe(true);
+			const path = functionsMap.get(fn);
+			const expectedFnPathRegex = new RegExp(`/${fn}.func.js$`);
+			expect(path).toMatch(expectedFnPathRegex);
+		});
+	});
+
+	// TODO: add tests that also test the functions map with the experimentalMinify flag
```
dario-piotrowicz, line 49: I think this would be valuable but Igor suggested that we should look into the experimental minification and see if it can be improved, so maybe we can add this test/tests later when we decide what to do

tests/src/buildApplication/middlewareManifest.test.ts
```diff
@@ -0,0 +1,119 @@
...
+			],
+		]);
+
+		expect(await parseMiddlewareManifest(functionsMap)).deep.equals({
+			hydratedMiddleware: new Map(),
+			hydratedFunctions: expectedHydratedFunction,
+		});
+	});
+
+	// TODO: add a 'should produce appropriate hydratedMiddleware' test similar to the hydratedFunctions one
```
dario-piotrowicz, line 118: I think this would be valuable but all this logic revolves around the middleware manifest present in `.next` which we should refactor (so that we only use `.vercel/output`).

Things may look quite differently after we do that, so maybe it's not too worth investing the time in defining tests here? 🤔

src/buildApplication/generateFunctionsMap.ts
```diff
@@ -0,0 +1,314 @@
...
+ * Fixes the function contents in miscellaneous ways.
+ *
+ * Note: this function contains hacks which are quite brittle and should be improved ASAP.
+ *
+ * @param contents the original function's file contents
+ * @returns the updated/fixed contents
+ */
+function fixFunctionContents(contents: string) {
+	contents = contents.replace(
+		// TODO: This hack is not good. We should replace this with something less brittle ASAP
```
dario-piotrowicz, line 167: @GregBrimble here I just added a url to the source code, but I don't 100% remember the reasoning for this?

Could you remind me so that I can improve the comment?

vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	test: {
+		include: ['tests/**/*.test.ts'],
+		environment: 'edge-runtime',
```
dario-piotrowicz, line 6: I initially created two different config files one with `edge-runtime` to test the worker code and one with `nodejs` for the actual cli (and two different scripts), I then checked and the nodejs one wasn't really necessary (since in the tests I am mocking the node specific modules anyways) so I just figured that we could simplify things and just keep only this one. Please let me know if you think it's a bad idea.

petebacondarwin (commented):
A bunch of comments. Not quite done but I ran out of time before meetings kicked off today.

src/buildApplication/vercelConfig.ts
```diff
@@ -0,0 +1,47 @@
+import { readFile } from 'fs/promises';
+
+const supportedConfigVersion = 3;
+
+/**
+ * In src we don't make use of this object (just pass it along to the worker)
+ * for the full types of this object see: `templates/_worker.js/types.d.ts`
```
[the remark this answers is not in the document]
dario-piotrowicz: I think the `types.d.ts` file is used to provide type safety in the `templates/_worker.js/index.ts` file, we do have some logic there so I think that it is quite helpful to have some type safety when we need to touch that file, don't you think?

Regarding `in defining and exporting this` you mean the type in this file?

We do need to access its version, so I guess we'd want to specify a type (unless we want to use `unknown` and casting?)

Would something like this be preferrable?
```diff
export type VercelConfig = {
	version: typeof supportedConfigVersion;
-	routes?: unknown;
-	images?: unknown;
-	wildcard?: unknown;
-	overrides?: unknown;
-	cache?: string[];
-	crons?: unknown;
};
```
(this would work but I fear that it might be a bit misleading and give the impression that the version is the only thing included in the file?)

___

Also I wanted to add that I did put there a todo for adding validation so I guess later we might want the correct type here.

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -0,0 +1,61 @@
...
+		);
+		expect(invalidFunctions.size).toEqual(0);
+		const expectedFunctions = ['index', 'index.rsc', 'api/hello'];
+		expect(functionsMap.size).toEqual(expectedFunctions.length);
+		expectedFunctions.forEach(fn => {
+			expect(functionsMap.has(fn)).toBe(true);
+			const path = functionsMap.get(fn);
+			const expectedFnPathRegex = new RegExp(`/${fn}.func.js$`);
+			expect(path).toMatch(expectedFnPathRegex);
+		});
```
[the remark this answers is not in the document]
dario-piotrowicz: yeah I totally agree, I tend to prefer duplication in tests too 🙂
but in some cases yeah I kind of sneak DRY code in when I feel like things are getting too verbose (or sometimes it just happens as it is the way in which I code 😅), yeah in this case anyways as you said it doesn't really help, thanks I'll use your suggested code 🙂👍

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -0,0 +1,61 @@
...
+	});
+
+	// TODO: add tests that also test the functions map with the experimentalMinify flag
+
+	test('should return invalid functions', async () => {
```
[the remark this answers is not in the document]
dario-piotrowicz: Yes, that's exactly it, it just tests that the runtime is the correct one.

I could add that detail in the test description (or in a comment) but I think this might (or might not) change in the future (there might be other aspects of a function that can make it invalid) so I wouldn't really want to go into details about it if I can help it :thinking: 

What do you think? do you have any idea/preference on it?

src/buildApplication/nextConfigs/index.ts
```diff
@@ -0,0 +1,18 @@
...
+	basePath?: string;
+};
+
+/**
+ * generates an object containing the next.config.js options that next-on-pages
+ * supports
+ *
+ * For the list of supported options see: __TODO: add link to support.md section after readme PR is merged__
+ */
+export async function getNextConfigs(): Promise<NextConfigs> {
```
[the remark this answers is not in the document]
dario-piotrowicz: Yeah you're right, sorry about it, I'll check the code and try to use `NextJs` instead of `Next` 🙇 

PS: I am not sure if I prefer `Nextjs` (as a single term) 🤔, I guess you prefer `NextJs`?

src/buildApplication/middlewareManifest.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+export async function parseMiddlewareManifest(
+	functionsMap: Map<string, string>
+) {
+	let middlewareManifest: MiddlewareManifest;
+	try {
+		// Annoying that we don't get this from the `.vercel` directory.
+		// Maybe we eventually just construct something similar from the `.vercel/output/functions` directory with the same magic filename/precendence rules?
+		middlewareManifest = JSON.parse(
+			await readFile('.next/server/middleware-manifest.json', 'utf8')
```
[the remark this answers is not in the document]
dario-piotrowicz: is this along the lines of what you head in mind? 🙂 

https://github.com/cloudflare/next-on-pages/pull/127/commits/a71ea2498e68af59808b6a4879f3877fa40a924d

(PS: I am trying to keep things in the middlewareManifest as isolated as I can since we should refactor its logic at some point, that's why I kept the file reading in this file)

src/cli.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * prepares a message for Cli printing (simple prefixes each line with the appropriate prefix)
+ *
+ * the function also removes leading tabs from each line, this allows us to indent the messages
+ * in the code as we please (for actual indentation in the message we can use spaces instead)
+ */
+function prepareCliMessage(message: string, fromVercelCli: boolean): string {
+	return message
+		.split('\n')
+		.map(line => line.replace(/^\t+/, ''))
```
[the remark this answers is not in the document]
dario-piotrowicz: sure, I've added that 🙂:+1:

(the package doesn't have dependencies and if we'll want to remove it later on that would be very easy so I am totally ok on using it 🙂 (it is also more robust that the quick solution I added which requires the dev to know/remember how it works))

tests/src/buildApplication/middlewareManifest.test.ts
```diff
@@ -63,7 +61,7 @@ describe('parseMiddlewareManifest', () => {
 					matchers: [{ regexp: 'regexpF' }],
 				},
 			},
-		};
+		} as unknown as MiddlewareManifest;
```
petebacondarwin, line 64: Out of curiosity, why the need for the double cast here?
Is this object not a valid `MiddlewareManifest` by itself?

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -36,14 +36,10 @@ describe('generateFunctionsMap', async () => {
...
-		expectedFunctions.forEach(fn => {
-			expect(functionsMap.has(fn)).toBe(true);
-			const path = functionsMap.get(fn);
-			const expectedFnPathRegex = new RegExp(`/${fn}.func.js$`);
-			expect(path).toMatch(expectedFnPathRegex);
-		});
+		expect(functionsMap.size).toEqual(3);
+		expect(functionsMap.get('index')).toMatch(/\/index.func.js$/);
+		expect(functionsMap.get('index.rsc')).toMatch(/\/index.rsc.func.js$/);
+		expect(functionsMap.get('api/hello')).toMatch(/\/api\/hello.func.js$/);
```
petebacondarwin, line 42: I'm being pedantic now 😄 

```suggestion
		expect(functionsMap.get('index')).toMatch(/\/index\.func\.js$/);
		expect(functionsMap.get('index.rsc')).toMatch(/\/index\.rsc\.func\.js$/);
		expect(functionsMap.get('api/hello')).toMatch(/\/api\/hello\.func\.js$/);
```

Right?

petebacondarwin (approved):
LGTM - thanks

tests/src/buildApplication/generateFunctionsMap.test.ts
dario-piotrowicz: yeah not pedantic at all, sorry I missed those 😓 
Thanks a lot for noticing them! 🙏

tests/src/buildApplication/middlewareManifest.test.ts
dario-piotrowicz: We need it because we don't provide all the fields of the functions:
Screenshot 2023-03-24 at 09 52 45
So typescript doesn't seem to accept a single cast 😕

Alternatively if I did a `DeepPartial` we could reduce the double cast into a simple one:
Screenshot 2023-03-24 at 09 58 01

Or I could provide all the values, probably with utility otherwise things would get quite verbose and not having any casting, something like:
Screenshot 2023-03-24 at 10 08 37

Do you have any preference? (or a different option?)

petebacondarwin: Do we really care about the accuracy of the `MiddlewareManifest` interface, since all we do is convert such an object to hydrated middleware? Could we make this stuff partial throughout?

But anyway, I don't mind either way, was just interested. Not a blocker from merging.

dario-piotrowicz: mh... I made fields in the MiddlewareManifest type partial 👍

Not sure if that's the correct way, but either way all this MiddlewareManifest related code should hopefully be heavily refactored at some point soon (since it relies on the `.next` directory which we shouldn't rely on), so I don't really mind

src/buildApplication/buildApplication.ts
```diff
@@ -3,6 +3,7 @@ import { exit } from 'process';
 import { resolve } from 'path';
 import { cliError, cliLog, CliOptions } from '../cli';
 import {
+	getParsedMiddlewareManifest,
 	MiddlewareManifestData,
 	parseMiddlewareManifest,
```
petebacondarwin, line 8: Don't forget to remove this!

jculvey (approved):

PR #134, opened 2023-03-29
feat: utilities for file-system operations + minor improvements to `fs/promises` mocks

james-elicx:
This is somewhat of a precursor to the new routing system build-time PR. It makes the following changes:
- Adds a VS Code setting for the prettier config path to fix the prettier extension not finding a `.prettierrc` file.
- Introduces some utility functions for file-system operations to dedupe try-catch code.
   - `validatePathType` - check a file exists at a path, and verify that it is either a directory or a file.
      - `validateFile`
      - `validateDir`
   - `readJsonFile` - generic function to read a file at a path and parse it to JSON. If it does not exist or fails, it returns null.
- Improves some of the typing to work better with the JSONValue type.
- Adds tests for the utility functions.
- Improves existing mocks
   - Puts them inside `beforeAll` with cleanups in `afterAll`.
   - Normalises paths where relevant to work when running tests on Windows.
   - Adapts to return better response for `stat` where relevant.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"prettier.configPath": ".prettierrc.js"
```
dario-piotrowicz, line 2: I'm really not sure why this is needed 😕 , did your vscode didn't find the config by itself?

(
as far as I can tell prettier should work fine without any sort of path configuration? https://prettier.io/docs/en/configuration.html )

src/utils/fs.ts
```diff
@@ -0,0 +1,65 @@
...
+	return path.startsWith('\\\\?\\') ? path : path.replace(/\\/g, '/');
+}
+
+/**
+ * Read and parse a JSON file.
+ *
+ * @param path File path to try and parse as JSON.
+ * @returns Parsed JSON file.
+ */
+export async function readJsonFile<T = object>(path: string) {
```
dario-piotrowicz, line 33: Instead of object and this sort of generic typings what would you think of using a `JSONValue` type like this? https://dev.to/ankittanna/how-to-create-a-type-for-complex-json-object-in-typescript-d81

.vscode/settings.json
james-elicx: If one has a global VS Code setting for the prettier config, this will override that. Not that I recall ever actually setting it, but I happen to have a global setting to use `.prettierrc` for the config path, so for me adding a project-scoped settings file will override the global setting I have.

src/utils/fs.ts
james-elicx: It can be changed from `object` to something more inclusive like in that article if you would like, sure. My intention behind it being generic is that it overrides the type inside this function, instead of doing `as ...` each time we read a json file, so I would like to keep the generic part. If that's okay with you, that is?

In the future, the generic could ideally be changed to use the inferred type from a validator function passed in if you opt to add file validation at some point.

dario-piotrowicz: Ah no yeah you're right, anyways sorry I wasn't suggesting to make the function not generic (although yeah my comment can make it look like that's what I meant 😓) but to do something like:
```
export async function readJsonFile<T extends JSONValue>(path: string) {
```
to make it less generic in the sense that we restrict what the type can be (so that for example we get compilation errors if we try to get something not JSON serializable like a date)

It's no big deal either way and as you mentioned if we had validation we could make it even quite specific, so I am fine with whatever here 🙂

src/buildApplication/generateFunctionsMap.ts
```diff
@@ -85,21 +84,25 @@ async function processDirectoryRecursively(
 	};
 }
 
+type FunctionConfig = {
+	runtime: 'edge';
```
dario-piotrowicz, line 88: I don't think this is right, we want to have the `edge` runtime but this can also be `nodejs`, and we do check for that in the code:
```
if (functionConfig?.runtime !== 'edge') {
```
(I'm actually surprised that typescript is not complaining here 😕)

(anyways what I am saying is that I think that your type is a bit misleading and for example makes that if statement look incorrect since you're saying the the runtime is always `edge`)

```suggestion
	runtime: 'edge'|'nodejs';
```

.vscode/settings.json
dario-piotrowicz: that definitely doesn't sound right 😕 (https://github.com/prettier/prettier-vscode#configuration)

I'm pretty sure that the local config should get picked up before the global one and that this shouldn't be necessary

anyways I am not sure if there are downsides in setting this so I guess we can just add it if it solves issues you're having locally

src/buildApplication/generateFunctionsMap.ts
james-elicx: Hey man, I just copied the type that was already used and added the additional properties to it for future usage, haha. Fixed.

There wouldn't be a listing complaint since runtime can technically be undefined if functionConfig is null, so it would still be possible to not equal 'edge'.

.vscode/settings.json
james-elicx: I thought the same thing, but my VS Code global settings are meaning it is only trying to find `.prettierrc` instead of `.prettierrc.js`, so the project-scoped VS Code settings file overrides that global setting I have.

There aren't any downsides to specifying this in a project-scoped settings file - it just means that the setting takes precedence over any global settings.

image

src/utils/fs.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Check that the path exists and is of the expected type.
+ *
+ * @param path Path to check.
+ * @param type Whether to check for a `file` or `directory`.
+ * @returns Boolean representing whether the path matched the expected type.
+ */
+export async function validatePathType(
+	path: string,
+	type: 'file' | 'directory'
+) {
```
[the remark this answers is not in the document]
james-elicx: I quite like that idea.

src/buildApplication/generateFunctionsMap.ts
dario-piotrowicz: > There wouldn't be a listing complaint since runtime can technically be undefined if functionConfig is null, so it would still be possible to not equal 'edge'.

ahhh yeah, good point 😁

src/utils/fs.ts
```diff
@@ -0,0 +1,90 @@
...
+	return false;
+}
+
+/**
+ * Check that the path exists and that it is a file.
+ *
+ * @param path Path to check.
+ * @returns Whether a file exists at the given path.
+ */
+export async function validateFile(path: string) {
```
dario-piotrowicz, line 78:
```suggestion
export function validateFile(path: string) {
```

src/utils/fs.ts
```diff
@@ -0,0 +1,90 @@
...
+	return validatePathType(path, 'file');
+}
+
+/**
+ * Check that the path exists and that it is a directory.
+ *
+ * @param path Path to check.
+ * @returns Whether a directory exists at the given path.
+ */
+export async function validateDir(path: string) {
```
dario-piotrowicz, line 88:
```suggestion
export function validateDir(path: string) {
```

dario-piotrowicz (approved):
Awesome, thanks a lot! 😃 

I've just left one very last couple minor suggestions please have a look, besides that I think this should be good to go 😄

PR #136, opened 2023-04-01
feat: remove private/telemetry files from the build output

james-elicx:
Vercel and Next.js generate *private* files for telemetry purposes that are accessible as static assets.

`.vercel/output/static/_next/__private/trace` includes a full output of all info accrued. It also has full path names, which may pose security concerns if accessible (depending on where you are building).

While the new routing system *should* prevent these files from being accessible (if they have a record in the outputted config), they would still be accessible if someone modified an `_routes.json` file to exclude them from invocations.

We do not need these files, nor do we want to run the risk of them being available. Therefore, we should purge them instead of letting them be uploaded to Cloudflare Pages.

### DIFF
--- a/src/buildApplication/buildApplication.ts
+++ b/src/buildApplication/buildApplication.ts
@@ -9,7 +9,7 @@ import { getNextJsConfigs } from './nextJsConfigs';
 import { getVercelConfig } from './getVercelConfig';
 import { buildWorkerFile } from './buildWorkerFile';
 import { generateFunctionsMap } from './generateFunctionsMap';
-import { buildVercelOutput } from './buildVercelOutput';
+import { buildVercelOutput, purgePrivateFiles } from './buildVercelOutput';
 import { buildMetadataFiles } from './buildMetadataFiles';
 import { validateDir } from '../utils';
 
@@ -25,6 +25,8 @@ export async function buildApplication({
 	if (!skipBuild) {
 		await buildVercelOutput();
 	}
+
+	await purgePrivateFiles();
 
 	await prepareAndBuildWorker({ experimentalMinify });
 	await buildMetadataFiles();
--- a/src/buildApplication/buildApplication.ts
+++ b/src/buildApplication/buildApplication.ts
@@ -9,7 +9,10 @@ import { getNextJsConfigs } from './nextJsConfigs';
 import { getVercelConfig } from './getVercelConfig';
 import { buildWorkerFile } from './buildWorkerFile';
 import { generateFunctionsMap } from './generateFunctionsMap';
-import { buildVercelOutput, purgePrivateFiles } from './buildVercelOutput';
+import {
+	buildVercelOutput,
+	deleteNextTelemetryFiles,
+} from './buildVercelOutput';
 import { buildMetadataFiles } from './buildMetadataFiles';
 import { validateDir } from '../utils';
 
@@ -26,7 +29,7 @@ export async function buildApplication({
 		await buildVercelOutput();
 	}
 
-	await purgePrivateFiles();
+	await deleteNextTelemetryFiles();
 
 	await prepareAndBuildWorker({ experimentalMinify });
 	await buildMetadataFiles();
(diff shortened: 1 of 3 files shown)


src/buildApplication/buildApplication.ts
```diff
@@ -26,6 +26,8 @@ export async function buildApplication({
 		await buildVercelOutput();
 	}
 
+	await purgePrivateFiles();
```
dario-piotrowicz, line 29: sorry for being picky, but could we use `delete` instead of `purge`, `purge` might be ok but just a bit more ambiguous/less straightforward than `delete`?

I'd also make more clear what this files are, something along the lines of `deleteNextTelemetryFiles`

if I read such name I'd immediately get the gist of it, if instead I read `purgePrivateFiles` I'd need to look at the code (or the tsdoc) to understand what it is about.

what do you think?

src/buildApplication/buildVercelOutput.ts
```diff
@@ -60,3 +61,24 @@ async function runVercelBuild(): Promise<void> {
...
+	const nextDir = resolve('.vercel/output/static/_next');
+	const privateDir = join(nextDir, '__private');
+
+	if (await validateDir(privateDir)) {
+		await rm(privateDir, { recursive: true, force: true });
+
+		// Remove the `_next` directory if it's now empty
+		if ((await readdir(nextDir)).length === 0) {
+			await rm(nextDir, { recursive: true, force: true });
+		}
```
dario-piotrowicz, line 82: is this needed? would the empty dir there cause any issue?

dario-piotrowicz: (also instead of `readdir` + `rm` I think you could use [`rmdir`](https://www.geeksforgeeks.org/node-js-fs-rmdir-method/) with a try catch (since it throws if the directory is not empty)?)

dario-piotrowicz (commented):
I've added a couple of comments/suggestions, I think this looks great! thanks @james-elicx! :slightly_smiling_face: 

One thing that I wanted to ask you however is, instead of deleting this directory have you checked if there is a way not to get this file created in the first place?

I have tried the out-out methods from the [next.js docs](https://nextjs.org/telemetry) since I think that this telemetry info is create by next and the handled by the vercel cli?
But I haven't had much luck and the file seems to get created anyways.

(I am not sure if it is possible but I'd definitely think it would be nicer if we could get Next/Vercel not to create this in the first place)

src/buildApplication/buildVercelOutput.ts
james-elicx: > is this needed? would the empty dir there cause any issue?

The empty directory wouldn't cause issues, but it's entirely unnecessary to keep if it's empty.

> (also instead of `readdir` + `rm` I think you could use [`rmdir`](https://www.geeksforgeeks.org/node-js-fs-rmdir-method/) with a try catch (since it throws if the directory is not empty)?)

Good call.

james-elicx:
> One thing that I wanted to ask you however is, instead of deleting this directory have you checked if there is a way not to get this file created in the first place?
>
> I have tried the out-out methods from the [next.js docs](https://nextjs.org/telemetry) since I think that this telemetry info is create by next and the handled by the vercel cli? But I haven't had much luck and the file seems to get created anyways.
> 
> (I am not sure if it is possible but I'd definitely think it would be nicer if we could get Next/Vercel not to create this in the first place)

I have not, but there is a good reason for that. If someone skips the build process (`--skip-build`) when using this tool, we still want to delete the file(s). So for that reason, I believe it would be better to handle it irrespective of any settings that impact its output at build time.

As we offer the option for people to build themselves and skip us triggering to build, I do think it's important to do it like this.

dario-piotrowicz (approved):

dario-piotrowicz:
> I have not, but there is a good reason for that. If someone skips the build process (`--skip-build`) when using this tool, we still want to delete the file(s). So for that reason, I believe it would be better to handle it irrespective of any settings that impact its output at build time.

yeah good point :slightly_smiling_face::+1:

PR #138, opened 2023-04-02
add next-on-pages version to CLI

dario-piotrowicz:
Adding the next-on-pages version to the CLI messages, I think it's going to be convenient for two reasons:
 - if we switch between different versions for example for testing things this can help double checking what the current version is
 - when people open issues and provide the CLI output the version might get included which would be helpful
 
Results:
Screenshot at 2023-04-02 17-38-02

Screenshot at 2023-04-02 17-38-26

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import { watch } from 'chokidar';
 import pLimit from 'p-limit';
 import { cliLog, CliOptions, getCliOptions, printCliHelpMessage } from './cli';
 import { buildApplication } from './buildApplication';
+import { nextOnPagesVersion } from './utils';
 
 const limit = pLimit(1);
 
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -2,22 +2,27 @@ import { watch } from 'chokidar';
...
 const limit = pLimit(1);
 
 const cliOptions = getCliOptions();
 runNextOnPages(cliOptions);
 
 function runNextOnPages(options: CliOptions): void {
-	cliLog('@cloudflare/next-on-pages CLI');
+	if (options.version) {
+		// eslint-disable-next-line no-console -- for the version lets simply print it plainly
+		console.log(nextOnPagesVersion);
```
dario-piotrowicz, line 15: this is not really super necessary but I didn't use `cliLog` because that adds the `⚡️` prefix which feels weird to me when showing the version

src/index.ts
```diff
@@ -2,22 +2,27 @@ import { watch } from 'chokidar';
...
+		return;
+	}
+
+	cliLog(`@cloudflare/next-on-pages CLI v.${nextOnPagesVersion}`);
```
dario-piotrowicz, line 19: I think it's much nicer to just show the version with `--version` so I moved this log below it (also this log shows the version already so it would be a bit awkward to include it)

jculvey (approved):

src/index.ts
```diff
@@ -2,22 +2,27 @@ import { watch } from 'chokidar';
...
 
 	if (options.help) {
 		printCliHelpMessage();
 		return;
 	}
 
-	runBuild(options);
```
jculvey, line 19: Was this intentional?

dario-piotrowicz: no 😓 😭

PR #139, opened 2023-04-02
allow setting of NODE_ENV (also fix bug that always enabled the experimental minification)

### LINKED ISSUE #4: Debug mode
Support `NODE_ENV` set to `development` to enable debug logging in the CLI and in the Worker itself (if not done with a `--debug` argument or similar).

dario-piotrowicz:
resolves #4

### DIFF
--- a/src/buildApplication/buildWorkerFile.ts
+++ b/src/buildApplication/buildWorkerFile.ts
@@ -2,15 +2,17 @@ import { writeFile } from 'fs/promises';
 import { join } from 'path';
 import { build } from 'esbuild';
 import { tmpdir } from 'os';
-import { cliLog, CliOptions } from '../cli';
+import { cliLog, NodeEnv } from '../cli';
 import { NextJsConfigs } from './nextJsConfigs';
 import { MiddlewareManifestData } from './middlewareManifest';
+import { generateGlobalJs } from './buildGlobalContext';
 
 export async function buildWorkerFile(
 	{ hydratedMiddleware, hydratedFunctions }: MiddlewareManifestData,
 	vercelConfig: VercelConfig,
 	nextJsConfigs: NextJsConfigs,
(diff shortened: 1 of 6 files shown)


src/cli.ts
```diff
@@ -20,9 +25,32 @@ export function getCliOptions(): CliOptions {
 		watch: process.argv.includes('--watch'),
 		skipBuild: process.argv.includes('--skip-build'),
 		experimentalMinify: process.argv.includes('--experimental-minify'),
+		nodEnv: getNodeEnv(),
 	};
 }
 
+function getNodeEnv(): NodeEnv {
```
dario-piotrowicz, line 32: I implemented this in a naive way, just checking if there is a `--node-env` argument and checking the argument next to it

so if for example someone does `--node-env=development` it's not going to work

I didn't think it was worth spending much time on it since we should soon-ish switch to an automated arguments parser such as yargs that will handle all of this for us anyways.

(@GregBrimble should we consider using your [zodcli](https://github.com/GregBrimble/zodcli)? :slightly_smiling_face:)

src/buildApplication/buildWorkerFile.ts
```diff
@@ -2,15 +2,17 @@ import { writeFile } from 'fs/promises';
...
 import { NextJsConfigs } from './nextJsConfigs';
 import { MiddlewareManifestData } from './middlewareManifest';
+import { generateGlobalJs } from './buildGlobalContext';
 
 export async function buildWorkerFile(
 	{ hydratedMiddleware, hydratedFunctions }: MiddlewareManifestData,
 	vercelConfig: VercelConfig,
 	nextJsConfigs: NextJsConfigs,
-	experimentalMinify: Pick<CliOptions, 'experimentalMinify'>
+	experimentalMinify: boolean,
```
dario-piotrowicz, line 14: I was passing an object here and treat is as a boolean, so it was always truthy... sorry about that :facepalm:

PR #140, opened 2023-04-03
fix experimental minification always on

dario-piotrowicz:
fixing a stupid mistake I made :facepalm:

### DIFF
--- a/src/buildApplication/buildWorkerFile.ts
+++ b/src/buildApplication/buildWorkerFile.ts
@@ -2,7 +2,7 @@ import { writeFile } from 'fs/promises';
 import { join } from 'path';
 import { build } from 'esbuild';
 import { tmpdir } from 'os';
-import { cliLog, CliOptions } from '../cli';
+import { cliLog } from '../cli';
 import { NextJsConfigs } from './nextJsConfigs';
 import { MiddlewareManifestData } from './middlewareManifest';
 import { generateGlobalJs } from './generateGlobalJs';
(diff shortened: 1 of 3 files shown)


src/buildApplication/buildWorkerFile.ts
```diff
@@ -2,15 +2,15 @@ import { writeFile } from 'fs/promises';
...
+import { cliLog } from '../cli';
 import { NextJsConfigs } from './nextJsConfigs';
 import { MiddlewareManifestData } from './middlewareManifest';
 
 export async function buildWorkerFile(
 	{ hydratedMiddleware, hydratedFunctions }: MiddlewareManifestData,
 	vercelConfig: VercelConfig,
 	nextJsConfigs: NextJsConfigs,
-	experimentalMinify: Pick<CliOptions, 'experimentalMinify'>
+	experimentalMinify: boolean
```
GregBrimble, line 13: You could, if you wanted to still reference the `CliOptions`, do something like

```suggestion
	experimentalMinify: CliOptions['experimentalMinify']
```

But I'm 50/50 on actually that. Dealer's choice.

GregBrimble (approved):

src/buildApplication/buildWorkerFile.ts
dario-piotrowicz: yeah I thought of that as well, but I figured that I don't really want to have this `CliOptions` in too many places if not necessary 🤔 (since it's an extra import, a slightly more complex type, etc... and the typesafety is equivalent)

(well, in this case it could indicate more clearly where this value comes from.... 🤔 🤷 )

I also don't have the strongest opinion on this one, since you're 50/50 I think I might just leave it as boolean 🙂 🤔  🤷

PR #141, opened 2023-04-03
allow setting of NODE_ENV

### LINKED ISSUE #4: Debug mode

dario-piotrowicz:
resolves #4

### DIFF
--- a/.changeset/gentle-melons-turn.md
+++ b/.changeset/gentle-melons-turn.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+allow setting of NODE_ENV
--- a/src/buildApplication/buildWorkerFile.ts
+++ b/src/buildApplication/buildWorkerFile.ts
@@ -5,6 +5,7 @@ import { tmpdir } from 'os';
 import { cliLog, CliOptions } from '../cli';
 import { NextJsConfigs } from './nextJsConfigs';
 import { MiddlewareManifestData } from './middlewareManifest';
+import { generateGlobalJs } from './generateGlobalJs';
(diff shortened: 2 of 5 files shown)


src/buildApplication/buildGlobalContext.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @returns the plain javascript string that should be added at the top of the the _worker.js file
+ */
+export function generateGlobalJs(): string {
+	const nodeEnv = getNodeEnv();
+	return `globalThis.process = { env: { NODE_ENV: '${nodeEnv}' } };`;
+}
+
+function getNodeEnv(): string {
+	const nextJsNodeEnvs = ['production', 'development', 'test'];
+	const defaultNodeEnv = nextJsNodeEnvs[0];
```
GregBrimble, line 16: Nit: to rely on the ordering of this array is pretty brittle, imo. I'd personally be explicit, or better yet, use an enum.

```suggestion
	const defaultNodeEnv = 'production';
```

GregBrimble (approved):

src/buildApplication/buildGlobalContext.ts
dario-piotrowicz: I've added a unit test which makes sure that the default value is `production` does that help? 🙂

dario-piotrowicz: I've added a string enum anyways 🙂👍

GregBrimble (approved):
✅✅

Issue #52, opened 2023-01-05
Error: spawn npm ENOENT


szymondlugolecki:
When trying to run `npx @cloudflare/next-on-pages` I get an error

```
C:\Users\...\...\project>npx @cloudflare/next-on-pages  
@cloudflare/next-to-pages CLI
⚡️
⚡️ Installing 'vercel' CLI...
⚡️
node:events:368
      throw er; // Unhandled 'error' event
      ^

Error: spawn npm ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
    at onErrorNT (node:internal/child_process:477:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (node:internal/child_process:288:12)
    at onErrorNT (node:internal/child_process:477:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn npm',
  path: 'npm',
  spawnargs: [ 'install', '-D', 'vercel' ]
}
```

I use
node@16.13.0 and @18.12.0
npm@8.1.0

LOLBRUHNICE:
I have already created a pr that can fix this issue, waiting for merge, ref https://github.com/cloudflare/next-on-pages/pull/34

fineshop:
How can we fix this issue on windows?

gvatsov:
> How can we fix this issue on windows?

I can say that using the @LOLBRUHNICE changes from the quoted PR are working for me on Windows 11.

surjithctly:
FYI: I'm still getting same error. Looks like its not fixed. Using latest version. cc: @LOLBRUHNICE @dario-piotrowicz 

<img width="1229" alt="CleanShot 2023-03-22 at 21 32 26@2x" src="https://user-images.githubusercontent.com/1884712/226965930-30e6293a-e111-4616-94dd-1d3320edf01e.png">

dario-piotrowicz:
Hi @surjithctly are you using `pnpm`?

if so can you delete all `pnpm` files and try again (just using `npx`)?

Are you on windows?

surjithctly:
Yes. I’m using PNPM.

This issue occurs on cloudflare pages build (not local). I’m on Mac.

dario-piotrowicz:
[...]

PR #145, opened 2023-04-05
Support utilizing the package manager used in the project

LOLBRUHNICE:
This pr is created to add support for next-on-pages to utilize the package manager used in the project,
if user using:
- npm - use npx to build as before
- pnpm - use pnpx to build
- yarn (classic) - install vercel first(since yarn classic doesn't have dlx) and build
- yarn (berry) - use yarn dlx to build
> Only allow to run with yarn berry with the nodeLinker was set to node-modules, because vercel doesn't support Plug'n'Play

This pr is backward compatible for the next-on-pages projects which ware created before.

### DIFF
--- a/.changeset/nice-eyes-flash.md
+++ b/.changeset/nice-eyes-flash.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Added support to utilize the package manager used in the project
--- a/src/buildApplication/buildVercelOutput.ts
+++ b/src/buildApplication/buildVercelOutput.ts
@@ -3,6 +3,7 @@ import { spawn } from 'child_process';
 import { join, resolve } from 'path';
 import { cliError, cliLog } from '../cli';
 import { validateDir, validateFile } from '../utils';
+import { checkPackageManager } from './checkPackageManager';
 
 /**
  * Builds the Next.js output via the Vercel CLI
@@ -16,18 +17,13 @@ import { validateDir, validateFile } from '../utils';
  *
  */
 export async function buildVercelOutput(): Promise<void> {
-	cliLog(`
-		Preparing project for 'npx vercel build'...
-	`);
+	const pkgMng = await checkPackageManager();
+	cliLog(`Detected Package Manager: ${pkgMng}\n`);
+	cliLog(`Preparing project...`);
 	await generateProjectJsonFileIfNeeded();
-	cliLog(
-		`Project ready for 'npx vercel build'...
-
-		Building project with 'npx vercel build'...
-	`
-	);
-	await runVercelBuild();
-	cliLog('Completed `npx vercel build`.\n');
+	cliLog(`Project is ready`);
+	await runVercelBuild(pkgMng);
+	cliLog('Building Completed.\n');
 }
 
 /**
@@ -44,9 +40,48 @@ async function generateProjectJsonFileIfNeeded(): Promise<void> {
 	}
 }
 
-async function runVercelBuild(): Promise<void> {
-	const npx = process.platform === 'win32' ? 'npx.cmd' : 'npx';
-	const vercelBuild = spawn(npx, ['vercel', 'build']);
+async function runVercelBuild(
+	pkgMng: 'pnpm' | 'yarn (berry)' | 'yarn (classic)' | 'npm'
+): Promise<void> {
+	cliLog(`Building project...`);
+
+	let cmd = '',
+		args = [];
+	const winCMD = process.platform === 'win32' ? '.cmd' : '';
+
+	if (pkgMng === 'yarn (classic)') {
+		cliLog(
+			`Installing vercel as dev dependencies with 'yarn add vercel -D'...`
+		);
+
+		const installVercel = spawn(`yarn${winCMD}`, ['add', 'vercel', '-D']);
+
+		installVercel.stdout.on('data', data => cliLog(`\n${data}`, true));
+		installVercel.stderr.on('data', data => cliError(`\n${data}`, false, true));
+
+		await new Promise((resolve, reject) => {
+			installVercel.on('close', code => {
+				if (code === 0) {
+					resolve(null);
+				} else {
+					reject();
+				}
+			});
+		});
+
+		cliLog(`Install completed`);
+
+		(cmd = `yarn${winCMD}`), (args = ['vercel', 'build']);
+	}
+
+	if (pkgMng === 'pnpm') (cmd = `pnpx${winCMD}`), (args = ['vercel', 'build']);
+	if (pkgMng === 'npm') (cmd = `npx${winCMD}`), (args = ['vercel', 'build']);
+	if (pkgMng === 'yarn (berry)')
+		(cmd = `yarn${winCMD}`), (args = ['dlx', 'vercel', 'build']);
+
+	cliLog(`Building project with '${cmd} ${args.join(' ')}'...`);
+
+	const vercelBuild = spawn(cmd, args);
 
 	vercelBuild.stdout.on('data', data => cliLog(`\n${data}`, true));
 	vercelBuild.stderr.on('data', data => cliError(`\n${data}`, false, true));
--- a/src/buildApplication/checkPackageManager.ts
+++ b/src/buildApplication/checkPackageManager.ts
@@ -1,0 +1,50 @@
+import YAML from 'js-yaml';
+import { spawn } from 'child_process';
+import { readFile } from 'fs/promises';
+import { validateFile } from '../utils';
+import { cliError } from '../cli';
+
+export async function checkPackageManager() {
+	const userAgent = process.env.npm_config_user_agent;
+
+	const hasYarnLock = await validateFile('yarn.lock');
+	const hasPnpmLock = await validateFile('pnpm-lock.yaml');
+
+	if ((userAgent && userAgent.startsWith('pnpm')) || hasPnpmLock) return 'pnpm';
+
+	if ((userAgent && userAgent.startsWith('yarn')) || hasYarnLock) {
+		const yarn = process.platform === 'win32' ? 'yarn.cmd' : 'yarn';
+		const getYarnV = spawn(yarn, ['-v']);
+		let yarnV = '';
+		getYarnV.stdout.on('data', data => {
+			yarnV = `${data}`.trimEnd();
+		});
+		getYarnV.stderr.on('data', data => {
+			cliError(data);
+		});
+		await new Promise((resolve, reject) => {
+			getYarnV.on('close', code => {
+				if (code === 0) {
+					resolve(null);
+				} else {
+					reject();
+				}
+			});
+		});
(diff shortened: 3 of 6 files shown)


src/buildApplication/buildVercelOutput.ts
```diff
@@ -16,18 +17,13 @@ import { validateDir, validateFile } from '../utils';
  *
  */
 export async function buildVercelOutput(): Promise<void> {
-	cliLog(`
-		Preparing project for 'npx vercel build'...
-	`);
+	const pkgMng = await checkPackageManager();
+	cliLog(`Detected Package Manager: ${pkgMng}\n`);
+	cliLog(`Preparing project...`);
```
dario-piotrowicz, line 22:
```suggestion
	cliLog('Preparing project...');
```

src/buildApplication/buildVercelOutput.ts
```diff
@@ -16,18 +17,13 @@ import { validateDir, validateFile } from '../utils';
...
 	await generateProjectJsonFileIfNeeded();
-	cliLog(
-		`Project ready for 'npx vercel build'...
-
-		Building project with 'npx vercel build'...
-	`
-	);
-	await runVercelBuild();
-	cliLog('Completed `npx vercel build`.\n');
+	cliLog(`Project is ready`);
```
dario-piotrowicz, line 24:
```suggestion
	cliLog('Project is ready');
```

src/buildApplication/buildVercelOutput.ts
```diff
@@ -44,9 +40,48 @@ async function generateProjectJsonFileIfNeeded(): Promise<void> {
 	}
 }
 
-async function runVercelBuild(): Promise<void> {
-	const npx = process.platform === 'win32' ? 'npx.cmd' : 'npx';
-	const vercelBuild = spawn(npx, ['vercel', 'build']);
+async function runVercelBuild(
+	pkgMng: 'pnpm' | 'yarn (berry)' | 'yarn (classic)' | 'npm'
+): Promise<void> {
+	cliLog(`Building project...`);
```
dario-piotrowicz, line 46:
```suggestion
	cliLog('Building project...');
```

src/buildApplication/buildVercelOutput.ts
```diff
@@ -44,9 +40,48 @@ async function generateProjectJsonFileIfNeeded(): Promise<void> {
...
+	}
+
+	if (pkgMng === 'pnpm') (cmd = `pnpx${winCMD}`), (args = ['vercel', 'build']);
+	if (pkgMng === 'npm') (cmd = `npx${winCMD}`), (args = ['vercel', 'build']);
+	if (pkgMng === 'yarn (berry)')
+		(cmd = `yarn${winCMD}`), (args = ['dlx', 'vercel', 'build']);
+
+	cliLog(`Building project with '${cmd} ${args.join(' ')}'...`);
+
+	const vercelBuild = spawn(cmd, args);
```
dario-piotrowicz, line 84: I feel like having `cmd` and `args` makes this much less readable and more complicated than needed (you than have to construct your command and args multiple times with different if statements)

what would you think of something along these lines instead?
Screenshot 2023-04-05 at 09 37 51

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
...
+			getYarnV.on('close', code => {
+				if (code === 0) {
+					resolve(null);
+				} else {
+					reject();
+				}
+			});
+		});
+		if (!yarnV.startsWith('1.')) {
+			await validateFile('.yarnrc.yml');
```
dario-piotrowicz, line 35: this line is not really doing anything, this function simply returns a boolean indicating if the file exists or not

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
...
+			const yarnYAML = YAML.load(await readFile('.yarnrc.yml', 'utf-8')) as {
+				nodeLinker: 'node-modules' | string;
+			};
```
dario-piotrowicz, line 38: nit, could we separate the lines?
```suggestion
                         const yarnrc = await readFile('.yarnrc.yml', 'utf-8');
			const yarnYAML = YAML.load(yarnrc) as {
				nodeLinker: 'node-modules' | string;
			};
```

also is there a benefit in `'node-modules' | string`?

```suggestion
                         const yarnrc = await readFile('.yarnrc.yml', 'utf-8');
			const yarnYAML = YAML.load(yarnrc) as {
				nodeLinker: string;
			};
```

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
...
+			if (yarnYAML.nodeLinker !== 'node-modules')
+				throw new Error(`
+				Next-On-Pages doesn't support Plug'n'Play features from yarn berry.
+
+				If you want to use Next-On-Pages with yarn berry,
+				please add "nodeLinker: node-modules" to your .yarnrc.yml
+				`);
+			return 'yarn (berry)';
+		} else return 'yarn (classic)';
```
dario-piotrowicz, line 47:
```suggestion
		} else {
		    return 'yarn (classic)';
		}
```

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
...
+			});
+		});
+		if (!yarnV.startsWith('1.')) {
+			await validateFile('.yarnrc.yml');
+			const yarnYAML = YAML.load(await readFile('.yarnrc.yml', 'utf-8')) as {
+				nodeLinker: 'node-modules' | string;
+			};
+			if (yarnYAML.nodeLinker !== 'node-modules')
+				throw new Error(`
+				Next-On-Pages doesn't support Plug'n'Play features from yarn berry.
```
dario-piotrowicz, line 41: We're trying to call this package consistently across the codebase as `@cloudflare/next-on-pages` 🙏 
```suggestion
				@cloudflare/next-on-pages doesn't support Plug'n'Play features from yarn berry.
```

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
...
+
+				If you want to use Next-On-Pages with yarn berry,
```
dario-piotrowicz, line 43:
```suggestion
				If you want to use @cloudflare/next-on-pages with yarn berry,
```

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function checkPackageManager() {
+	const userAgent = process.env.npm_config_user_agent;
+
+	const hasYarnLock = await validateFile('yarn.lock');
+	const hasPnpmLock = await validateFile('pnpm-lock.yaml');
+
+	if ((userAgent && userAgent.startsWith('pnpm')) || hasPnpmLock) return 'pnpm';
+
+	if ((userAgent && userAgent.startsWith('yarn')) || hasYarnLock) {
+		const yarn = process.platform === 'win32' ? 'yarn.cmd' : 'yarn';
```
dario-piotrowicz, line 16: I think that we can have a utility function in the utils folder, based on my previous comment I would suggest something like this:
Screenshot 2023-04-05 at 09 53 16

so you don't need to do the window checking every time, and this line would become
```diff
-		const yarn = process.platform === 'win32' ? 'yarn.cmd' : 'yarn';
+               const yarn = getSpawnCommand('yarn');
```

src/buildApplication/checkPackageManager.ts
```diff
@@ -0,0 +1,50 @@
+import YAML from 'js-yaml';
+import { spawn } from 'child_process';
+import { readFile } from 'fs/promises';
+import { validateFile } from '../utils';
+import { cliError } from '../cli';
+
+export async function checkPackageManager() {
```
dario-piotrowicz, line 7: could you please add unit tests for this function? 🙏

dario-piotrowicz:
Thanks a lot @LOLBRUHNICE 🙂, sorry I left a bunch of comments please have a look 🙏 

___

I would also like to add that Cloudflare Pages don't currently have built-in support for pnpm (see: https://github.com/cloudflare/next-on-pages/issues/52#issuecomment-1480345217) so your changes would work locally but not for deployment. We can land the PR anyways even though it's not fixing going to help there and we can make a followup PR to address the deployment issue if we want to wait for pages to support pnpm (which I believe should happen at some point soon-ish).

.changeset/nice-eyes-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Added support to utilize the package manager used in the project
```
dario-piotrowicz, line 5:
```suggestion
Added support to utilize the package manager used in the project for local development
```
(to reflect what I mentioned here: https://github.com/cloudflare/next-on-pages/pull/145#issuecomment-1497161287)

src/buildApplication/checkPackageManager.ts
LOLBRUHNICE: Hi, I have a problem that I could not unmock the spawn function from the child_process modules, do you know how to solve it?

LOLBRUHNICE: I'm trying to mock it to return an EventEmitter(fake child process) that return 3.0.0 to test it detecting yarn berry, then mock it again to return another EventEmitter that return 1.0.0 to test it detecting yarn classic.
The problem i have is the spawn function still returning 3.0.0 in the yarn classic test

LOLBRUHNICE: nvm, I found a way to fix it

types.d.ts
```diff
@@ -0,0 +1 @@
+declare type packageManager = 'pnpm' | 'yarn (berry)' | 'yarn (classic)' | 'npm';
```
dario-piotrowicz, line 1: sorry for being nitpicky but could you have this in the `src/utils/getSpawnCommand.ts` file and define it as:
```ts
type PackageManager = keyof typeof packageManagers;
```
?

I think that that would be better because you don't need to repeat anything more than once so it's more robust and maintainable (for example, if in the future we wanted to rename `"yarn (classic)"` to `"yarn 1"`, with this implementation we'd have to change it both here and in the packageManagers object, it's not a big deal since we'd get type errors if we'd forget one but still you get what I mean)

You're also not importing that type anywhere so it's not really going to add extra exports/imports

(also types should be capitalized 😜)

(and finally I am not sure, I think it is quite uncommon to create `.d.ts` for local types? we do that for the vercel types mainly I think because they are shared between the node code and the edge template code, but I don't think we should have other `.d.ts` files if we can help it)

dario-piotrowicz (approved):
@LOLBRUHNICE just one last round of comments, please have a look 🙇 

other than that it's all good to me, I'll just need to test the prerelease and we can merge it 😃

dario-piotrowicz:
Thank you very much @LOLBRUHNICE 😄

PR #153, opened 2023-04-06
feat: new routing system build time changes

james-elicx:
This PR is the first in a series of changes for the new routing system 😁. It addresses the build time aspect and does the following.

- New utility functions.
  - Formatting route paths.
    - Stripping `/index` so that file system matching in the new routing system works for requests to `/`.
    - Stripping route groups for matching in the file system in the new routing system as the build output config doesn't rewrite to route groups.
    - Stripping `.func` from path names.
    - Adding leading slashes to path names.
  - Reading path names recursively from the file system.
- Function map generation.
  - Warn for invalid middleware functions that are created outside of a base path as these shouldn't matter.
  - Account for middleware targeting an uncompiled entry when there is a valid compiled version in the same directory.
  - Check invalid functions against the functions map to see if valid ones were added from stripping route groups or `/index`.
  - Tests for new behaviour.
- Middleware manifest.
  - Support for additional function map entry for stripped index routes.
  - Strip the base path to match the middleware manifest.
  - Tests for new behaviour.
- Vercel config.
  - Process the Vercel config into a form that maps handler phase -> source routes.
  - Retrieve static assets from the file system.
  - Collect middleware paths from the build output config.
  - Rewrite the middleware paths into the same structure to match the Vercel config + avoid accidental path name matches in the new routing system.
  - Tests for new behaviour.
- Worker file.
  - Separate the `hasField` and `adjustRequestForVercel` into their own files.
  - Use new processed config and function map for requests.
- Types for processing the Vercel config and build output.

### DIFF
--- a/src/buildApplication/buildApplication.ts
+++ b/src/buildApplication/buildApplication.ts
@@ -15,6 +15,10 @@ import {
 } from './buildVercelOutput';
 import { buildMetadataFiles } from './buildMetadataFiles';
 import { validateDir } from '../utils';
+import {
+	getVercelStaticAssets,
+	processVercelOutput,
+} from './processVercelOutput';
 
 /**
  * Builds the _worker.js with static assets implementing the Next.js application
@@ -62,9 +66,7 @@ async function prepareAndBuildWorker(
 		cliLog(`Using basePath ${nextJsConfigs.basePath}`);
 	}
 
-	const functionsDir = resolve(
-		`.vercel/output/functions${nextJsConfigs.basePath ?? ''}`
-	);
+	const functionsDir = resolve(`.vercel/output/functions`);
 	if (!(await validateDir(functionsDir))) {
 		cliLog('No functions detected.');
 		return;
@@ -80,10 +82,14 @@ async function prepareAndBuildWorker(
 		return;
 	}
 
+	// NOTE: Middleware manifest logic will be removed in the new routing system.
 	let middlewareManifestData: MiddlewareManifestData;
 
 	try {
-		middlewareManifestData = await getParsedMiddlewareManifest(functionsMap);
+		middlewareManifestData = await getParsedMiddlewareManifest(
+			functionsMap,
+			nextJsConfigs
+		);
 	} catch (e: unknown) {
 		if (e instanceof Error) {
 			cliError(e.message, true);
@@ -96,9 +102,16 @@ async function prepareAndBuildWorker(
 		exit(1);
 	}
 
-	await buildWorkerFile(
-		middlewareManifestData,
+	const staticAssets = await getVercelStaticAssets();
+
+	const processedVercelOutput = processVercelOutput(
 		vercelConfig,
+		staticAssets,
+		middlewareManifestData
+	);
+
+	await buildWorkerFile(
+		processedVercelOutput,
 		nextJsConfigs,
 		options.experimentalMinify
 	);
--- a/src/buildApplication/buildWorkerFile.ts
+++ b/src/buildApplication/buildWorkerFile.ts
@@ -4,12 +4,28 @@ import { build } from 'esbuild';
 import { tmpdir } from 'os';
 import { cliLog } from '../cli';
 import { NextJsConfigs } from './nextJsConfigs';
-import { MiddlewareManifestData } from './middlewareManifest';
 import { generateGlobalJs } from './generateGlobalJs';
+import { ProcessedVercelOutput } from './processVercelOutput';
 
+/**
+ * Construct a record for the build output map.
+ *
+ * @param item The build output item to construct a record for.
+ * @returns Record for the build output map.
+ */
+function constructBuildOutputRecord(item: BuildOutputItem) {
+	return item.type === 'static'
+		? `{ type: ${JSON.stringify(item.type)} }`
+		: `{
+				type: ${JSON.stringify(item.type)},
+				entrypoint: AsyncLocalStoragePromise.then(() => import('${item.entrypoint}')),
+				matchers: ${JSON.stringify(item.matchers)}
+			}`;
+}
+
+// NOTE: `nextJsConfigs`, and accompanying logic will be removed in the new routing system.
 export async function buildWorkerFile(
-	{ hydratedMiddleware, hydratedFunctions }: MiddlewareManifestData,
-	vercelConfig: VercelConfig,
+	{ vercelConfig, functionsMap }: ProcessedVercelOutput,
 	nextJsConfigs: NextJsConfigs,
 	experimentalMinify: boolean
 ) {
@@ -25,22 +41,8 @@ export async function buildWorkerFile(
 			globalThis.AsyncLocalStorage = AsyncLocalStorage;
 		}).catch(() => undefined);
 
-		export const __FUNCTIONS__ = {${[...hydratedFunctions.entries()]
-			.map(
-				([name, { matchers, filepath }]) =>
-					`"${name}": { matchers: ${JSON.stringify(
-						matchers
-					)}, entrypoint: AsyncLocalStoragePromise.then(() => import('${filepath}'))}`
-			)
-			.join(',')}};
-
-		export const __MIDDLEWARE__ = {${[...hydratedMiddleware.entries()]
-			.map(
-				([name, { matchers, filepath }]) =>
-					`"${name}": { matchers: ${JSON.stringify(
-						matchers
-					)}, entrypoint: AsyncLocalStoragePromise.then(() => import('${filepath}'))}`
-			)
+		export const __BUILD_OUTPUT__ = {${[...functionsMap.entries()]
+			.map(([name, item]) => `"${name}": ${constructBuildOutputRecord(item)}`)
 			.join(',')}};`
 	);
 
--- a/src/buildApplication/generateFunctionsMap.ts
+++ b/src/buildApplication/generateFunctionsMap.ts
@@ -4,12 +4,14 @@ import { dirname, join, relative } from 'path';
 import { parse, Node } from 'acorn';
 import { generate } from 'astring';
 import {
+	formatRoutePath,
 	normalizePath,
 	readJsonFile,
+	stripIndexRoute,
 	validateDir,
 	validateFile,
 } from '../utils';
-import { cliError, CliOptions } from '../cli';
+import { cliError, CliOptions, cliWarn } from '../cli';
 import { tmpdir } from 'os';
 
 /**
@@ -27,10 +29,9 @@ import { tmpdir } from 'os';
 export async function generateFunctionsMap(
 	functionsDir: string,
 	experimentalMinify: CliOptions['experimentalMinify']
-): Promise<{
-	functionsMap: Map<string, string>;
-	invalidFunctions: Set<string>;
-}> {
+): Promise<
+	Pick<DirectoryProcessingResults, 'functionsMap' | 'invalidFunctions'>
+> {
 	const processingSetup = {
 		functionsDir,
 		tmpFunctionsDir: join(tmpdir(), Math.random().toString(36).slice(2)),
@@ -43,6 +44,8 @@ export async function generateFunctionsMap(
 		functionsDir
 	);
 
+	await checkInvalidFunctions(processingResults);
+
 	if (experimentalMinify) {
 		await buildWebpackChunkFiles(
 			processingResults.webpackChunks,
@@ -51,6 +54,39 @@ export async function generateFunctionsMap(
 	}
 
 	return processingResults;
+}
+
+/**
+ * Process the invalid functions and check whether and valid function was created in the functions
+ * map to override it.
+ *
+ * The build output sometimes generates invalid functions at the root, while still creating the
+ * valid functions. With the base path and route groups, it might create the valid edge function
+ * inside a folder for the route group, but create an invalid one that maps to the same path
+ * at the root.
+ *
+ * When we process the directory, we might add the valid function to the map before we process the
+ * invalid one, so we need to check if the invalid one was added to the map and remove it from the
+ * set if it was.
+ *
+ * @param processingResults Object containing the results of processing a directory.
+ */
+async function checkInvalidFunctions({
+	functionsMap,
+	invalidFunctions,
+}: DirectoryProcessingResults): Promise<void> {
+	if (invalidFunctions.size > 0) {
+		for (const rawPath of invalidFunctions) {
+			const formattedPath = formatRoutePath(rawPath);
+
+			if (
+				functionsMap.has(formattedPath) ||
+				functionsMap.has(stripIndexRoute(formattedPath))
+			) {
+				invalidFunctions.delete(rawPath);
+			}
+		}
+	}
 }
 
 async function processDirectoryRecursively(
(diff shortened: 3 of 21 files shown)


src/buildApplication/buildApplication.ts
```diff
@@ -80,10 +82,14 @@ async function prepareAndBuildWorker(
 		return;
 	}
 
+	// NOTE: Middleware manifest logic will be removed in the new routing system.
```
dario-piotrowicz, line 85: sorry for being nitpicky but wherever you have these comments could you please also mention your issue?
```suggestion
	// NOTE: Middleware manifest logic will be removed in the new routing system (see issue #129).
```
sorry I am sure you're going to clean everything up but it's just to be extra extra sure everything gets cleaned up (and in the eventuality something is left behind later we can see what "new routing system" is referring to and know that it can be cleaned up)

src/buildApplication/processVercelOutput.ts
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+
+	return paths;
+}
+
+/**
+ * Rewrite middleware paths in the functions map to match the build output config.
+ *
+ * Request path names will no longer accidently match middleware functions as the leading slash is
```
dario-piotrowicz, line 104: I must admit that I don't get the full picture of why this function is needed 🤔 😓 (could you maybe add some more context to the comment?)

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -7,27 +7,67 @@ beforeAll(() => {
...
+					path.includes('invalidTest') &&
+					path.includes('middlewarejs') &&
+					path.endsWith('.js');
 				const isFile = path.endsWith('.js') || path.endsWith('.json');
+
 				return {
-					isDirectory: () => !isFile,
-					isFile: () => isFile,
+					isDirectory: () => !invalidFile && !isFile,
+					isFile: () => !invalidFile && isFile,
```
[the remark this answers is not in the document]
james-elicx: So this is actually correct for the testing. `invalidFile` is true when it's a file ending with `.js` for a function called `middlewarejs` inside `invalidTest`. The `middlewarejs` function has an entry point to `middleware.js` and I needed a quick way to simulate changing the entry point from `middleware.js` to `index.js` and having that new entry point not exist in the file system, so this was a cheating way to do that.

This file system test is not pretty, elegant, or nice. This was more of a quick hack to make it have a specific behaviour without changing how you were already doing it. The way that we mock the file system needs replacing as previously discussed.

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -50,11 +90,33 @@ describe('generateFunctionsMap', async () => {
 			'validTest/functions',
 			false
 		);
-		expect(invalidFunctions.size).toEqual(0);
-		expect(functionsMap.size).toEqual(3);
-		expect(functionsMap.get('index')).toMatch(/\/index\.func\.js$/);
-		expect(functionsMap.get('index.rsc')).toMatch(/\/index\.rsc\.func\.js$/);
-		expect(functionsMap.get('api/hello')).toMatch(/\/api\/hello\.func\.js$/);
+
+		expect(invalidFunctions.size).toEqual(1);
```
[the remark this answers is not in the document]
james-elicx: I know. Sorry about that. I was trying not to completely change your function mock implementation while extending the tests to support the new behaviour where it squashes route groups. So, I am testing that invalid functions at the base do not work, while a valid squashed function replaces an invalid function at the base.

I know, it's not nice in the current implementation. Changing how we mock the file system will vastly help and improve these tests. I just wanted to try and test the additional edge cases without changing much about how you were mocking it.

src/buildApplication/buildWorkerFile.ts
```diff
@@ -4,12 +4,28 @@ import { build } from 'esbuild';
...
 import { generateGlobalJs } from './generateGlobalJs';
+import { ProcessedVercelOutput } from './processVercelOutput';
 
+/**
+ * Construct a record for the build output map.
+ *
+ * @param item The build output item to construct a record for.
+ * @returns Record for the build output map.
+ */
+function constructBuildOutputRecord(item: BuildOutputItem) {
```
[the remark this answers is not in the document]
james-elicx: If you're okay with me putting type assertions for everything then sure, although the type safety is kind of not really there if we're having to put `x as ...` for each property value. The problem is all the property values need to be strings to be able to be consumed as valid JS when we inject it.

tests/src/buildApplication/generateFunctionsMap.test.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah I see, ok thanks, could you please change the test description accordingly and/or add a comment here then? (just so that we can later split this when we get better fs mocking 🙂)

src/buildApplication/buildWorkerFile.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah no sorry you're right... you are producing strings 🤔 (sorry thought you were returning the objects)

You could just return the objects and have the types be correct, then do a stringify where you use the results or something like that 🤔 

Well not a big deal, I'm totally fine with this 🙂

tests/src/buildApplication/generateFunctionsMap.test.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah I see 😕 

thanks for explaining, not the best but I'm totally fine with it 🙂 

(if you could add a small comment here mentioning this that would be great! 🙏 )

src/buildApplication/buildWorkerFile.ts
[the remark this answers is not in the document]
james-elicx: Well both the `type` value and `entrypoint` value can't strictly match up with the actual type that it would be, so returning the object with the correct types would still need to heavily use assertions, and then you could return the object without stringifying it and stringify it later, but at that point, I feel like it's more beneficial to just remove the function and do it all in the same place like before.

I just wanted to separate the creating the stringified object from the actual consumption so it was a bit more readable - sadly I think this is the only thing that won't get real type safety 😅

src/buildApplication/processVercelOutput.ts
```diff
@@ -0,0 +1,126 @@
...
+		cliLog('No static assets detected.');
+		return [];
+	}
+
+	return (await readPathsRecursively(dir)).map(file =>
+		normalizePath(file.replace(dir, ''))
+	);
+}
+
+export type ProcessedVercelOutput = {
```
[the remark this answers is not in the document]
james-elicx: This was specific to the function so I defined it in the same place, while the ones I put in the `.d.ts` were with the intention of those being shared between the tool and the worker script.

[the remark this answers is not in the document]
dario-piotrowicz: yeah that's what I imagined, thanks for clarifying :slightly_smiling_face:

src/buildApplication/processVercelOutput.ts
james-elicx: Adding the following context to the comment (pasting here so you can read it without crawling through the code :P).

In the build output config, the `middlewarePath` value is used to denote where the entry point for a middleware function is located. This path does not have a leading slash.

For matching requests in the routing system, we use the path name from the request and check against the map of functions. The path name from the request will have a leading slash.

It might be possible to accidentally call a middleware function if the request path names matches the middleware path name in the functions map, so to avoid accidental calls, and to match the value in the build output config, we remove the leading slash from the key in the map for each middleware path.

dario-piotrowicz:
@james-elicx looks like you forgot to update some code after renaming the function 😢 
Screenshot 2023-04-06 at 22 12 53

tests/src/buildApplication/processVercelOutput.test.ts
```diff
@@ -0,0 +1,71 @@
...
+					filesystem: [{ src: '/test-3', dest: '/test-4' }],
+					miss: [{ src: '/test-2', dest: '/test-6' }],
+					rewrite: [],
+					resource: [],
+					hit: [],
+					error: [],
+				},
+			},
+			functionsMap: new Map([
+				['/static/test.png', { type: 'static' }],
```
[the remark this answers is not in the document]
james-elicx: Renaming to `vercelOutput` to be similar to `vercelConfig` which is also returned.

src/buildApplication/generateFunctionsMap.ts
```diff
@@ -43,6 +44,8 @@ export async function generateFunctionsMap(
 		functionsDir
 	);
 
+	await checkInvalidFunctions(processingResults);
```
dario-piotrowicz, line 47: @james-elicx 
Screenshot 2023-04-06 at 22 12 53

james-elicx: Yeah, sorry about that, I was using the commit changes thing on the GitHub website for some of your suggestions and I had to change that one manually, just pushed a commit which includes that fix as well addressing some other suggestions

src/buildApplication/processVercelOutput.ts
dario-piotrowicz: Awesome! thank you very much!!! 🙏

src/buildApplication/generateFunctionsMap.ts
dario-piotrowicz: ah no problem it happens :sweat_smile: , sorry about making suggestions that can break your code :sweat_smile: :bow:

dario-piotrowicz (approved):
Thanks a lot for addressing all my comments @james-elicx 🙏 

It's all good for me, thanks a lot for the awesome work! 🤩 

I've manually tested the changes on a couple of apps (since we unfortunately don't have e2e tests in place), I'll try to build another couple of apps tomorrow just to get even more confidence 🙂 

___

Since it this is a significant change I would really like for @jculvey and/or @GregBrimble to also have a look at the PR 🙂, but if they don't manage to do that soon-ish I think it's ok to merge this and not block your process with the routing refactoring 🙂

james-elicx:
Note that I have made some additional changes:
- Mapping invalid RSC functions to their valid non-RSC alternative, as discussed with @dario-piotrowicz.
- Add `overrides` from the Vercel build output to the build output map so that they can be retrieved by the pathname in the runtime functionality, instead of having to loop through the overrides object entries to check them.
- Added tests for the `overrides` functionality.

Overrides in the Vercel build output config are used to say that a static asset can be served from a different path, and the asset is served with a specific content type. This will help with serving pages that are generated statically and don't have corresponding functions in the new routing system. For example, to say that `/404` actually serves the static file `/404.html` with content type `text/html; charset=utf-8`.

src/buildApplication/processVercelOutput.ts
```diff
@@ -0,0 +1,192 @@
...
+				vercelOutput.set(servedPath, newValue);
+			}
+
+			// If the served path is an index route, add a squashed version of the path to the map.
+			const strippedServedPath = stripIndexRoute(servedPath);
+			if (strippedServedPath !== servedPath) {
+				vercelOutput.set(strippedServedPath, newValue);
+			}
+		}
+	}
```
dario-piotrowicz, line 191: The `if` + `for of` is not super readable in my opinion, what about something like this?

```suggestion
        
        Object.entries(overrides ?? []).forEach(([
			rawAssetPath,
			{ path: rawServedPath, contentType },        
        ]) => {
			const assetPath = addLeadingSlash(rawAssetPath);
			const servedPath = addLeadingSlash(rawServedPath);

			const newValue: BuildOutputStaticOverride = {
				type: 'override',
				path: assetPath,
				contentType,
			};

			// Update the existing static record to contain the new `contentType` and `assetPath`.
			const existingStaticRecord = vercelOutput.get(assetPath);
			if (existingStaticRecord?.type === 'static') {
				vercelOutput.set(assetPath, newValue);
			}

			// Add the new served path to the map, overriding the existing record if it exists.
			if (servedPath) {
				vercelOutput.set(servedPath, newValue);
			}

			// If the served path is an index route, add a squashed version of the path to the map.
			const strippedServedPath = stripIndexRoute(servedPath);
			if (strippedServedPath !== servedPath) {
				vercelOutput.set(strippedServedPath, newValue);
			}
	});
```

src/buildApplication/processVercelOutput.ts
```diff
@@ -0,0 +1,192 @@
...
+function collectMiddlewarePaths(routes: VercelSource[]): Set<string> {
+	const paths = new Set<string>();
+
+	for (const route of routes) {
+		if (route.middlewarePath) {
+			paths.add(route.middlewarePath);
+		}
+	}
+
+	return paths;
```
dario-piotrowicz, line 102:
```suggestion
	return new Set(
	    routes
                .map(route => route.middlewarePath)
                .filter(Boolean)
        );
```

tests/src/buildApplication/processVercelOutput.test.ts
```diff
@@ -0,0 +1,198 @@
...
+						entrypoint: '/middleware/index.js',
+						matchers: [],
+						type: 'middleware',
+					},
+				],
+			]),
+		});
+	});
+
+	test('applies overrides from the config to the outputted functions', () => {
```
dario-piotrowicz, line 72: thanks a lot for adding these! :pray:

src/buildApplication/generateFunctionsMap.ts
```diff
@@ -53,6 +56,51 @@ export async function generateFunctionsMap(
...
+			} else if (formattedPath.endsWith('.rsc')) {
+				const value = functionsMap.get(formattedPath.replace(/\.rsc$/, ''));
+
+				if (value) {
+					functionsMap.set(formattedPath, value);
+					invalidFunctions.delete(rawPath);
+				}
+			}
+		}
+	}
```
dario-piotrowicz, line 101: if instead of checking invalidFunctions.size > 0, you could check if it's equal 0 and return, so that we don't need to nest everything here, it might help in readability

not a big deal, what do you think? do you have a preference?

```suggestion
	if (invalidFunctions.size === 0) {
		return;
	}

	for (const rawPath of invalidFunctions) {
		const formattedPath = formatRoutePath(rawPath);

		if (
			functionsMap.has(formattedPath) ||
			functionsMap.has(stripIndexRoute(formattedPath))
		) {
			invalidFunctions.delete(rawPath);
		} else if (formattedPath.endsWith('.rsc')) {
			const value = functionsMap.get(formattedPath.replace(/\.rsc$/, ''));

			if (value) {
				functionsMap.set(formattedPath, value);
				invalidFunctions.delete(rawPath);
			}
		}
	}
```

src/buildApplication/middlewareManifest.ts
```diff
@@ -74,9 +79,12 @@ export function parseMiddlewareManifest(
 	const functionsEntries = Object.values(middlewareManifest.functions);
 
 	for (const [name, filepath] of functionsMap) {
+		// the .vc-config name includes the basePath, so we need to strip it for matching in the middleware manifest.
+		const fileName = (basePath ? name.replace(basePath, '') : name).slice(1);
```
dario-piotrowicz, line 83: maybe we could make this clearer with a regex? what do you think?
```suggestion
		const prefixRegex = new RegExp(`^(${basePath})?/`);
		const fileName = name.replace(basePathRegex, '');
```

src/utils/routing.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * The build output config does not rewrite `/` to `/index`, so we need to strip `/index` from the
+ * path name for request matching.
+ *
+ * @param path Path name to strip `/index` from.
+ * @returns Path name with `/index` stripped.
+ */
+export function stripIndexRoute(path: string) {
+	// add leading slash back if it is stripped when `/index` is removed
+	return addLeadingSlash(path.replace(/\/index$/, ''));
```
dario-piotrowicz, line 27:
```suggestion
	return addLeadingSlash(path.replace(/^\/index$/, ''));
```

dario-piotrowicz: spoke with @james-elicx, we might want to check that the `/index` is just at the end as well (or anyways it doesn't do any harm not checking that the string is precisely the `/index`)

dario-piotrowicz:
Thank you so so very much @james-elicx super awesome work! ❤️

Issue #169, opened 2023-04-12
[🐛 Bug]: Nextjs 13.3 Build Fail on Fresh Install


hjaber:
### Cli version

@cloudflare/next-on-pages@0.8.0

### Next.js related information

``` Operating System:       Platform: darwin       Arch: arm64       Version: Darwin Kernel Version 22.4.0: Mon Mar  6 21:01:02 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8112     Binaries:       Node: 18.12.1       npm: 9.6.4       Yarn: N/A       pnpm: 8.1.1     Relevant packages:       next: 13.3.0       eslint-config-next: 13.3.0       react: 18.2.0       react-dom: 18.2.0```

### Description

Failed with pnpm and npm for different reasons

Added:
* ```NODE_VERSION``` ```16``` to env variables
* ```Compatibility Flags``` ```nodejs_compat```
*  Change app/api and app/page to `export const runtime = 'edge'`
## Build Output pnpm
```js
14:06:12.041 | Cloning repository...
-- | --
14:06:12.920 | From https://github.com/hjaber/wind
14:06:12.921 | * branch            50f2bf47ce16c3b53ac519301c9e12bbf4def7aa -> FETCH_HEAD
14:06:12.921 |  
14:06:12.974 | HEAD is now at 50f2bf4 changed api/page runtime to edge
14:06:12.975 |  
14:06:13.132 |  
14:06:13.166 | Success: Finished cloning repository files
14:06:13.853 | Installing dependencies
14:06:13.871 | Python version set to 2.7
14:06:17.809 | v16.20.0 is already installed.
14:06:19.660 | Now using node v16.20.0 (npm v8.19.4)
14:06:19.924 | Started restoring cached build plugins
14:06:19.942 | Finished restoring cached build plugins
14:06:20.455 | Attempting ruby version 2.7.1, read from environment
14:06:22.495 | Using /opt/buildhome/.rvm/gems/ruby-2.7.1
14:06:23.603 | Using PHP version 5.6
14:06:23.761 | 5.2.5 is already installed.
14:06:23.791 | Using Swift version 5.2.5
14:06:23.792 | Started restoring cached node modules
14:06:23.812 | Finished restoring cached node modules
14:06:24.465 | Installing NPM modules using NPM version 8.19.4
14:06:25.757 | npm WARN config tmp This setting is no longer used.  npm stores temporary files in a special
[...]

PR #173, opened 2023-04-13
Next.js 13.3.0 favicon fix

dario-piotrowicz:
(Hopefully) temporary solution for https://github.com/cloudflare/next-on-pages/issues/169

### DIFF
--- a/src/buildApplication/generateFunctionsMap.ts
+++ b/src/buildApplication/generateFunctionsMap.ts
@@ -396,7 +396,7 @@ async function tryToFixFaviconFunc(): Promise<void> {
 			);
 			await copyFile(faviconFilePath, vercelStaticFavicon);
 		}
-		// let's delete  the .vercel/output/static/static directory so that extra media
+		// let's delete the .vercel/output/static/static directory so that extra media
 		// files are not uploaded unnecessarily to Cloudflare Pages
 		const staticStaticDir = resolve('.vercel', 'output', 'static', 'static');
 		rm(staticStaticDir, { recursive: true, force: true });


src/buildApplication/generateFunctionsMap.ts
```diff
@@ -354,6 +362,27 @@ async function buildWebpackChunkFiles(
...
 
+/**
+ * Next.js generates a favicon nodejs function instead of providing the favicon as
+ * a standard static asset. Since it is not using the edge runtime we can't use the
+ * function itself, so this function tries to see if it can find the standard favicon
+ * in the ./static/media directory generated by Next.js and moves that one in the static
+ * output directory instead
+ *
+ * (Note: this is hopefully just a temporary solution, hopefully Next.js will allow to opt
+ *        out from the favicon function generation or allow it to use the edge runtime)
```
dario-piotrowicz, line 373: - [ ] TODO: before merging create a new issue in the Next repository and reference it here

jculvey (approved):

src/buildApplication/generateFunctionsMap.ts
dario-piotrowicz: I've removed the comment entirely as I actually don't think Next is going to do anything about this (not soon at the very least) so this fix might need to stay 😕 😓 

@jculvey does this shound ok-ish to you?

dario-piotrowicz:
Not only https://github.com/vercel/next.js/pull/48528 didn't fix it but it seems like Next introduced requires to `node:buffer` so I needed to add a hack to solve that 😓 

(on the bright side they are saving the favicon in `.vercel/output` so we don't need to get the file from outside the dir)

PR #175, opened 2023-04-13
chore: update supported config properties

james-elicx:
Update the supported doc to reflect:
- no support for `crons` in the build output config
- possible future support for `i18n` in `next.config.js`.

### DIFF
--- a/docs/supported.md
+++ b/docs/supported.md
@@ -41,7 +41,7 @@ Earlier and Later versions might be only partially supported, we don't fully kno
 
 ## next.config.js Properties
 
-> The following options have been gathered from the Next.js [next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction), alongside the [Advanced routing documentation](https://nextjs.org/docs/advanced-features/i18n-routing).
+> The following options have been gathered from the Next.js [next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction), alongside the [Internationalized routing documentation](https://nextjs.org/docs/advanced-features/i18n-routing).
 
 | Option                                     | Support |
 | ------------------------------------------ | ------- |
--- a/docs/supported.md
+++ b/docs/supported.md
@@ -31,6 +31,7 @@ Earlier and Later versions might be only partially supported, we don't fully kno
 | wildcard                | 🔄      |
 | overrides               | 🔄      |
 | cache                   | ❌      |
+| crons                   | ❌      |
 
     - ✅: Supported
     - 🔄: Not currently supported, but it's probably possible and we may add support in the future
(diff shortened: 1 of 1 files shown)


docs/supported.md
```diff
@@ -69,6 +70,7 @@ Earlier and Later versions might be only partially supported, we don't fully kno
 | URL imports                                | ✅      |
 | build indicator<sup>7</sup>                | ❌      |
 | Turbopack-specific options<sup>8</sup>     | ❌      |
+| internationalized (i18n) routing           | 🔄      |
```
dario-piotrowicz, line 73: yeah I was thinking we should do this too :slightly_smiling_face: 

Too bad that it is not covered in the [Next.js next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction), making the note at the top of the page slightly incorrect, could you update that one as well? :pray: 

for example like so:
```
> The following options have been gathered from the [Next.js next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction) alongside the [Advanced routing documentation](https://nextjs.org/docs/advanced-features/i18n-routing).
```
truthfully I think we'd need to check the whole advanced-features section then and add all the missing configs from there here (or in a separate "advanced configs" table), but that's definitely going to be time consuming so if you can't do that now I don't want it to necessarily block this PR for it and we can just go with a temporary fix like this instead.

dario-piotrowicz (approved):
Thanks! :smile:

docs/supported.md
james-elicx: Yeah, that'd be a whole big task for another day 😅

docs/supported.md
```diff
@@ -40,7 +41,7 @@ Earlier and Later versions might be only partially supported, we don't fully kno
 
 ## next.config.js Properties
 
-> The following options have been gathered from the [Next.js next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction).
+> The following options have been gathered from the Next.js [next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction), alongside the [Advanced routing documentation](https://nextjs.org/docs/advanced-features/i18n-routing).
```
dario-piotrowicz, line 44: sorry my bad :bow: :pray: 
```suggestion
> The following options have been gathered from the Next.js [next.config.js documentation](https://nextjs.org/docs/api-reference/next.config.js/introduction), alongside the [Internationalized routing documentation](https://nextjs.org/docs/advanced-features/i18n-routing).
```

james-elicx: technically they would both be okay :p

dario-piotrowicz: yeah... :thinking: :sweat_smile: 

(but I guess using their same wording is better :sweat_smile: :grin:)

PR #180, opened 2023-04-17
chore: construct all paths consistenly using join

### LINKED ISSUE #154: Consistently use the `join` node path utility to construct paths
We should be using [`join`](https://nodejs.org/api/path.html#pathjoinpaths) to construct all our paths consistently in a platform independent way.

In places we even mix `join` with manually constructed paths such as the following:
https://github.com/cloudflare/next-on-pages/blob/2d72906c52e5c3a4cebd92a6311be5ba98b4cb98/bin/index.js#L9

it should instead be
```js
 join(__dirname, '..', 'dist', 'index.js'), 
```

We should fix all such paths.

tarunrajput:
Construct all our paths consistently in a platform independent way

resolves #154


dario-piotrowicz:
This is great! thanks a bunch @tarunrajput 😃 

All looks great, could you please rebase and solve the merge conflicts? (I tried to do that for you but it seems like I don't have permissions to push to your branch)

tarunrajput:
Sure @dario-piotrowicz, I've just updated the PR

dario-piotrowicz (approved):
Just one small comment and this should be ready to do 😄

dario-piotrowicz:
Thanks a lot @tarunrajput 😄

PR #181, opened 2023-04-17
chore: improve linting and typescript strictness

james-elicx:
This PR does the following:
- Bumps `@cloudflare/workers-types` to 4.x.
- Bumps TypeScript to 5.x.
- Makes the tsconfig much stricter.
- Modified the eslint config ever so slightly (references tsconfig.json and adds consistent type import rule).
- Removes the `.js` extension from the eslint and prettier configs.
- Applies necessary changes to comply with the stricter tsconfig.


james-elicx:
I think I messed up that rebase...

dario-piotrowicz:
> I think I messed up that rebase...

😢

src/buildApplication/buildMetadataFiles.ts
```diff
@@ -36,7 +36,7 @@ async function buildRoutes() {
 			{ flag: 'ax' } // don't generate file if it's already manually maintained
 		);
 	} catch (e) {
-		if (e.code != 'EEXIST') {
+		if ((e as { code?: string }).code != 'EEXIST') {
```
dario-piotrowicz, line 39:
```suggestion
		if ((e as { code?: string }).code !== 'EEXIST') {
```
😨 

could you please enable [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq) as well?

templates/_worker.js/utils/matcher.ts
```diff
@@ -35,5 +35,8 @@ export const hasField = (
 
 			return url.searchParams.has(has.key);
 		}
+		default: {
+			return false;
+		}
```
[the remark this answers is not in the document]
james-elicx: Ah yeah, I added that when the type for the has value was still nullable, but then I added NonNullable to the type and didn't remove it, so yeah that can be removed.

templates/_worker.js/index.ts
```diff
@@ -54,7 +54,10 @@ export const routesMatcher = (
 		}
 
 		if ('src' in route) {
-			const regExp = new RegExp(route.src, caseSensitive ? undefined : 'i');
+			const regExp = new RegExp(
+				route.src as string,
```
dario-piotrowicz, line 58:
```suggestion
				route.src!,
```
Not too important since this stuff is going to go await with the routing change but `as string` makes it look like `route.src` could be something other than a string, I think the correct thing here would be to use a non null assertion instead

although it makes eslint complain... generally I'm not sure about `@typescript-eslint/no-non-null-assertion`, what do you think? should we disable it? (personally I don't see anything wrong with non-null assertions when used correctly)

templates/_worker.js/index.ts
```diff
@@ -91,11 +94,11 @@ export default {
 		for (const route of routes) {
 			if (
 				'middlewarePath' in route &&
-				route.middlewarePath in __BUILD_OUTPUT__
+				(route.middlewarePath as string) in __BUILD_OUTPUT__
 			) {
-				const item = __BUILD_OUTPUT__[route.middlewarePath];
+				const item = __BUILD_OUTPUT__[route.middlewarePath as string];
```
dario-piotrowicz, line 99: again, I would avoid unnecessary/confusing casting and would go with something like this instead:
Screenshot 2023-04-18 at 10 05 42

env.d.ts
```diff
@@ -0,0 +1,10 @@
+declare global {
+	namespace NodeJS {
+		interface ProcessEnv {
+			NODE_ENV?: string;
+			npm_config_user_agent?: string;
+		}
```
dario-piotrowicz, line 6: Since we're already touching `ProcessEnv` I would also include generic keys:
```suggestion
		interface ProcessEnv {
			NODE_ENV?: string;
			npm_config_user_agent?: string;
			[key: string]: string|Fetcher;
		}
```
So that we can also remove the casting to unknown from here:
https://github.com/cloudflare/next-on-pages/blob/3fa5b0e9d4609393ec056df6f1073993807cba6c/templates/_worker.js/index.ts#L83

dario-piotrowicz (approved):

dario-piotrowicz:
Thanks you very much @james-elicx for the nice improvements! 😄

PR #184, opened 2023-04-18
add info CLI flag

tarunrajput:


src/cli.ts
```diff
@@ -182,3 +189,44 @@ function prepareCliMessage(
...
+			react-dom: ${getPackageVersion('react-dom')}
+			eslint-config-next: ${getPackageVersion('eslint-config-next')}
+	`);
+}
+
+function getPackageVersion(packageName: string): string {
+	try {
+		// eslint-disable-next-line @typescript-eslint/no-var-requires
+		return require(`${packageName}/package.json`).version;
+	} catch {
```
tarunrajput, line 220: @dario-piotrowicz, I am wondering whether we should use the package.json from project root, which would ensure that the package version is always available, even if the package is not installed or if we should require package.json for each package. While requiring package.json for each package works fine locally, I am encountering difficulties in figuring out why it is not working with prerelease.

dario-piotrowicz: @tarunrajput I don't think we can use the `package.json` from the project root, there you often specify packages using `^` or `~` so we could know the generic version there but not what the specific version the project is actually using (if this is what you were suggesting).

You could try using the `package-lock.json` file but that might be complicated, and you'd also have to deal with other type of lock files :sweat: 

one alternative I think is trying to use `npm list --json` as that outputs a json which should give you all the info you need (but only for installed packages, which might be fine?):
Screenshot at 2023-04-23 15-07-42

___

By the way, your prerelease does look to be working fine for me:
Screenshot at 2023-04-23 14-43-08

The only issue I can see if that if I link a project locally with a local version of `next-on-pages` the process is unable to get the version of the various packages since the working directory for the `next-on-pages` package is different from that of the current project (which is not ideal, my solution above should hopefully handle that correctly, although it's not a very common/important use case so if we can't get that to work I don't think it would be a deal breaker)

dario-piotrowicz: (PS: with `npm list` we would also get rid of that ugly `require` :smiley:)

dario-piotrowicz (commented):
Thanks a lot @tarunrajput, awesome stuff! :star_struck: 

Let's just address the few comments/issues mentioned and this should be good to go :smiley:

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -25,10 +25,10 @@ body:
       required: true
 
   - type: input
```
james-elicx, line 27:
```suggestion
  - type: textarea
```
Could you change this to a `textarea` so that formatting doesn't get messed up when copy-pasting info?

You could also remove the Vercel version input below, now that it is included in the environment info.

src/cli.ts
```diff
@@ -182,3 +189,44 @@ function prepareCliMessage(
...
+			Arch: ${os.arch()}
+			Version: ${os.version()}
+			CPU: (${os.cpus().length}) ${os.arch()} ${os.cpus()[0]?.model}
+			Memory: ${Math.round(os.totalmem() / 1024 / 1024 / 1024)} GB
+			Shell: ${process.env['SHELL']?.toString() ?? 'Unknown'}
+		Binaries:
+			Node: ${process.versions.node}
+			Yarn: ${getBinaryVersion('yarn')}
+			npm: ${getBinaryVersion('npm')}
+			pnpm: ${getBinaryVersion('pnpm')}
```
james-elicx, line 206: On Windows, these need to use `{name}.cmd`. You could use the utility in `utils/getSpawnCommand.ts` to do this.

.github/ISSUE_TEMPLATE/bug-report.yml
dario-piotrowicz: ah... I thought it was already a textarea! :sweat: 

that's why all current issues have the info on one line :sweat:

src/cli.ts
```diff
@@ -182,3 +191,60 @@ function prepareCliMessage(
...
+Relevant Packages:
+  @cloudflare/next-on-pages: ${nextOnPagesVersion}
+  vercel: ${getPackageVersion('vercel')}
+  next: ${getPackageVersion('next')}
+  react: ${getPackageVersion('react')}
+  react-dom: ${getPackageVersion('react-dom')}
+  eslint-config-next: ${getPackageVersion('eslint-config-next')}
+`.trim()
+	);
+}
```
dario-piotrowicz, line 221: I would suggest the following, so that:
 - we get extra lines before and after the info, which does look nice I think?
 - no eslint-config-next info, I'm not totally sure we care about it (we could probably also remove react and react-dom 🤷)
 - we can indent the message in the code however we like

```suggestion
export async function printEnvInfo(): Promise<void> {
	const envInfoMessage = dedent(`
		System:
			Platform: ${os.platform()}
			Arch: ${os.arch()}
			Version: ${os.version()}
			CPU: (${os.cpus().length}) ${os.arch()} ${os.cpus()[0]?.model}
			Memory: ${Math.round(os.totalmem() / 1024 / 1024 / 1024)} GB
			Shell: ${process.env['SHELL']?.toString() ?? 'Unknown'}
		Binaries:
			Node: ${process.versions.node}
			Yarn: ${getBinaryVersion('yarn')}
			npm: ${getBinaryVersion('npm')}
			pnpm: ${getBinaryVersion('pnpm')}
		Package Manager Used: ${await getCurrentPackageManager()}
		Relevant Packages:
			@cloudflare/next-on-pages: ${nextOnPagesVersion}
			vercel: ${getPackageVersion('vercel')}
			next: ${getPackageVersion('next')}
			react: ${getPackageVersion('react')}
			react-dom: ${getPackageVersion('react-dom')}
	`);

	// eslint-disable-next-line no-console
	console.log(`\n${envInfoMessage}\n`);
}
```

@tarunrajput what do you think? 🙂

dario-piotrowicz (approved):
Awesome! thank you so very much @tarunrajput! 🤩 

I left just one last small suggestion, please have a look 🙏

src/cli.ts
tarunrajput: Sure, I agree with your suggestions. I have updated the PR with these changes.

dario-piotrowicz: Awesome, thanks! 😄

PR #186, opened 2023-04-19
fix require("node:buffer") issue

dario-piotrowicz:
I have noticed that in 13.3.1-canary.13 Next introduces `require("node:buffer")`s which break the build of next-on-pages apps, this change is a preemptive fix for that


.changeset/six-teachers-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+fix `require("node:buffer")` breaking applications
```
petebacondarwin, line 5: Would be good to explain a bit more here

petebacondarwin (approved):
:shipit:

GregBrimble:
We should look into exporting stuff like this from Wrangler so others things (like next-on-pages) can consume them.

e.g. list of `node:*` importables, list of compatibility flags, mapping of compatibility flags to compat date that they are on-by-default for etc. etc.

_Probably_ only makes sense when we do the JS API, but worth keeping in mind.

PR #190, opened 2023-04-21
Use mock-fs for mocking the fs module (instead of vi.mock)

### LINKED ISSUE #158: [⚡ Feature]: Improve unit testing in regard to filesystem access
### Description

We currently have a ad-hoc/brittle/hacky/not great way of unit testing functions using node's fs module, we need to improve it  so that we can write unit tests in nice/pure manners.

Potentially using something like https://github.com/tschaub/mock-fs

dario-piotrowicz:
resolves #158


tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -1,105 +1,75 @@
...
+				api: {
+					'hello.func': validFuncDir,
+				},
+				'middlewarejs.func': {
+					'.vc-config.json': JSON.stringify({
+						name: 'middleware',
+						runtime: 'edge',
+						entrypoint: 'middleware.js',
+					}),
+					'index.js': '', // is this correct? @james-elicx
```
james-elicx, line 39: This directory should also have a `middleware.js` file present which we ignore, but I think it's worth still listing it to make it a bit clearer that it exists.

I believe the case we're not currently testing now is where it targets an uncompiled `middleware.js`, and there is no compiled `index.js` present. So, adding another function that targets `middleware.js` with only a `middleware.js` file present, that should fail, and then I believe we'll have tested all cases. This can probably be added in the invalid functions tests I guess.

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -1,105 +1,75 @@
...
 
-		// NOTE: The invalid function here is used to test that invalid functions on the root-level are considered invalid, while a valid squashed function (in a route group) replaces an equivalent invalid function that exists on the root-level.
-		// i.e. `(is-valid)/should-be-valid.func` replaces the invalid `should-be-valid.func` on the root-level, and `should-be-valid-alt.func` is still invalid.
-		expect(invalidFunctions.size).toEqual(1);
-		expect(Array.from(invalidFunctions.values())).toEqual([
-			'should-be-valid-alt.func',
-		]);
+		const { functionsMap } = await generateFunctionsMap('functions', false);
+
+		// Do we still need to test this? @james-elicx
```
james-elicx, line 67: So the reason I had this was to verify that route squashing was working properly. There was a root file that was invalid that wouldn't be overridden by squashing routes, and another that was invalid that would be overridden by squashing routes. This test would verify that the one without a path inside a group would be invalid, while the one expected to be overridden is valid.

At the moment, I don't believe such a case is being tested anymore, it is instead just being assumed. Probably worth adding a separate test just to make sure this expected behaviour is actually happening.

Input:
- should-be-valid.func -> invalid
- should-be-valid-alt.func -> invalid
- (is-valid)/should-be-valid.func -> valid

Expected output:
- should-be-valid.func -> valid as it was overridden by squashing a route group.
- should-be-valid-alt.func -> invalid as there was no other route to override it.

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -1,148 +1,159 @@
...
+				}),
+				'middleware.js': '',
+			},
+		});
+
+		expect(Array.from(invalidFunctions.values())).toEqual([]);
+		expect(fn.mock.calls.length).toBe(1);
+		expect(fn.mock.calls[0]?.[0]).toMatch(
+			/invalid middleware function for middlewarejs.func/
+		);
```
james-elicx, line 123: We should spy on the method and mock its implementation for `console.warn`, instead of setting it to `vi.fn()`. This also means that we can then clean it up and restore the original function at the end of the test.

```suggestion
		const mockedWarn = vi.spyOn(console, 'warn').mockImplementation(() => null);

		const { invalidFunctions } = await generateFunctionsMapFrom({
			'middlewarejs.func': {
				'.vc-config.json': JSON.stringify({
					name: 'middleware',
					runtime: 'edge',
					entrypoint: 'middleware.js',
				}),
				'middleware.js': '',
			},
		});

		expect(Array.from(invalidFunctions.values())).toEqual([]);
		expect(mockedWarn).toHaveBeenCalledTimes(1);
		expect(mockedWarn).toHaveBeenLastCalledWith(
			expect.stringMatching(/invalid middleware function for middlewarejs.func/)
		);

		mockedWarn.mockRestore();
```

tests/src/utils/fs.test.ts
```diff
@@ -70,69 +31,118 @@ describe('normalizePath', () => {
...
+		mockFs.restore();
+	});
 	test('should read all paths recursively', async () => {
-		await expect(readPathsRecursively('validTest/functions')).resolves.toEqual([
-			'validTest/functions/(route-group)/page.func/index.js',
-			'validTest/functions/index.func/index.js',
-			'validTest/functions/home.func/index.js',
-		]);
+		const paths = await readPathsRecursively('root/functions');
+		expect(paths.length).toBe(3);
```
james-elicx, line 135:
```suggestion
		expect(paths.length).toEqual(3);
```
consistency 🙂

tests/src/utils/fs.test.ts
```diff
@@ -70,69 +31,118 @@ describe('normalizePath', () => {
...
+	afterAll(() => {
+		mockFs.restore();
+	});
 	test('should read all paths recursively', async () => {
-		await expect(readPathsRecursively('validTest/functions')).resolves.toEqual([
-			'validTest/functions/(route-group)/page.func/index.js',
-			'validTest/functions/index.func/index.js',
-			'validTest/functions/home.func/index.js',
-		]);
+		const paths = await readPathsRecursively('root/functions');
```
james-elicx, line 134:
```suggestion
		const paths = (await readPathsRecursively('root/functions')).map(path =>
			normalizePath(path)
		);
```
the paths need to be normalized for the tests to pass for Windows users

tests/src/utils/fs.test.ts
dario-piotrowicz: from what I can tell we have a mix of both, I'm accepting the suggestion but if we want consistency we might need to change things anyways

tests/src/buildApplication/generateFunctionsMap.test.ts
dario-piotrowicz: doh, yeah you're totally right! silly mistake on my part! 😓👍

tests/src/utils/fs.test.ts
james-elicx: > from what I can tell we have a mix of both, I'm accepting the suggestion but if we want consistency we might need to change things anyways

Yeah, I was just looking in terms of the files changed in this PR - they all use `toEqual` and then there was this one straggler. The other couple of uses of `toBe` outside of this could do with changing at some point.

dario-piotrowicz: do you prefer `toEqual` in general?

because I think that if values are simply built-in types then `toBe` is more appropriate and `toEqual` is for more complex types such as objects and arrays

seems kind of more correct to me but I doubt that there is any real difference between the two (even performance wise)

james-elicx: > do you prefer `toEqual` in general?
> 
> because I think that if values are simply built-in types then `toBe` is more appropriate and `toEqual` is for more complex types such as objects and arrays
> 
> seems kind of more correct to me but I doubt that there is any real difference between the two (even performance wise)

I generally default to it, yeah - it's just personal preference at the end of the day. I don't really mind which we use, it's just that searching in VS Code, it looks like `toBe` is used a total of 4 times outside of the 2 you previously had in this PR. 🤷🏻‍♂️

jculvey (approved):

tests/src/utils/fs.test.ts
dario-piotrowicz: yeah as I said I _think_ `toBe` is more appropriate for simple types and `toEqual` for more complex ones, but I'm not really sure if that makes much of a difference (mostly this for me stems from the fact that if all can be done with `toEqual` why is `toBe` also provided in the first place? 😕)

PR #193, opened 2023-04-22
Ignore '.wrangler' directory in watch mode

Swagadon:
I'm using a D1 DB locally at `.wrangler\state\d1\DB.sqlite3`, and running `next-on-pages --watch`

When a function writes to the DB, the sqlite file is modified and triggers a rebuild
I can't think of any valid reason to have .wrangler under the watch paths - I've tested locally and this fixes it


dario-piotrowicz (approved):
Looks good to me, thanks a lot for the fix @Swagadon :smiley:

PR #201, opened 2023-04-25
Improve vercel build error msg

dario-piotrowicz:
Result:
Screenshot 2023-04-27 at 13 07 57


src/buildApplication/buildApplication.ts
```diff
@@ -32,10 +32,17 @@ export async function buildApplication({
 	if (!skipBuild) {
 		try {
 			await buildVercelOutput();
-		} catch (err) {
+		} catch {
 			cliError(
-				(err as Error)?.message ??
-					'Error: The Vercel build failed. For more details see the Vercel logs above.'
+				`
+					The Vercel build command failed. For more details see the Vercel logs above.
```
jculvey, line 38: Could we inline the command so that people can see what was run? If I was in this position, my next step might be to `npx vercel build` and see the output directly.

dario-piotrowicz: mh.... ok good point 🙂👍 , I'll add that in, although I hope that this doesn't get too verbose 😓

src/buildApplication/buildApplication.ts
```diff
@@ -32,10 +32,17 @@ export async function buildApplication({
...
+					
+					Note: The Vercel building process is run untouched by @cloudflare/next-on-pages,
```
jculvey, line 40: I like the intent here but this feels a little verbose. I think we should drop the first sentence and just say `If you need help solving the issue, refer to the Vercel or Next.js documentation`

dario-piotrowicz: I totally agree that it is verbose.... but I'm kind of trying to make it crystal clear here that the issue is not at all related to next-on-pages (so that we don't get extra issues etc...)

I'm afraid that removing the first sentence could make the message less clear? 😢

jculvey (approved):

dario-piotrowicz:
@jculvey I need to update the `getCurrentPackageManager` related files' names, I didn't do it because I think it might mess with the diff and show them as new files, please have a review and I'll update the filenames after you had a chance to have a look

jculvey (approved):

PR #204, opened 2023-04-25
Add pages debug options to bug report form

GregBrimble:


dario-piotrowicz (approved):
I'm a bit afraid that this might make the starting template a bit too big (making people less willing to follow it)

But I guess these could be valuable, I think we can go with it and anyways if we decide that it's too much we can remove them later 🙂

GregBrimble:
I'd pretty strongly like to keep the direct upload vs. CI one, but you can chop the deployment ID if it's too much hassle. Made it optional to hopefully not add too much friction though.

dario-piotrowicz:
@GregBrimble nah it's fine, I was just pointing out this small concern I've got, but it's likely not a real issue :slightly_smiling_face:

PR #205, opened 2023-04-28
Fix Wrangler Argument

SirCremefresh:
I had to change --compatibility-flag to --compatibility-flags for it to work.

It is also with a s in the documentation: https://developers.cloudflare.com/workers/platform/nodejs-compatibility/

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ npx @cloudflare/next-on-pages
 This command will build your Next.js application and produce a `.vercel/output/static` directory which you can then use with Wrangler:
 
 ```sh
-npx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat
+npx wrangler pages dev .vercel/output/static --compatibility-flags=nodejs_compat
 ```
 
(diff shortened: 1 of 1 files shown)


dario-piotrowicz:
Thanks a lot for taking the time to open the PR @SirCremefresh 😃 

However I am not sure I get it, I've been using `--compatibility-flag` for quite a while and it has always worked, if you run `wrangler pages dev -h` you can also see that both `--compatibility-flags` and `--compatibility-flag` should work:
Screenshot 2023-04-28 at 10 30 56

Maybe this is a bug in wrangler? could double check and eventually you provide some more details? (like your command, node version, wrangler version, etc...)

SirCremefresh:
You are right it now works as well, i did not change the version. I don't know why I got the error.
I can't reproduce it. I will close this issue.

my error:
Screenshot 2023-04-28 at 13 17 41

SirCremefresh:
Hi @dario-piotrowicz 

I don't think so when I copy paste the command that gave the error and run it again it works.
I also now checked the wrangler release log and didn't see anything related.
I can't think of anything to to further investigate the error. 

command as plain text: npx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat

PR #206, opened 2023-04-29
feat: new routing system runtime implementation and changes

### LINKED ISSUE #21: Middleware locally and on CF Pages have different response when returning cookies. Local returns the index page with cookie set. CF only returns cookies without the html.
Not certain if this is intended behaviour. When running using `npm run dev`, the middleware below would add the cookie to the current request and then display the home page content. 

[Same method as nextjs middleware docs](https://nextjs.org/docs/advanced-features/middleware#using-cookies)
```
    // Curr URL = /

export function middleware(request: NextRequest) {

    const response = NextResponse.next();
    response.cookies.set("myCookie", cookieVal);

    return response;
}

// See "Matching Paths" below to learn more
// Only set edge runtime for middleware, otherwise typewriter effect will break due to nodejs requirement
export const config = {
  runtime: 'experimental-edge',
  matcher: "/",
};

```

However, on CF Pages, it would not display the actual page content. Only rendering a pre in the body, without the index page html. The network tab does return the cookie though. 

# Edit Further Testing

Just tested with a blank return next() and it does the same thing. Nothing loads in the html. 
`return NextResponse.next()`

Is there something wrong with using the above code? Not even doing anything fancy either.

# Versions 
```
    "next": "^13.0.2",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
     ....

### LINKED ISSUE #70: Page redirection fully reloads the page
Hello, implementing the experimental edge on the CloudFlare page causes the pages to reload, I made texts using route.push and Link, will this be resolved?

### LINKED ISSUE #129: Proposed Changes to use the Vercel Build Output for Routing
Hello, this issue outlines the proposed changes that I would like to make to how `next-on-pages` handles routing.

The Vercel [build output config](https://vercel.com/docs/build-output-api/v3#build-output-configuration) contains the different steps taken during the routing process used by Vercel, in a logical format, and was designed around making it easy for frameworks to build and deploy on Vercel. This is advantageous in that we can attempt to mirror how their routing works by using this config. 

I previously intended to create a PR once I had a working concept, although I feel it would be more beneficial for me to create an issue outlining how I intend it to function. This is so that these thoughts can be discussed and prevent me from wasting time implementing a section that people may disagree on.

I imagine parts of the proposed system may seem confusing at first glance. I am happy to answer any questions you may have about how I have laid it out.

I would appreciate it if this issue could be assigned to me as I would like to build it out :)

### Build time

We extract the middleware manifest from the Next.js build and use this to find our function files. We then minify the entry points and inject the middleware manifest matches and the entry point into the worker script. Additionally, we inject the Vercel build output config.

### Runtime

Upon each request, we currently loop through the Vercel build output config, check requirements such as `method`/`has`/`missing`/`caseSensitive`, and then check if the `src` value matches. If it does, we add the route to a list of matching routes. If there was no `src` value at all, we also add that handle to the matching routes list. After all that, we check for middleware matches in that list of matching routes and run any middleware. We then loop through all the injected functions and check for a match there. If one is found, we return the value of calling the entry point.

## Proposed System

These changes remove the need to use the Next.js build output's middleware manifest and rely solely on the Vercel build output. The routing process during runtime is based on the routing [handler phases](https://vercel.com/docs/build-output-api/v3#build-output-configuration/supported-properties/routes/handler-route) from the Vercel build output config.

Through analysing the result of building with various rewrite options and consulting the [Next.js documentation](https://nextjs.org/docs/api-reference/next.config.js/rewrites), I further established that dynamic routes and non-dynamic routes are handled at two different points. Additionally, it became clear that dynamic params are handled in the rewrite phase during runtime through two separate regular expressions that add the params as search params. With respect to middleware, the [documentation](https://nextjs.org/docs/advanced-features/middleware#matching-paths) also detailed that they run between redirects and the first rewrite options.

All the phases listed under runtime are in the order that they occur in the build output's config, except for the `resource` phase. In the outputted config, the `resource` phase is written between `filesystem` and `miss`. I do not believe it should be run at this point based on what I have read, and the below list reflects the points at which each phase will logically run, mostly matching up with the Vercel config. The reason for this is that the `resource` phase handles the `fallback` rewrite option. Per the Next.js [docs](https://nextjs.org/docs/api-reference/next.config.js/rewrites), this rewrite option is applied after all routes+static assets are checked, thus leading me to put it where I believe it would logically take place.

Tests should of course be added when any implementations of the new system are written.

### Build time

Inject the build output config (`.vercel/output/config.json`) into the worker script.

Static assets (`.vercel/output/static`):
- Extract the path names.
- Inject list of path names into worker script.

Functions (`.vercel/output/functions`):
- Extract the path names.
- Extract the entry points.
- Minify the entry points.
- Inject list of path names + entry points into worker script.

### Runtime

no handler phase
  - applies headers.
  - applies redirects.
  - applies/processes middleware.
  - applies `beforeFiles` rewrites.
  - `continue` === false -> return response.

`filesystem` phase (occurs if path name not found in functions list (non-dynamic) or static paths)
  - applies `afterFiles` rewrites.
  - `check` === true -> check filesystem phase again after the `rewrite` phase.

`miss` phase (occurs if `filesystem` phase misses - no `src` match)
  - `check` === true -> check filesystem phase again after the `rewrite` phase.

`rewrite` phase (occurs always)
  - applies dynamic params.
  - check pathname is in functions (incl. dynamic) or static paths.
    - miss ->
      - if `check` === true earlier:
        - enter `filesystem` phase.
          - no `src` match -> enter `miss` phase.
      - else:
        - enter `resource` phase.

`resource` phase (occurs if `rewrite` phase ultimately resulted in filesystem miss)
  - applies `fallback` rewrites.
  - `check` === true
    - call `filesystem` phase.
      - no `src` match -> final result is miss.

`hit` phase (occurs if a function file path or static path matched)
  - call the function's entrypoint.
  - return response with applied headers + status code.

`error` phase (occurs if response code was 4xx or 5xx)
  - if status matches, rewrite to dest.
  - return response.

### LINKED ISSUE #163: [🐛 Bug]: navigating directly to a dynamic route
### Cli version

0.7.0

### Next.js related information

     Operating System:       Platform: darwin       Arch: arm64       Version: Darwin Kernel Version 21.6.0: Mon Dec 19 20:46:01 PST 2022; root:xnu-8020.240.18~2/RELEASE_ARM64_T8101     Binaries:       Node: 19.6.0       npm: 9.4.0       Yarn: 1.22.19       pnpm: N/A     Relevant packages:       next: 13.3.0       eslint-config-next: 13.3.0       react: 18.2.0       react-dom: 18.2.0

### Vercel version

28.18.3

### Description

I've seen this problem mentioned in a few other places but I don't see an issue specifically documenting it with steps to reproduce.

The file `src/pages/members/[userId]/index.tsx` should allow you to paste the url `localhost:3000/members/1234` into the browsers and have the code in the index.tsx file render the page.  This works fine with the default next commands `npm run dev` and `npm run build && npm run start`, but when you do this with `npx @cloudflare/next-on-pages` and `npx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat --port=3000` (or upload to the cloudflare pages environment) navigating directly to the page returns a 404 page.

Navigating to a member page with a Next `<Link>` component, or using Next's `router.push(url)` works fine in the cloudflare environment.

### Reproduction

https://github.com/dodrian/pages-next-app

From the root index page there are links (including dynamically generated links) to different member pages, and an input box with a button that will use router.push to navigate you to a member page.  Following these links works fine, but if you then reload the page it returns a 404.

### Additional Information

_No response_

### Would you like to help?

### LINKED ISSUE #177: [🐛 Bug]: `NextResponse.rewrite(new URL(xxx))` does not work correctly.
### Cli version

v0.8.0

### Next.js related information

     Operating System:       Platform: darwin       Arch: arm64       Version: Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:28 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6000     Binaries:       Node: 19.6.1       npm: 9.4.0       Yarn: 3.5.0       pnpm: N/A     Relevant packages:       next: 13.3.0       eslint-config-next: 13.3.0       react: 18.2.0       react-dom: 18.2.0  warn  - Latest canary version not detected, detected: "13.3.0", newest: "13.3.1-canary.6".         Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.         Read more - https://nextjs.org/docs/messages/opening-an-issue

### Vercel version

28.18.5

### Description

I want to change the default API prefix `/api` of the nextjs to another one, so I created `middleware.ts` in `src`, and fill with following content:

```ts
import { NextRequest, NextResponse } from "next/server";

export const config = {
  matcher: "/v1/:path*",
};

export function middleware(request: NextRequest) {
  if (request.nextUrl.pathname.startsWith("/v1")) {
    return NextResponse.rewrite(
      new URL(`/api${request.nextUrl.pathname}`, request.url)
    );
  }
}
```

When I test with `next dev` command, it's all correct. when I fetch `/v1/hello` will get the response that `/api/hello` responses. But when I test with `yarn dlx @cloudflare/next-on-pages` and `yarn dlx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat` command, the `/v1/hello` will responses with `200` status code but with empty body.

I don't know is it a bug from `next-on-pages`, or it's I used incorrectly. Or is there another way to implement my requiremet.

### Reproduction

_No response_

### Additional Information

_No response_

### Would you like to help?

### LINKED ISSUE #185: [🐛 Bug]: Next.js catchall and edge routes
### Cli version

Latest

### Next.js related information

Operating System:   Platform: linux   Arch: x64   Version: Binaries:   Node: 19.8.1   npm: 8.19.2   Yarn: 1.22.19   pnpm: N/A Relevant packages:   next: 13.3.0   eslint-config-next: 13.3.0   react: 18.2.0   react-dom: 18.2.0

### Vercel version

_No response_

### Description

When creating a Next.js catchall `[[...index]].tsx` it overwrites the edge API routes.

### Reproduction

https://ff2c783e.next-on-pages-catchall-api.pages.dev/catchall - Should be catchall page
https://ff2c783e.next-on-pages-catchall-api.pages.dev/api/hello - Should be `hello` in plaintext (edge function)

https://github.com/x052/next-on-pages-catchall-api

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Implements a new routing system designed around the Vercel build output.
- Replaces the old routing implementation in the _worker.js template.
- Removes the old build-time logic that is no longer necessary.
- Fixes a bug where we were unable to fix functions that had valid squashed versions.
- Allows building a worker file when no functions are found.

I know that we previously discussed doing 3 PRs, with the third PR being the one where the redundant build-time logic is removed and the routing system is enabled, however, Dario and I agreed that it would probably be best to do it in this PR instead.

Some of the functionality that this PR also introduces support for is as follows:
- next.config.js: rewrites, redirects, custom headers.
- middleware: redirects, rewrites, request header overrides, response manipulation, etc.
- build output config route properties: dest, headers, continue, check, status.
- correct dynamic/catch-all parameter handling.
- statically generated routes.

<details>
<summary>Overview of routing</summary>

- Call the function to start finding a match for the request. The matching function is recursive and calls itself again every time it enters a new phase.
  - Starts matching from the `none` phase. 
  - For each route in the current phase.
    - Checks the `src` value for a match, and extracts the dynamic parameters.
    - Checks `override` and reset the response status and headers if it exists
    - Checks for `middlewarePath`, calls the middleware function, and processes the response.
    - Checks for `headers` and adds them to the response.
    - Checks for `status` and sets the response status.
    - Checks for `dest` and updates the path for matching with the new destination, applying the dynamic parameters from earlier.
    - Checks for `check` if the path is not an external URL.
      - If the rewritten path is equal to the previous path, enter the next phase i the current is not `miss`.
      - otherwise, enter the `none` phase.
    - Checks for `continue` to know if routing should continue.
  - If the route is found in the build output, it returns the match.
  - Decides what happens next, based on the current phase.
    - `hit` phase or for URLs / redirects -> return the match.
    - `miss` phase -> update status and enter `hit` phase to update headers.
    - `error` phase -> enter `hit` phase to update headers.
  - If routing should continue, it decides what should happen next for the following phases.
    - `none` phase -> enter `filesystem` phase.
    - `filesystem` phase -> enter `rewrite` phase.
    - `rewrite` phase -> enter `resource` phase.
    - `resource` phase -> enter `miss` phase.
- If the final status is an error, it calls matching with the `error` phase to find the custom error page.
- Serves a match.
  - If the match is a URL / redirect, it redirects to the URL.
  - It runs the entrypoint for the route.
  - Headers and status codes from routing are applied to the response.
  - The response gets returned.
  </details>

I have tested this with some of the apps I have, but I warmly welcome others to test the prerelease against their projects.

fixes #185 
fixes #177
fixes #163
fixes #21
fixes #70
fixes #129


dario-piotrowicz (commented):
Needless to say, very awesome stuff thanks a lot :heart: 

First round of comments :pray: :sweat_smile:

templates/_worker.js/utils/pcre.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param captureGroupKeys Named capture group keys from the `src` matcher.
+ * @returns The processed string with replaced parameters.
+ */
+export function applyPCREMatches(
+	rawStr: string,
+	match: RegExpMatchArray,
+	captureGroupKeys: string[]
+): string {
+	return rawStr.replace(/\$([a-zA-Z0-9]+)/g, (_, key) => {
+		const index = captureGroupKeys.indexOf(key);
```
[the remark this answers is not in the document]
james-elicx: I don't believe so. The named keys are the dynamic route params, and if you use nested dynamic routes with the name, you get the following error from nextjs.
```
Error: You cannot have the same slug name "test" repeat within a single dynamic path
```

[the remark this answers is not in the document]
dario-piotrowicz: ok sounds good thanks :slightly_smiling_face::+1:

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+
+describe('checkRouteMatch', () => {
+	const testCases: CheckRouteMatchTestCase[] = [
+		{
+			name: 'matches a route with only `src`',
+			route: { src: '^/index$' },
+			expected: true,
+		},
+		{
+			name: 'doesnt match a route with invalid `src`',
```
dario-piotrowicz, line 107:
```suggestion
			name: "doesn't match a route with invalid `src`",
```

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+		{
+			name: 'matches a route with `src` and `missing`',
+			route: {
+				src: '^/index$',
+				missing: [{ type: 'host', value: 'example.com' }],
+			},
+			expected: true,
+		},
+		{
+			name: 'doesnt match a route with `src` and invalid `missing`',
```
dario-piotrowicz, line 137:
```suggestion
			name: "doesn't match a route with `src` and invalid `missing`",
```

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+				src: '^/index$',
+				missing: [
+					{ type: 'host', value: 'example.com' },
+					{ type: 'query', key: 'baz' },
+				],
+			},
+			expected: true,
+		},
+		{
+			name: 'doesnt match with `src` and multiple `missing` (one valid, one invalid)',
```
dario-piotrowicz, line 156:
```suggestion
			name: "doesn't match with `src` and multiple `missing` (one valid, one invalid)",
```

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+			expected: true,
+		},
+		{
+			name: 'match with `status` and required status',
+			route: { src: '^/index$', status: 404 },
+			requiredStatus: 404,
+			expected: true,
+		},
+		{
+			name: 'doesnt match with invalid `status` and required status',
```
dario-piotrowicz, line 182:
```suggestion
			name: "doesn't match with invalid `status` and required status",
```

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+		{
+			name: 'header: has with key+value mismatch returns false',
+			has: { type: 'header', key: 'headerWithValue', value: 'value2' },
+			expected: false,
+		},
+		{
+			name: 'header: has with key match returns true',
+			has: { type: 'header', key: 'headerWithoutValue' },
+			expected: true,
+		},
```
dario-piotrowicz, line 54:
```suggestion
		},
		{
			name: 'header: has with key but no value mismatch returns false',
			has: { type: 'header', key: 'headerWithoutValue', value: 'value' },
			expected: false,
		},
		}
```

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+		{
+			name: 'query: has with key+value mismatch returns false',
+			has: { type: 'query', key: 'foo', value: 'bar2' },
+			expected: false,
+		},
+		{
+			name: 'query: has with key match returns true',
+			has: { type: 'query', key: 'bar' },
+			expected: true,
+		},
```
dario-piotrowicz, line 84:
```suggestion
		},
		{
			name: 'query: has with key but no value mismatch returns false',
			has: { type: 'query', key: 'bar' , value: 'baz' },
			expected: false,
		},
```

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+	route: VercelSource;
+	requiredStatus?: number;
+	expected: boolean;
+};
+
+const req = new Request('https://test.com/index?foo=bar&bar=', {
+	headers: {
+		headerWithValue: 'value',
+		headerWithoutValue: undefined as unknown as string,
+		cookie: 'foo=bar; bar=',
```
dario-piotrowicz, line 22: I would suggest using the same convention you used for the headers:
```suggestion
		cookie: 'cookieWithValue=value; cookieWithoutValue=',
```
(I especially dislike the double use of `bar` here since in the tests when you will see `bar` you might get confused to which one the test is referring to)

tests/templates/utils/matcher.test.ts
```diff
@@ -0,0 +1,212 @@
...
+				url,
+				cookies,
+				headers: req.headers,
+			});
+			expect(result).toEqual(testCase.expected);
+		});
+	});
+});
+
+describe('checkRouteMatch', () => {
```
dario-piotrowicz, line 99: I think you don't have any tests for non-matching HTTP methods? could you add some?

dario-piotrowicz (commented):
second round of comments :sweat_smile: 

@james-elicx I put there some comments regarding `checkRouteMatch` and related, but I still need to double check things later and step through the function (just letting you know, the comments should still be valid regardless but I might put new ones if I notice that something isn't working as expected :slightly_smiling_face:)

templates/_worker.js/utils/matcher.ts
```diff
@@ -1,39 +1,105 @@
...
+	headers: Headers;
+};
+
+/**
+ * Check if a Vercel source route's `has` record conditions match a request.
+ *
+ * @param has The `has` record conditions to check against the request.
+ * @param param.url The URL object.
+ * @param param.cookies The cookies object.
+ * @param param.headers The headers object.
```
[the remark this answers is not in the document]
james-elicx: just a personal preference i often do when destructuring in the function arguments, normally i do `args.[key]`, although i accidently did `param.[key]` there.

Anyway, I've just removed all instanced of `args.[key]` and `param.[key]`.

templates/_worker.js/utils/routing.ts
```diff
@@ -0,0 +1,169 @@
...
+		}
+		// `rewrite` applied dynamic params to requests. It checked dynamic routes.
+		case 'rewrite': {
+			return 'resource';
+		}
+		// `resource` applied `fallback` rewrites. It checked the final routes.
+		case 'resource': {
+			return 'miss';
+		}
+		default: {
```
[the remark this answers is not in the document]
james-elicx: The default should never actually be hit - only the 4 phases listed should be hit, it was just cleaner to separate the switch into it's own function. I shall add a comment to reflect that and change the accepted type for `phase` to reflect this.

templates/_worker.js/utils/routing.ts
```diff
@@ -0,0 +1,169 @@
...
+	status: number | undefined;
+	headers: { normal: Headers; important: Headers };
+	searchParams: URLSearchParams;
+};
+
+/**
+ * Gets the next phase of the routing process.
+ *
+ * Determines which phase should follow the `none`, `filesystem`, `rewrite`, or `resource` phases.
+=======
```
dario-piotrowicz, line 19:
```suggestion
```

templates/_worker.js/utils/http.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @param source Headers to apply.
+ * @param target Headers object to apply to.
+ * @param pcreMatch PCRE match result to apply to header values.
+ */
+export function applyHeaders(
+	source: Record<string, string> | Headers,
+	target: Headers,
+	pcreMatch?: MatchPCREResult
+): void {
```
dario-piotrowicz, line 15: sorry for the nitpick, this threw me off when reviewing your PR :sweat:

Could you swap target and source?
I think it's standard that the target is first and then comes the source(/s)
like for Object.assign: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

templates/_worker.js/matcher.ts
```diff
@@ -0,0 +1,414 @@
...
+	 *
+	 * @param route Build output config source route.
+	 * @param checkStatus Whether to check the status code of the route.
+	 * @returns The source path match result if the route matches, otherwise `undefined`.
+	 */
+	private checkRouteMatch(
+		route: VercelSource,
+		checkStatus?: boolean
+	): MatchPCREResult | undefined {
+		return checkRouteMatchUtility(route, this.path, {
```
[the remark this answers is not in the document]
james-elicx: If I move the function inside the class, then I'll delete the tests that are written specifically for the function because, well, I'm not making it a static method, and it's basically tested as part of the router tests anyway. If you're fine with that, I can.

templates/_worker.js/matcher.ts
```diff
@@ -0,0 +1,414 @@
...
+		}
+
+		// NOTE: Special handling for `.rsc` requests. If the Vercel CLI failed to generate an RSC
+		// version of the page and the build output config has a record mapping the request to the
+		// RSC variant, we should strip the `.rsc` extension from the path.
+		if (/\.rsc$/i.test(this.path) && !(this.path in this.output)) {
+			this.path = this.path.replace(/\.rsc/i, '');
+		}
+
+		// If not an external URL, merge search params for later use.
```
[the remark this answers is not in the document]
james-elicx: If it's an external URL, we don't apply the search parameters before redirecting. We only apply search parameters to rewritten URLs so that they can be consumed by the page/function when it is run.

Perhaps this behaviour should be changed so that we always apply search parameters, regardless of whether it's a full URL or not. What do you think? There wasn't a specific reason I didn't do that, so maybe I should?

templates/_worker.js/router.ts
```diff
@@ -0,0 +1,120 @@
...
+	private async findMatch(
+		matcher: Matcher,
+		phase: 'none' | 'error' = 'none',
+		skipErrorMatch = false
+	): Promise<MatchedSet> {
+		const result = await matcher.run(phase);
+
+		if (
+			result === 'error' ||
+			(!skipErrorMatch && matcher.status && matcher.status >= 400)
```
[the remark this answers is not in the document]
james-elicx: TypeScript complains if you don't check that it's not undefined.

templates/_worker.js/matcher.ts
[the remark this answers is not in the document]
dario-piotrowicz: yeah sounds like that could be simpler (and if it turns out problematic we can always add this external check back in)

I would go with that if you're ok with it, just for the sake of keeping things (slightly) simpler and not having extra logic which might not be needed.

Anyways If you want I am also ok with the current implementation, I was asking `why` just because honestly I wasn't sure as to why the check is here, I wasn't saying that it is wrong or anything. But if you do decide to keep the check please expand the comment to explain also the reason as to why we do this :slightly_smiling_face: :pray:

templates/_worker.js/router.ts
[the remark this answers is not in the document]
dario-piotrowicz: :confused: 

Screenshot at 2023-05-01 18-42-31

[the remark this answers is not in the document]
dario-piotrowicz: ah yeah... on my machine it didn't for some reason but I can see that now :sweat::+1:

templates/_worker.js/matcher.ts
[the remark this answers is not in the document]
james-elicx: I ended up adapting the tests to use the new router test suite instead. 👍🏻

templates/_worker.js/matcher.ts
```diff
@@ -0,0 +1,414 @@
...
+	getNextPhase,
+	isUrl,
+	runOrFetchBuildOutputItem,
+	checkRouteMatch as checkRouteMatchUtility,
+} from './utils';
+
+export type CheckRouteStatus = 'skip' | 'next' | 'done' | 'error';
+export type CheckPhaseStatus = Extract<CheckRouteStatus, 'error' | 'done'>;
+
+export class Matcher {
```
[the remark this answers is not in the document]
james-elicx: Renamed the file in a separate commit so you can check the diff for all the other changes from the suggestions without it making it hard to tell what happened 🙂

james-elicx:
Note that I have just added a commit that fixes an issue where `tryToFixInvalidFunctions()` was unable to function correctly when fixing functions for squashed route groups.

Since we were adding just the file name to the `invalidFunctions` map instead of the relative path, when it tried to format the path and check that a matching one exists in the `functionsMap`, it was unable to find one.

Instead, we need to add the relative path to the `invalidFunctions` map so that it is able to check for valid squashed functions in the `functionsMap` properly.

bin/index.js
dario-piotrowicz, line 1: I think this is not correct, when building the app my bin/index.js gets update to the previous version:
Screenshot 2023-05-02 at 14 16 35

the js should be executable, shouldn't it?

src/buildApplication/buildApplication.ts
```diff
@@ -73,58 +71,40 @@ async function prepareAndBuildWorker(
...
-	const nextJsConfigs = await getNextJsConfigs();
-
-	if (nextJsConfigs.basePath) {
-		cliLog(`Using basePath ${nextJsConfigs.basePath}`);
-	}
+	let generatedFunctionsMaps: DirectoryProcessingResults | undefined;
 
 	const functionsDir = resolve('.vercel', 'output', 'functions');
 	if (!(await validateDir(functionsDir))) {
 		cliLog('No functions detected.');
```
dario-piotrowicz, line 78: @james-elicx what do you think, should we change this to be distinct from the `No functions detected` below?

Something like:
```suggestion
		cliLog('No functions detected (Vercel didn't generate a functions directory).');
```
?

tests/templates/routerTestData/basicEdgeAppDir.ts
```diff
@@ -0,0 +1,236 @@
...
+				status: 404,
+				data: '<html>404</html>',
+				headers: {
+					'content-type': 'text/html; charset=utf-8',
+					'x-matched-path': '/404',
+				},
+			},
+		},
+		{
+			name: 'issue 70 - rewrite phase should enter after filesystem rewrites $1 to $1 with check: true',
```
dario-piotrowicz, line 224: `rewrites $1 to $1`?

james-elicx: I guess `$1 to $1` is a little unclear. I can probably improve the test name if you want, perhaps `rewrites a path to the same path`?

For reference, this is the record it relates to.
```
{ src: '/_next/data/(.*)', dest: '/_next/data/$1', check: true }
```

dario-piotrowicz: ah I see, yeah `rewrites a path to the same path` sounds better to me 😄 

(could you also please add a comment or something there? 🙏 )

dario-piotrowicz (approved):
Awesome stuff 🤩 🤩 🤩 

Thanks so very much for this incredible effort/improvement ❤️

Issue #210, opened 2023-05-02
[🐛 Bug]: Failed Vercel build doesn't produce a non-zero result


dario-piotrowicz:
If the `--watch` flag is not used and the Vercel build fails the whole process should exit with a non-zero status (so that potential callers can see that the command failed), this is currently not happening since we sort of just ignore a failed build:
https://github.com/cloudflare/next-on-pages/blob/d7e62c7fb99241565b41a664abfde29bbac99755/src/buildApplication/buildApplication.ts#L48

This should be improved so that a non-zero status is returned instead (meaning that a failed build, without using the `--watch` flag should call `process.exit(X)`).

___

Basically if I have the following nodejs code:
```js
    const { status } = spawnSync("npx", ["@cloudflare/next-on-pages"], {
      cwd: join(appDir),
      stdio: "inherit",
    });
```
and the application in `appDir` causes the Vercel build to error, then `status` should be non-zero (it currently is `0` and we get a non-zero value only if the process fails for different reasons)

PR #215, opened 2023-05-03
fix: Failed Vercel build doesn't produce a non-zero result

tarunrajput:
Resolves: #210

### DIFF
--- a/.changeset/strong-llamas-protect.md
+++ b/.changeset/strong-llamas-protect.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Exit with non-zero status if vercel build fails in non-watch mode
--- a/src/buildApplication/buildApplication.ts
+++ b/src/buildApplication/buildApplication.ts
@@ -29,9 +29,10 @@ export async function buildApplication({
 	skipBuild,
 	disableChunksDedup,
 	disableWorkerMinification,
+	watch,
 }: Pick<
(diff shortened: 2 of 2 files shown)


dario-piotrowicz:
Thanks a lot for the PR @tarunrajput :smiley: 

Sorry just quickly asking you before reviewing the PR, did you check if this also works in watch mode? in watch mode I think we would like to keep the CLI running even if the Vercel build fails (because you can fail the build whilst developing but that shouldn't cause next-on-pages to stop)

By looking at your change I feel like that's not going to be the case :thinking: 

(also please add a patch changeset via `npm run changeset` :pray: :slightly_smiling_face:)

tarunrajput:
> Thanks a lot for the PR @tarunrajput 😃
> 
> Sorry just quickly asking you on small thing before reviewing the PR, did you check if this also works in watch mode? in watch mode I think we would like to keep the CLI running even if the Vercel build fails (because you can fail the build whilst developing but that shouldn't cause next-on-pages to stop)
> 
> By looking at your change I feel like that's not going to be the case 🤔
> 
> (also please add a patch changeset via `npm run changeset` 🙏 🙂)

@dario-piotrowicz, I just pushed a commit, it should work in watch mode now.

dario-piotrowicz (approved):

dario-piotrowicz:
@tarunrajput awesome yeah, all good! thanks again! 😄

PR #216, opened 2023-05-04
Commit to AsyncLocalStorage

### LINKED ISSUE #121: Commit to rely on AsyncLocalStorage on next major release
PR #115 introduced AsyncLocalStorage in a non breaky way, in order to do so the code is more complex than it needs to.

We went for such solution since at this stage we didn't want to create a major release, when we're ready to do so the changes applied in #115 to `src/index.ts` and `templates/_worker.js/index.ts` should be reverted/removed and we should apply a much simpler solution, which consist just in:

Add `AsyncLocalStorage` to `globalThis` in`templates/globals.js`:
```diff
+import { AsyncLocalStorage } from 'node:async_hooks';

globalThis.process = { env: { NODE_ENV: 'production' } };
+globalThis.AsyncLocalStorage = AsyncLocalStorage;
```

Add `node:async_hooks` as an external import in the esbuild build in `src/index.ts`:
```diff
	await build({
		entryPoints: [join(__dirname, '../templates/_worker.js')],
		bundle: true,
		inject: [
			join(__dirname, '../templates/_worker.js/globals.js'),
			functionsFile,
		],
		target: 'es2021',
		platform: 'neutral',
		define: {
			__CONFIG__: JSON.stringify(config),
			__BASE_PATH__: JSON.stringify(basePath),
		},
		outfile: '.vercel/output/static/_worker.js',
		minify: experimentalMinify,
+              external: ['node:async_hooks'],
	});
```

> **Note**
> I believe this should be enough, but it needs to be properly tested out and tweaked if needed

dario-piotrowicz:
resolves #121

### DIFF
--- a/src/buildApplication/generateGlobalJs.ts
+++ b/src/buildApplication/generateGlobalJs.ts
@@ -19,7 +19,5 @@ export function generateGlobalJs(): string {
 					set: (_, property, value) => Reflect.set(__ENV_ALS__.getStore(), property, value),
 			}),
 		};
-	`
-		.replace(/^\s+/, '')
-		.replace(/\n\s+/g, ' ');
+	`;
 }
--- a/src/buildApplication/generateGlobalJs.ts
+++ b/src/buildApplication/generateGlobalJs.ts
@@ -1,5 +1,3 @@
-import { cliWarn } from '../cli';
-
 /**
  * Generates the javascript content (as a plain string) that deals with the global scope that needs to be
  * added to the built worker.
@@ -7,14 +5,11 @@ import { cliWarn } from '../cli';
  * @returns the plain javascript string that should be added at the top of the the _worker.js file
  */
 export function generateGlobalJs(): string {
-	const nodeEnv = getNodeEnv();
-
 	return `
 		import { AsyncLocalStorage } from 'node:async_hooks';
 		globalThis.AsyncLocalStorage = AsyncLocalStorage;
 
 		const __ENV_ALS__ = new AsyncLocalStorage();
-		__ENV_ALS__['NODE_ENV'] = '${nodeEnv}';
 
 		globalThis.process = {
 			env: new Proxy(
@@ -27,35 +22,4 @@ export function generateGlobalJs(): string {
 	`
 		.replace(/^\s+/, '')
 		.replace(/\n\s+/g, ' ');
-}
-
-enum NextJsNodeEnv {
-	PRODUCTION = 'production',
-	DEVELOPMENT = 'development',
-	TEST = 'test',
-}
-
-function getNodeEnv(): string {
-	const processNodeEnv = process.env.NODE_ENV;
-
-	if (!processNodeEnv) {
-		return NextJsNodeEnv.PRODUCTION;
-	}
-
-	const nextJsNodeEnvs = Object.values(NextJsNodeEnv);
-	if (!(nextJsNodeEnvs as string[]).includes(processNodeEnv)) {
-		cliWarn(
-			`
-			WARNING:
-			    The current value of the environment variable NODE_ENV is "${processNodeEnv}",
-			    but the supported values are: ${nextJsNodeEnvs
-						.map(env => `"${env}"`)
-						.join(', ')}.
-			    See: https://nextjs.org/docs/basic-features/environment-variables.
-		`,
-			{ spaced: true }
-		);
-	}
-
-	return processNodeEnv;
 }
--- a/src/buildApplication/generateGlobalJs.ts
+++ b/src/buildApplication/generateGlobalJs.ts
@@ -8,7 +8,25 @@ import { cliWarn } from '../cli';
  */
 export function generateGlobalJs(): string {
 	const nodeEnv = getNodeEnv();
-	return `globalThis.process = { env: { NODE_ENV: '${nodeEnv}' } };`;
+
+	return `
+		import { AsyncLocalStorage } from 'node:async_hooks';
+		globalThis.AsyncLocalStorage = AsyncLocalStorage;
+
+		const __ENV_ALS__ = new AsyncLocalStorage();
+		__ENV_ALS__['NODE_ENV'] = '${nodeEnv}';
+
+		globalThis.process = {
+			env: new Proxy(
+				{},
+				{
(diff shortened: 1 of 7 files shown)


tests/src/buildApplication/generateGlobalJs.test.ts
GregBrimble, line 1: It would be cool to have a test which actually evaluates the ALS store. That might be outwith the capabilities of the current test setup, but if you can think of a way to actually run the code, that might be worthwhile.

src/buildApplication/generateGlobalJs.ts
```diff
@@ -8,7 +8,25 @@ import { cliWarn } from '../cli';
...
+			env: new Proxy(
+				{},
+				{
+					get: (_, property) => Reflect.get(__ENV_ALS__.getStore(), property),
+					set: (_, property, value) => Reflect.set(__ENV_ALS__.getStore(), property, value),
+			}),
+		};
+	`
+		.replace(/^\s+/, '')
+		.replace(/\n\s+/g, ' ');
```
GregBrimble, line 29:
```suggestion
	`;
```

Wouldn't bother with stuff like this. It just confuses any reader of the code for basically no benefit.

src/buildApplication/generateGlobalJs.ts
```diff
@@ -8,7 +8,25 @@ import { cliWarn } from '../cli';
...
 export function generateGlobalJs(): string {
 	const nodeEnv = getNodeEnv();
-	return `globalThis.process = { env: { NODE_ENV: '${nodeEnv}' } };`;
+
+	return `
+		import { AsyncLocalStorage } from 'node:async_hooks';
+		globalThis.AsyncLocalStorage = AsyncLocalStorage;
+
+		const __ENV_ALS__ = new AsyncLocalStorage();
+		__ENV_ALS__['NODE_ENV'] = '${nodeEnv}';
```
GregBrimble, line 17: Is that definitely correct? I'm still getting familiar with ALS APIs, but it seems a bit weird to be setting a value directly on the ALS instance here, rather than on the store.

Does `__ENV_ALS__.getStore()` return something that references back to props assigned on the `__ENV_ALS__` instance somehow?

Double-check that `process.env.NODE_ENV` and `process.env.SOME_OTHER_SET_ENV_VAR` are set properly in an example project.

src/buildApplication/generateGlobalJs.ts
dario-piotrowicz: 😕

src/buildApplication/generateGlobalJs.ts
dario-piotrowicz: nono, to be totally clear, `__ENV_ALS__['NODE_ENV']` is a hack, I just wanted to have the `NODE_ENV` available and didn't want to create a new variable for it, so I am just piggybacking it on top of the ASL (as a plain object)

(
you can see here: https://github.com/cloudflare/next-on-pages/blob/cbc3395270816c20a4f60f980e6679e11cc4788b/templates/_worker.js/index.ts#L15 that as soon as I can, I take it and put it in the actual store
)

Alternatively I can have a `__NODE_ENV__` variable instead (I don't think I can use it in the ASL initialization)

dario-piotrowicz: PS: I did check that I get all the `process.env` variables in an example project 😅 (specifically here: https://github.com/dario-piotrowicz/next-apps-for-testing/blob/802e9478a3b322f9b2850008aec9e4abd7bc8215/apps/simple-app-dir-13.3.1/app/api/hello/route.js)

GregBrimble: ohhh... yeah, I am not a fan of that, I'm afraid. Why not just do `__ENV_ALS__.run({ ...env, NODE_ENV: ${nodeEnv} })`?

GregBrimble: Ah, because it's in a separate unprocessed file. I'd maybe use esbuild's `define` feature then.

`__ENV_ALS__.run({ ...env, NODE_ENV: __NODE_ENV__ })`

and `define({ __NODE_ENV__: "production" })` in the esbuild command.

GregBrimble: Ah, in fact you've changed it even more. I'm just now seeing `getNodeEnv()` too.

Okay, whatever. Last suggestion would be to put it in the proxy maybe?

```
get (_, property) => {
  if (property === 'NODE_ENV') {
    return ${nodeEnv}
  } else {
    return Reflect.get(__ENV_ALS__.getStore(), property)
  }
}
```

But not the end of the world if you ship what you have right now.

dario-piotrowicz: test...

tests/src/buildApplication/generateGlobalJs.test.ts
dario-piotrowicz: sorry I can't test it with eval because if fails because of the ALS import:
Screenshot 2023-05-09 at 13 21 29

___

In a followup PR I want to change the import to be dynamic anyways (like so: https://github.com/cloudflare/next-on-pages/commit/42c6b7983768e79cd1cb2ea6ed9bcbb2bd732bc0) so I can try to unit test this then?

___

(if all fails we will still have the e2e? 😅)

src/buildApplication/generateGlobalJs.ts
dario-piotrowicz: done! (https://github.com/cloudflare/next-on-pages/pull/216/commits/97b1afb1501ca5c1dbe6692335c9b9abd4859926)

Thank you so very much for the suggestion! 🙏  (and sorry for the silly mistake 😓)

GregBrimble (approved):

templates/_worker.js/index.ts
```diff
@@ -1,24 +1,31 @@
 import { handleRequest } from './handleRequest';
 import { adjustRequestForVercel } from './utils';
+import type { AsyncLocalStorage } from 'node:async_hooks';
+
+declare const NODE_ENV: string;
```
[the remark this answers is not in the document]
dario-piotrowicz: done! 😃

PR #217, opened 2023-05-04
remove Router class and define handleRequest function instead, also addd RequestContext type

dario-piotrowicz:


tests/templates/handleRequest.ts
james-elicx, line 1: This file needs to be renamed `handleRequest.test.ts`! Everything else looks good though :)

dario-piotrowicz: 🤦 thanks for noticing it 🙂

dario-piotrowicz: fixed 🙂

jculvey (approved):

PR #222, opened 2023-05-06
fix: prevent infinite loops during routing edge case

james-elicx:
This PR does the following:
- Introduces a counter for keeping track of how many times the `checkPhase` function is called for each run of the route matcher.
  - Returns a 500 error after the counter exceeds 50 loops - this indicates infinite looping.
- Changes the `check: true` behaviour for the `miss` phase.
  - If the path is not in the build output, it enters the `filesystem` phase as the `none` phase may cause infinite loops with i18n.
  - If a path is in the build output, it resets the status code if it was previously set to 404, and continues with routing like normal.
- Unintentionally introduces basic i18n support.

With i18n, there is a rule in the `none` phase that maps a request from `/*` to `/{default-locale}/*`. At the same time, there is a rule in the `miss` phase to map requests to `/{locale}/*` to `/*`. Because of this, when we encounter `check: true` in the `miss` phase, we need to change the default behaviour to act as detailed above.

Although infinite loops when checking phases *shouldn't* happen, this edge case proves it's possible, hence why the counter has been introduced.

Anyway, onto the exciting part! This fix unintentionally introduces basic support for internationalized routing with the `i18n` option in `next.config.js`! There is still some additional work to do to get full i18n support, so I'm not doing this PR as an i18n one - that'll come when I finish the other bits for i18n and write tests.

### DIFF
--- a/.changeset/unlucky-chicken-join.md
+++ b/.changeset/unlucky-chicken-join.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Prevent infinite loops from occuring when checking phases during routing.
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -33,6 +33,9 @@ export class RoutesMatcher {
 	/** Search params for the response object */
 	public searchParams: URLSearchParams;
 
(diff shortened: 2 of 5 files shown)


dario-piotrowicz (approved):
Awesome stuff as always! :star_struck:

Issue #199, opened 2023-04-25
[🐛 Bug]: Nextjs v13.3.1 fresh install worker threw exception


hjaber:
### Cli version

0.10.0

### next-on-pages environment related information

```
System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 22.4.0: Mon Mar  6 21:01:02 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8112
        CPU: (8) arm64 Apple M2
        Memory: 24 GB
        Shell: /bin/zsh
Binaries:
        Node: 18.12.1
        Yarn: N/A
        npm: 9.6.5
        pnpm: N/A
Package Manager Used: npm
Relevant Packages:
        @cloudflare/next-on-pages: 0.10.0
        vercel: N/A
        next: 13.3.1
```

### Description

I see that issue #186 is already merged but I'm having this exception thrown in prod

```Error 1101 Worker threw exception```

```json
{
  "outcome": "exception",
  "scriptName": "pages-worker--1029254-production",
  "exceptions": [
    {
      "name": "Error",
      "message": "Dynamic require of \"node:buffer\" is not supported",
      "timestamp": 1682397982459
    }
  ],
  "logs": [],
  "eventTimestamp": 1682397982459,
  "event": {
    "request": {
      "url": "https://08541378.mech.pages.dev/",
      "method": "GET",
      "headers": {
        "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8",
        "accept-encoding": "gzip",
        "accept-language": "en-US,en;q=0.8",
        "cache-control": "max-age=0",
        "cf-connecting-ip": "107.199.200.137",
        "cf-ipcountry": "US",
        "cf-ray": "7bd3f11e5f0ecfa8",
        "cf-visitor": "{\"scheme\":\"https\"}",
        "connection": "Keep-Alive",
        "dnt": "1",
        "host": "08541378.mech.pages.dev",
        "priority": "u=0, i",
        "sec-ch-ua": "\"Chromium\";v=\"112\", \"Brave\";v=\"112\", \"Not:A-Brand\";v=\"99\"",
        "sec-ch-ua-mobile": "?0",
        "sec-ch-ua-platform": "\"macOS\"",
        "sec-fetch-dest": "document",
        "sec-fetch-mode": "navigate",
        "sec-fetch-site": "cross-site",
[...]

PR #223, opened 2023-05-07
Update README.md

mountainash:
Updated the config formatting for the latest version of NextJS as per https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#runtime

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,10 @@ cd my-next-app
 
 ### 2. Configure the application to use the Edge Runtime
 
-In order for your application to run on Cloudflare Pages, it needs to be set to use the Edge Runtime. Make sure that all the files in your application containing server-side code (e.g. any API Routes and any pages which use `getServerSideProps`) export a `config` object specifying the use of the Edge Runtime:
+In order for your application to run on Cloudflare Pages, it needs to be set to use the Edge Runtime. Make sure that all the files in your application containing server-side code (e.g. any API Routes and any pages which use `getServerSideProps`) export a [Runtime Route Segement config](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#runtime) specifying the use of the Edge Runtime:
 
 ```ts
-export const config = {
-	runtime: 'edge',
-};
+export const runtime = 'edge';
 ```
 
 Additionally, ensure that your application is not using any [unsupported APIs](https://nextjs.org/docs/api-reference/edge-runtime#unsupported-apis) and that its API routes are defined as [Edge API Routes](https://nextjs.org/docs/api-routes/edge-api-routes).


dario-piotrowicz (commented):
Awesome, thank you so very much for the update @mountainash :slightly_smiling_face: 

I'm afraid however that people using an older version of Next.js could want to use next-on-pages (we still have to decide how we want to handle Next.js versions support :sweat:)

Could you also add a note somewhere that if people are using versions prior to `13.x.y` then they need to export a config object instead?
(where we put the correct `x` and `y` which on the stop I am not sure which are :sweat: do you know from which version they removed the `config`? I think `13.3.1` but I am not too sure about it)

dario-piotrowicz:
@mountainash could you please also update these two? :pray: 

https://github.com/cloudflare/next-on-pages/blob/701f0c257ceed22c19ff85fce8597351a47ef69f/docs/examples.md?plain=1#L29-L31

https://github.com/cloudflare/next-on-pages/blob/701f0c257ceed22c19ff85fce8597351a47ef69f/docs/examples.md?plain=1#L57-L59

mountainash:
Thanks for the review @dario-piotrowicz, but I can't get a deployment working yet. I'll keep trying before making these changes (as they may not be necessary if Pages can't support the newer versions of Next.JS).

dario-piotrowicz:
@mountainash if you can't deploy an app please open an issue or create a discussion so we can help figure it out :slightly_smiling_face:, next-on-pages should be able to support the latest Next.js versions

PS: There is currently an issue which prevents Next.js apps to be deployed to Cloudflare Pages (because a node compatibility related issue), this should be solved by the next wrangler release (which I think/hope will happen tomorrow), you check if you're hitting that issue by checking your console logs and see if the build fails because of an import from `node:buffer`? (as mentioned in this issue: https://github.com/cloudflare/next-on-pages/issues/199)

mountainash:
This is **the** reason for the build failure. I'll wait for the next Release before pushing ahead with updates to the Docs and error messages.

dario-piotrowicz:
@mountainash I'm so sorry for the inconvenience :sweat: 

If you want you can manually deploy you app via wrangler by using this [prerelease version](https://github.com/cloudflare/workers-sdk/pull/3133#issuecomment-1533440046), or yeah just wait for wrangler to get released (and that should fix the Pages CI without you needing to do anything)

mountainash:
> (we still have to decide how we want to handle Next.js versions support 😓)

Is this being discussed somewhere by the Cloudflare team? It's a pretty important issue (and esp. while NextJS/Vercel has so many alpha/beta releases in-progress which contain many "low level" [infra/config] changes).

If it's something that the "next-on-pages" team is about to solve - it might change the implementation (double up of info) of this PR.

dario-piotrowicz:
> > (we still have to decide how we want to handle Next.js versions support 😓)
> 
> Is this being discussed somewhere by the Cloudflare team? It's a pretty important issue (and esp. while NextJS/Vercel has so many alpha/beta releases in-progress which contain many "low level" [infra/config] changes).
> 
> If it's something that the "next-on-pages" team is about to solve - it might change the implementation (double up of info) of this PR.

Thanks for your feedback I totally get it and you're right, however no, we haven't discussed it too much, we did have some discussions and have some ideas but we are yet to make any decision (we'll need to do this soon I think, but I don't realistically think is going to happen too soon (mainly for external reasons)).

However the current strategy is that:
 - we don't specify any support for alpha/beta releases (and whatever strategy we'll adapt I doubt we'll ever properly support such versions as that could prove exceedingly costly/difficult/unmaintainable),
 - we support all Next.js 13 (`so 13.x.y`) (including `latest`),
 - and we also support Next.js 12 (`12.x.y`) versions (since the adapter has been implemented when Next was at version 12 and it's always worked with it)
 - versions 11 and earlier we don't support them (meaning that we aren't even looking into them, the adapter could actually work with them but I have no idea 🤷)
 
 I hope this helps, as you can see we sort of/kind of have a strategy in place right now, but it's not the best and it'll need to be improved

dario-piotrowicz:
ah by the way @mountainash wrangler has been updated, please have a look and check if now it works for you (if you haven't done already 🙂)

dario-piotrowicz:
@mountainash the github Pages integration has also been fixed (I just checked) 🙂👍

dario-piotrowicz:
Closing this PR as the readme has been updated in https://github.com/cloudflare/next-on-pages/pull/245

Sorry about that @mountainash but thank you very much for your PR anyhow 🙏

PR #227, opened 2023-05-08
feat: i18n locale redirects support

### LINKED ISSUE #28: i18n with `next-intl` fails to register routes
I tried using next-intl with next-on-pages, but I'm getting the message below:
```
⚡️ ERROR: Failed to produce a Cloudflare Pages build from the project.
⚡️ The following functions were not configured to run with the Edge Runtime:
⚡️
⚡️  - de
⚡️  - en
```
Seems like the dynamic routes don't get picked up

### LINKED ISSUE #37: Pages not created when using Localize / Internationalize on NextJS
I'm using the follow configuration to use internationalization on my app:

1) on next.config.js:

```
const nextTranslate = require('next-translate')

const nextConfig = {
    experimental: {
        runtime: 'experimental-edge',
    },

module.exports = withPlugins(
    [nextTranslate, ..],
    nextConfig,
);
```

2)i18n.js file: 
```
module.exports = {
    locales: ['en-IE'],
    defaultLocale: 'en-IE',
    pages: {
        '*': ['common'],
    },
    loadLocaleFrom: (lang, ns) => {
        const translation = lang.slice(0, 2);

        return import(`./locales/${translation}/${ns}.json`).then(
            (m) => m.default,
        );
    },
};
```

3)   NextJS version:
` "next": "12.2.6"`

Deployments are successful on Deploying to Cloudflare global network messages (no errors), but in the end, pages are not created for every single language.

### LINKED ISSUE #157: [🐛 Bug]: Nextra i18n Build Does Not Show Anything
### Cli version

0.7.0

### Next.js related information

13.2.4

### Vercel version

28.18.3

### Description

Nextra will success compiling, but will return an empty html.

### Reproduction

github.com/shuding/nextra

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Fixes an issue where redirects weren't breaking out of routing at the right point.
- Adds a utility function for parsing the `accept-language` header value.
- Introduces support for the source route `locale` option and the `next.config.js` option for `i18n`.

It is somewhat of a continuation of https://github.com/cloudflare/next-on-pages/pull/222 as that fixed a major blocker for i18n.

So, there is a little quirk with how the locale redirects have to be handled. There are two different types of ways internationalization works. These are sub-path routing and domain routing. The build output config `route.src` value for sub-path routing is simply `/`, which causes issues.

The Next.js [docs](https://nextjs.org/docs/pages/building-your-application/routing/internationalization#automatic-locale-detection) state that automatic locale detection occurs when a user visits the root of the website, which would mean that the regex should be `^/$`, except it isn't. Therefore, we have to add custom behaviour to check whether the `route.src` is a regex, and if it isn't, check that it is equal to the current path before determining whether we should apply the locale redirects.

Sub-path routing entry:
```
{
	src: '/',
	locale: {
		redirect: { en: '/', fr: '/fr', nl: '/nl', es: '/es' },
		cookie: 'NEXT_LOCALE',
	},
	continue: true,
}
```

Domain routing entry:
```
{
	src: '^//?(?:en|fr|nl|es)?/?$',
	locale: {
		redirect: { es: 'https://example.es/' },
		cookie: 'NEXT_LOCALE',
	},
	continue: true,
}
```

As you can see, the two values for `route.src` are very different. I find this to be rather bizarre behaviour, and to me, it looks like the Vercel build output isn't generating the correct matching patterns. Anyway, so to address the quirks, I have done what I detailed above.

Also tested it using [test apps](https://github.com/dario-piotrowicz/next-apps-for-testing) for plain Next.js i18n, next-translate, and next-i18n. The test app shouldn't actually matter anyway, because they all use next.config.js i18n under the hood, which is what this PR introduces better support for.

closes #28
closes #37
closes #157

### DIFF
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -1,5 +1,6 @@
 import { parse } from 'cookie';
 import type { MatchPCREResult, MatchedSet } from './utils';
+import { parseAcceptLanguage } from './utils';
 import {
 	applyHeaders,
 	applyPCREMatches,
@@ -296,6 +297,54 @@ export class RoutesMatcher {
 	}
 
 	/**
+	 * Applies the route's redirects for locales and internationalization.
+	 *
+	 * @param route Build output config source route.
+	 */
+	private applyLocaleRedirects(route: VercelSource): void {
+		if (!route.locale?.redirect) return;
+
+		// Automatic locale detection is only supposed to occur at the root. However, the build output
+		// sometimes uses `/` as the regex instead of `^/$`. So, we should check if the `route.src` is
+		// equal to the path if it is not a regular expression, to determine if we are at the root.
+		// https://nextjs.org/docs/pages/building-your-application/routing/internationalization#automatic-locale-detection
+		const srcIsRegex = /^\^(.)*$/.test(route.src);
+		if (!srcIsRegex && route.src !== this.path) return;
+
+		// If we already have a location header set, we might have found a locale redirect earlier.
+		if (this.headers.normal.has('location')) return;
+
+		const {
+			locale: { redirect: redirects, cookie: cookieName },
+		} = route;
+
+		const cookieValue = cookieName ? this.cookies[cookieName] : '';
+		const cookieLocales = parseAcceptLanguage(cookieValue);
+
+		const headerLocales = parseAcceptLanguage(
+			this.reqCtx.request.headers.get('accept-language')
+		);
+
+		// Locales from the cookie take precedence over the header.
+		const locales = [...cookieLocales, ...headerLocales];
+
+		for (const locale of locales) {
+			const redirectValue = redirects[locale];
+			if (!redirectValue) continue;
+
+			// If the path starts with the redirect, we should already be in the right place. Bail out.
+			if (this.path.startsWith(redirectValue)) {
+				return;
+			}
+
+			// Redirect found, set the location header and bail out.
+			this.headers.normal.set('location', redirectValue);
+			this.status = 307;
+			return;
+		}
+	}
+
+	/**
 	 * Checks a route to see if it matches the current request.
 	 *
 	 * @param phase Current phase of the routing process.
@@ -315,6 +364,9 @@ export class RoutesMatcher {
 
 		// If this route overrides, replace the response headers and status.
 		this.applyRouteOverrides(route);
+
+		// If this route has a locale, apply the redirects for it.
+		this.applyLocaleRedirects(route);
 
 		// Call and process the middleware if this is a middleware route.
 		const success = await this.runRouteMiddleware(route.middlewarePath);
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -323,8 +323,8 @@ export class RoutesMatcher {
 			locale: { redirect: redirects, cookie: cookieName },
 		} = route;
 
-		const cookieValue = cookieName ? this.cookies[cookieName] ?? '' : '';
-		const cookieLocales = parseAcceptLanguage(cookieValue);
+		const cookieValue = cookieName && this.cookies[cookieName];
+		const cookieLocales = parseAcceptLanguage(cookieValue ?? '');
 
 		const headerLocales = parseAcceptLanguage(
 			this.reqCtx.request.headers.get('accept-language') ?? ''
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -43,9 +43,7 @@ export class RoutesMatcher {
 	 *
 	 * @param routes The processed Vercel build output config routes.
 	 * @param output Vercel build output.
-	 * @param assets Static assets fetcher.
-	 * @param ctx Execution context.
-	 * @param req Request object.
+	 * @param reqCtx Request context object; request object, assets fetcher, and execution context.
 	 * @param prevMatch The previous match from a routing phase to initialize the matcher with.
 	 * @returns The matched set of path, status, headers, and search params.
 	 */
@@ -405,7 +403,7 @@ export class RoutesMatcher {
 		if (
 			phase === 'hit' ||
 			isUrl(this.path) ||
-			(!shouldContinue && this.headers.normal.has('location'))
+			this.headers.normal.has('location')
 		) {
 			return 'done';
 		}
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -328,18 +328,15 @@ export class RoutesMatcher {
 		// Locales from the cookie take precedence over the header.
 		const locales = [...cookieLocales, ...headerLocales];
 
-		for (const locale of locales) {
-			const redirectValue = redirects[locale];
-			if (!redirectValue) continue;
+		const redirectLocales = locales.map(locale => redirects[locale]).filter(Boolean) as string[];
 
-			// If the path starts with the redirect, we should already be in the right place. Bail out.
-			if (this.path.startsWith(redirectValue)) {
-				return;
+		const redirectValue = redirectLocales[0];
+		if (redirectValue) {
+			const needsRedirecting = !this.path.startsWith(redirectValue);
+			if (needsRedirecting) {
+				this.headers.normal.set('location', redirectValue);
+				this.status = 307;
 			}
-
-			// Redirect found, set the location header and bail out.
-			this.headers.normal.set('location', redirectValue);
-			this.status = 307;
 			return;
 		}
 	}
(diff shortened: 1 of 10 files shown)


tests/templates/requestTestData/middleware.ts
```diff
@@ -93,7 +93,6 @@ export const testSet: TestSet = {
 				data: '',
 				headers: {
 					location: 'http://localhost/somewhere-else?redirect=',
-					'x-matched-path': '/api/hello',
```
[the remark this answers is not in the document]
james-elicx: this is related to fixing not exiting routing to make a redirect when continue is true (first commit). fix was in part to ensure that redirects are entered properly for i18n, but also just because I'm pretty sure that's how I should have done it in the first place.

tests/templates/requestTestData/basicEdgeAppDir.ts
```diff
@@ -166,9 +166,9 @@ export const testSet: TestSet = {
 			},
 		},
 		{
-			name: 'trailing slash redirects to non-trailing slash',
+			name: 'trailing slash redirects (308) to non-trailing slash',
 			paths: ['/api/hello/'],
-			expected: { status: 307, data: '', headers: { location: '/api/hello' } },
+			expected: { status: 308, data: '', headers: { location: '/api/hello' } },
```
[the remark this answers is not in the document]
james-elicx: this is from the first commit which was a fix to when redirects cause exiting the routing system. 308 is actually the correct status code that it should use in the build output config, which is further reason for the change to when it exits routing in that commit being necessary

templates/_worker.js/utils/http.ts
```diff
@@ -87,3 +87,23 @@ export function createRouteRequest(req: Request, path: string) {
...
+ *
+ * @param headerValue Accept-Language header value.
+ * @returns Array of locales sorted by quality.
+ */
+export function parseAcceptLanguage(headerValue?: string | null): string[] {
+	return (headerValue ?? '')
+		.split(',')
+		.map(val => {
+			const [lang, qual] = val.split(';') as [string, string | undefined];
+			const quality = parseFloat((qual ?? 'q=1').replace(/q ?= ?/gi, ''));
```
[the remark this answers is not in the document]
james-elicx: there shouldn't be any realistically but you never know 😅

tests/templates/requestTestData/middleware.ts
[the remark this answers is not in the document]
dario-piotrowicz: awesome, thanks for clarifying this :slightly_smiling_face::+1:

james-elicx:
We identified an issue where there is a record in the `miss` phase that rewrites all requests to the default locale to `/`.

To prevent this behaviour that we don't want, we modify the source route's `src` regex to be locale friendly using any of the previously found locales.

Problematic record example:
```
{
  "src": "/en",
  "dest": "/",
  "check": true
},
```

This should instead have an `src` value of `^/en$`, so that it only remaps paths for exactly the default locales to `/`.

templates/_worker.js/utils/http.ts
```diff
@@ -87,3 +87,23 @@ export function createRouteRequest(req: Request, path: string) {
...
 	return new Response(resp.body, resp);
 }
+
+/**
+ * Parses the Accept-Language header value and returns an array of locales sorted by quality.
+ *
+ * @param headerValue Accept-Language header value.
+ * @returns Array of locales sorted by quality.
+ */
+export function parseAcceptLanguage(headerValue?: string | null): string[] {
```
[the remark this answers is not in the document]
dario-piotrowicz: did you forget to push stuff up? or do you disagree with the suggestion?

tests/templates/utils/http.test.ts
```diff
@@ -94,3 +95,19 @@ describe('createRouteRequest', () => {
...
+			['en-US,en', ['en-US', 'en']],
+			['en-US,en;q=0.9,es;q=0.8', ['en-US', 'en', 'es']],
+			['en-US,fr;q=0.7,en;q=0.9,es;q=0.8', ['en-US', 'en', 'es', 'fr']],
+			['fr;q=0.7,en;q=0.9,en-US,es;q=0.8', ['en-US', 'en', 'es', 'fr']],
+			['fr;q = 0.7,en;q =0.9,en-US,es;q= 0.8', ['en-US', 'en', 'es', 'fr']],
+		].forEach(([input, output]) => {
+			expect(parseAcceptLanguage(input as string)).toEqual(output);
+		});
+	});
+});
```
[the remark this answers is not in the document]
dario-piotrowicz: :cry:

templates/_worker.js/routes-matcher.ts
```diff
@@ -297,6 +298,87 @@ export class RoutesMatcher {
...
+		}
+	}
+
+	/**
+	 * Modifies the source route's `src` regex to be friendly with previously found locale's in the
+	 * `miss` phase.
+	 *
+	 * Sometimes, there is a source route with `src: '/{locale}'`, which rewrites all paths containing
+	 * the locale to `/`. This is problematic for matching, and should only do this if the path is
+	 * exactly the locale, i.e. `^/{locale}$`.
```
dario-piotrowicz, line 357:
```suggestion
	 * Sometimes, there is a source route with `src: '/{locale}'`, which rewrites all paths containing
	 * the locale to `/`. This is problematic for matching, and should only do this if the path is
	 * exactly the locale, i.e. `^/{locale}$` (since as indicated in the Next.js docs:
	 * https://nextjs.org/docs/pages/building-your-application/routing/internationalization#automatic-locale-detection
	 * and as we've confirmed via manual testing, only requests to the application's root should get redirected)
```

too verbose? :thinking:

templates/_worker.js/routes-matcher.ts
```diff
@@ -297,6 +298,87 @@ export class RoutesMatcher {
...
+
+		// If we already have a location header set, we might have found a locale redirect earlier.
+		if (this.headers.normal.has('location')) return;
+
+		const {
+			locale: { redirect: redirects, cookie: cookieName },
+		} = route;
+
+		const cookieValue = cookieName ? this.cookies[cookieName] ?? '' : '';
+		const cookieLocales = parseAcceptLanguage(cookieValue);
```
dario-piotrowicz, line 327: nit:

```suggestion
		const cookieValue = cookieName && this.cookies[cookieName];
		const cookieLocales = parseAcceptLanguage(cookieValue ?? '');
```

(if you prefer your current version I am 100% fine with it, I was just thinking that the `?? '' : ''` can look a bit unclear :thinking:)

templates/_worker.js/routes-matcher.ts
dario-piotrowicz: (PS: I might also be confusing this with the issue below? :thinking:)

templates/_worker.js/routes-matcher.ts
james-elicx: Shouldn't that be `||` instead?

templates/_worker.js/routes-matcher.ts
james-elicx: I don't actually think this is the right place to use that, as that pertains to the *hack* done for locale source routes from the `none` phase, meanwhile this is a *hack* to prevent it remapping to the wrong path with the stripped locale for matching non-locale assets/routes, if that makes sense.

templates/_worker.js/routes-matcher.ts
dario-piotrowicz: the `&&` you mean?

I don't think so, if `cookieName` is falsy (say `undefined`) then we get:
`const cookieValue = undefined && this.cookies[undefined]` which results in `undefined`

if `cookieName` is truhy (say `'str'`) then we get:
`const cookieValue = 'str' && this.cookies['str']` which results in `this.cookies['str']`

isn't this correct? :thinking:

templates/_worker.js/routes-matcher.ts
dario-piotrowicz: yeah makes sense, thanks :slightly_smiling_face::+1:

templates/_worker.js/routes-matcher.ts
james-elicx: The ?? not the &&

dario-piotrowicz: lol sorry :sweat_smile:

templates/_worker.js/utils/http.ts
```diff
@@ -87,3 +87,23 @@ export function createRouteRequest(req: Request, path: string) {
...
 }
+
+/**
+ * Parses the Accept-Language header value and returns an array of locales sorted by quality.
+ *
+ * @param headerValue Accept-Language header value.
+ * @returns Array of locales sorted by quality.
+ */
+export function parseAcceptLanguage(headerValue: string): string[] {
+	return (headerValue ?? '')
```
dario-piotrowicz, line 98:
```suggestion
	return headerValue
```

dario-piotrowicz (approved):
So awesome!!! thank you very much! :heart:

PR #232, opened 2023-05-10
fix entrypoint dynamic imports being always run

dario-piotrowicz:


src/buildApplication/generateGlobalJs.ts
```diff
@@ -15,7 +15,15 @@ export function generateGlobalJs(): string {
...
 				{},
 				{
-					get: (_, property) => Reflect.get(__ENV_ALS__.getStore(), property),
+					get: (_, property) => {
+						${/* TODO: remove try-catch ASAP (after runtime fix) @dario */ ''}
+						try {
+							const result = Reflect.get(__ENV_ALS__.getStore(), property);
+							return result;
+						} catch (e) {
+							return undefined;
```
dario-piotrowicz, line 24: @GregBrimble this is kind of broken.... I don't think we can do much better (but the runtime fix should come quite soon, this or next week)

(it's either this or reverting the ALS PR I think)

tests/templates/handleRequest.test.ts
```diff
@@ -1,115 +1,124 @@
...
-		assetsFetcher,
-		ctx: {} as ExecutionContext,
-	};
+// /**
+//  * Runs a test set.
+//  *
+//  * @param testSet Test set to run.
+//  */
+// async function runTestSet({ config, files, testCases }: TestSet) {
+// 	const { vercelConfig, buildOutput, assetsFetcher } =
```
james-elicx, line 82:
```suggestion
// 	const { vercelConfig, buildOutput, assetsFetcher, restoreMocks } =
```

tests/templates/handleRequest.test.ts
```diff
@@ -1,115 +1,124 @@
...
-		build: (options: { stdin?: { contents: string }; outfile: string }) => {
-			const contents = options.stdin?.contents ?? 'built code';
-			writeFile(options.outfile, contents);
-		},
-	};
-});
+// 	testCases.forEach(testCase =>
+// 		runTestCase(reqCtx, vercelConfig, buildOutput, testCase)
+// 	);
+// }
```
james-elicx, line 93:
```suggestion
// 	);

// 	afterAll(() => restoreMocks());
// }
```

tests/templates/handleRequest.test.ts
```diff
@@ -1,115 +1,124 @@
...
-	i18nTestSet,
-	infiniteLoopTestSet,
-	middlewareTestSet,
-} from './requestTestData';
-import type { TestCase, TestSet } from '../_helpers';
-import { createRouterTestData } from '../_helpers';
-import type { RequestContext } from '../../src/utils/requestContext';
-import { handleRequest } from '../../templates/_worker.js/handleRequest';
-import { writeFile } from 'fs/promises';
+// import { describe, expect, suite, test, vi } from 'vitest';
```
james-elicx, line 1:
```suggestion
// import { afterAll, describe, expect, suite, test, vi } from 'vitest';
```

james-elicx (commented):
I've added suggestions for changes that will fix the routing test stuff I wrote to work with your changes.

I didn't add a suggestion to uncomment everything because that would just make all my comments really painful to read through.

If you apply them all in bulk from the GitHub web view, and then go into the two files and make an additional commit to uncomment everything, it should all work again.

In essence, it's just changing it to mock the import instead.

dario-piotrowicz:
> I've added suggestions for changes that will fix the routing test stuff I wrote to work with your changes.
> 
> I didn't add a suggestion to uncomment everything because that would just make all my comments really painful to read through.
> 
> If you apply them all in bulk from the GitHub web view, and then go into the two files and make an additional commit to uncomment everything, it should all work again.
> 
> In essence, it's just changing it to mock the import instead.

Thanks a lot! ❤️ 

Very considerate of you 😄

Cool, I've applied all the changes seems all good 😃👍 

(PS: sorry I gave it a quick try yesterday but thought that this needed a bigger refactoring to work, I actually didn't think of trying `vi.mock` 😓)

templates/_worker.js/utils/routing.ts
```diff
@@ -70,8 +70,10 @@ export async function runOrFetchBuildOutputItem(
 
 	try {
 		switch (item?.type) {
-			case 'function': {
-				resp = await (await item.entrypoint).default(req, ctx);
+			case 'function':
+			case 'middleware': {
+				const edgeFunction: EdgeFunction = await import(item.entrypoint);
```
GregBrimble, line 75: I checked with `wrangler dev --experimental-local`, and this **does not** re-evaluate the module every time. It only runs evaluates it the first time it gets imported. This is good!

tests/_helpers/index.ts
GregBrimble, line 1: Damn, these are some cool tests that we've got now. Nice work, @dario-piotrowicz / James! :)

dario-piotrowicz: yeah I can't take the credit, it's all @james-elicx doing here 😄

(although I mentioned to him and think that (some of?) these __might__ be more suited as e2es, as I think the raw config snapshots we have are a bit tricky to read/maintain/update)

GregBrimble (approved):
LGTM

tests/_helpers/index.ts
james-elicx: > Damn, these are some cool tests that we've got now. Nice work, @dario-piotrowicz / James! :)

Thank you, my pleasure 🙂

PR #233, opened 2023-05-11
introduce `cloudflare` platform object containing all the cloudflare context relative to the request

dario-piotrowicz:


GregBrimble (changes requested):
I don't think we should move `env` away from `process.env`. This means apps won't "just work" when deployed to Pages.

I'm not sold on `cf` or `ctx` either, but don't really have any meaningfully different ideas for those.

dario-piotrowicz:
> I don't think we should move `env` away from `process.env`. This means apps won't "just work" when deployed to Pages.

ah.... I guess for `env` you do have a point 😕

I'm not sure what you don't like about `cf` and `ctx` 😢 (we should however have a way to expose them to the users)
we could also move them directly into the `process.env` but I am afraid of name conflicts (unless we call them something else)

dario-piotrowicz:
After discussing it offline with @GregBrimble we decided to try something different, so I am closing this PR

shiqimei:
> 

Yeah, why hide `cf` from users?

dario-piotrowicz:
@shiqimei thank you for your comment 🙂 

I will bring this to the team and see if we can decide, it does sound like it would be a valuable change (especially for now and we can change things later in a major version if need be)

One small thing to note is that your solution does introduce a sort of breaking change because if users has a binding called `CloudflareContext` it would get overridden by yours... but yeah that sounds extremely unlikely so maybe it's something that can be overlooked...

___

Regarding why we haven't yet implemented a solution, that's only because we've been trying to decide on an agnostic solution that could be applied generally by anyone using Cloudflare workers/pages, we have some ideas but it's taking quite some time since different teams are involved, truthfully I need to chase that up!
Anyways as I said we could go with a temporary solution anyways now and revert it in a major maybe... let me check with the team and come back to you 👍

Issue #218, opened 2023-05-05
13.4, Server Actions and Turbopack


GregBrimble:
Might want to break this down into multiple things, but this issue tracks the work to check our compatibility with these latest features:

- [ ] Next.js 13.4
- [ ] Server Actions
- [ ] Turbopack

james-elicx:
From what I've been able to tell, the only blocker at the moment is `node:buffer` (#199), which is taken care of by https://github.com/cloudflare/workers-sdk/pull/3133. Using the Wrangler prerelease, I've had success with basic server actions and 13.4 - both seem to be functioning as expected 🙂

Beingbook:
Anyone else who faced `node:async_hook` module resolution problem?

I'm using next.js 13 (app router & server action enabled) with @cloudflare/next-on-pages.

In the build result `.vercel/output/_worker.js`, I can see:
```js
// ../../../../tmp/functions-g41wrtb3ztm.js
var AsyncLocalStoragePromise, __BUILD_OUTPUT__;
var init_functions_g41wrtb3ztm = __esm({
  "../../../../tmp/functions-g41wrtb3ztm.js"() {
    AsyncLocalStoragePromise = import("node:async_hooks").then(({ AsyncLocalStorage: AsyncLocalStorage2 }) => {
      globalThis.AsyncLocalStorage = AsyncLocalStorage2;
    }).catch(() => void 0);
    __BUILD_OUTPUT__ = { "/404.html": {
      type: "override",
      path: "/404.html",
      headers: { "content-type": "text/html; charset=utf-8" }
    }, "/500.html": {
      type: "override",
      path: "/500.html",
      headers: { "content-type": "text/html; charset=utf-8" }
    }, "/_next/static/chunks/800-9e0cf48029271fa9.js": { type: "static" }, "/_next/static/chunks/app/layout-876cdf19b325dd89.js": { type: "static" }, "/_next/static/chunks/app/page-ae43855f96c3b6c6.js": { type: "static" },
// ... skip
"/_next/static/yvP752PVRYqo9GMwF85QY/_buildManifest.js": { type: "static" }, "/_next/static/yvP752PVRYqo9GMwF85QY/_ssgManifest.js": { type: "static" }, "/favicon.ico.body": {
      type: "override",
      path: "/favicon.ico.body",
[...]

PR #234, opened 2023-05-12
add doc for the lazy loading strategy

dario-piotrowicz:


docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
+## Lazy Loading
+
+`@cloudflare/next-on-pages` implements a lazy loading strategy which allows only the necessary javascript code to be evaluated and run when handling a request.
```
GregBrimble, line 3:
```suggestion
`@cloudflare/next-on-pages` implements a lazy loading strategy which allows only the necessary JavaScript code to be evaluated and run when handling a request.
```

docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
...
+
+The CLI builds a Next.js application by running the Vercel build command, afterwards it extrapolates the built code in essentially two different kind of javascript files:
```
GregBrimble, line 5:
```suggestion
The CLI builds a Next.js application by running the `vercel build` command and then it sorts the output into essentially two categories of JavaScript files:
```

docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
...
+
+- **route files**\
+   files each exporting a function that we run to produce a route result for the current request (i.e. based on a request they produce a server side rendered page or the result from an api route)
+- **chunk files**\
+   files containing code that is shared among different route files, this is done as part of our chunks deduplication strategy which allows us to remove unnecessary code duplication present in the Vercel build output (thus reducing the application's overall javascript code size)
```
GregBrimble, line 10:
```suggestion
   Files containing code that is shared among different route files, this is done as part of our chunk deduplication strategy which allows us to remove any unnecessary copies of identical code present in `vercel build`'s output. This reduces the application's overall JavaScript code size.
```

docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
+## Lazy Loading
+
+`@cloudflare/next-on-pages` implements a lazy loading strategy which allows only the necessary javascript code to be evaluated and run when handling a request.
+
+The CLI builds a Next.js application by running the Vercel build command, afterwards it extrapolates the built code in essentially two different kind of javascript files:
+
+- **route files**\
+   files each exporting a function that we run to produce a route result for the current request (i.e. based on a request they produce a server side rendered page or the result from an api route)
+- **chunk files**\
```
GregBrimble, line 9:
```suggestion
- **Route files**\
   Files each exporting a function that we run to produce a route result for the current request (i.e. based on a request they produce a server side rendered page or the result from an api route)
- **Chunk files**\
```

docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
...
+   files containing code that is shared among different route files, this is done as part of our chunks deduplication strategy which allows us to remove unnecessary code duplication present in the Vercel build output (thus reducing the application's overall javascript code size)
+
+The worker we generate for the Pages project then contains custom logic which is used to map a request to one of the route files (or some static asset file but that's out of scope here).
```
GregBrimble, line 12:
```suggestion
The Worker we generate for the Pages project then contains custom logic which is used to map a request to one of the route files (or to a static asset when relevant).
```

docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
...
+
+Such route file statically imports the chunk files it needs, and is in turn dynamically imported by the worker only when the route code is needed to handle the current request, making sure that such code is evaluated and run only when necessary.
```
GregBrimble, line 14:
```suggestion
The Worker only imports the route code when the route code is needed to handle the current request, and the route code statically imports the chunk files it needs. This ensures that an application's code is evaluated and run only when necessary.
```

docs/technical/lazy-loading.md
```diff
@@ -0,0 +1,49 @@
...
+    apiRouteX --> chunk6
+    apiRouteY --> chunk6
+    apiRouteY --> chunk4
+
+    classDef multi opacity:0.8,stroke:transparent,fill:transparent
+    classDef worker fill:transparent
+```
+
+> **Note**
+> The strategy also applies if you disable chunk deduplication (which anyhow is likely useful only for local development and debugging purposes), in such case there won't be any chunk file and the now much bigger route files will still get dynamically imported by the worker.
```
GregBrimble, line 49: I'd like to review the CLI arg you pass to disable chunk deduping. I imagine Turbopack comes out of beta soon-ish, so we'd want to get our CLI ready to be able to switch modes.

GregBrimble (approved):

docs/technical/lazy-loading.md
GregBrimble, line 49:
```suggestion
> The strategy also applies if you disable chunk deduplication (which anyhow is likely useful only for local development and debugging purposes), in such case there won't be any chunk file and the now much bigger route files will still get dynamically imported by the Worker.
```

dario-piotrowicz: regarding the arg, I just made `--experimental-minify` the default (since anyways most apps wouldn't get deployed without it anyways) and added a `--disable-chunks-dedup` flag to disabled it (so it's basically just again a boolean flag to turn the thing on/off, nothing's really changed in that regard)

regarding tubopack I think we do have the issue you created (https://github.com/cloudflare/next-on-pages/issues/218) to track it

I have no idea how painful is going to be and/or how much we'll need to change things 🤷 😓

dario-piotrowicz: This is the PR that introduced the change https://github.com/cloudflare/next-on-pages/pull/208 (but again it's basically just changing the name and default)

GregBrimble: https://github.com/cloudflare/next-on-pages/pull/208/files#r1192279816

PR #236, opened 2023-05-12
add nodejs_compat runtime check

dario-piotrowicz:


src/buildApplication/generateGlobalJs.ts
```diff
@@ -6,26 +6,29 @@
...
+							} catch (e) {
+								return undefined;
+							}
+						},
+						set: (_, property, value) => Reflect.set(envAsyncLocalStorage.getStore(), property, value),
+				}),
+			};
+			return envAsyncLocalStorage;
+		})
+		.catch(() => null);
```
petebacondarwin, line 32: Silently swallow any errors?

petebacondarwin: Is the idea that it won't error unless you actually try to access the ALS promise? So if you are not using `process.env` then you don't need to add `nodejs_compat` flag?

petebacondarwin: Oh no, I see below that you are always accessing the ALS promise in the fetch handler. So why not put the error message in the catch handler here?

dario-piotrowicz: Next.js apps using the app directory do rely on ALS (since v.13.3.0 or something like that) so we do need ALS/nodejs_compat for them in any case. I've also started relying on it in the worker itself (that's the only v1 breaking change, btw) since anyways for most apps will follow in that category (especially moving forward).

Anyways that's was just to clarify why things here are like this 🙂 

____

Regarding why not putting the catch handler here, the code here gets injected at the top of the worker file and is outside the scope of the request handling (and is run at the very beginning when the isolate starts), so I can't returns a response here, I think I can't even create a response object here because it won't be allowed by the workers runtime.

I could put the error message here in the `catch` callback but then below I would have something like:
```ts
declare const __ENV_ALS_PROMISE__: Promise<string | AsyncLocalStorage<unknown>>;

export default {
		const envAsyncLocalStorageOrErrorMessage = await __ENV_ALS_PROMISE__;
		if (typeof envAsyncLocalStorageOrErrorMessage === string) {
			return new Response(
				envAsyncLocalStorageOrErrorMessage,
				{ status: 503 }
			);
		}
		return envAsyncLocalStorageOrErrorMessage.run({ ...env, NODE_ENV: __NODE_ENV__ }, () => {
```
or similar, is this what you had in mind? or am I misunderstanding your question?

I feel like passing `null` here to indicates the absence of the ASL (thus nodejs compatibility) is clearer, don't you think so?

petebacondarwin: Oh I see. Of course, the other code is outside of the fetch handler. OK.

petebacondarwin (approved):

PR #238, opened 2023-05-12
fix: prerendered favicon handling

james-elicx:
This PR does the following:
- Fixes the favicon.ico handling to use its prerendered route.

The build output generates a prerendered function for favicons. We should simply update our prerendered route handling to process the favicon properly, instead of using the `tryToFixFaviconFunc` function.

fixes #237

image
image

PR #242, opened 2023-05-13
fix: prerendered json handling

james-elicx:
Basically #238 but for prerendered JSON for pages in `_next/data/...`.

image
image

### DIFF
--- a/src/buildApplication/fixPrerenderedRoutes.ts
+++ b/src/buildApplication/fixPrerenderedRoutes.ts
@@ -100,7 +100,7 @@ async function getRouteDest(
 		join(
 			dirName,
 			fallback.fsPath.replace(
-				/\.prerender-fallback(\.((rsc)|(body)|(json)))?/gi,
+				/\.prerender-fallback(?:\.(?:rsc|body|json))?/gi,
 				''
 			)
 		)
--- a/src/buildApplication/fixPrerenderedRoutes.ts
+++ b/src/buildApplication/fixPrerenderedRoutes.ts
@@ -83,6 +83,7 @@ async function getRoutePath(
  * // index.prerender-fallback.html -> index.html
  * // index.rsc.prerender-fallback.rsc -> index.rsc
  * // favicon.ico.prerender-fallback.body -> favicon.ico
+ * // 1.json.prerender-fallback.json -> 1.json
  * ```
  *
  * @param config.fallback Fallback file configuration.
@@ -98,7 +99,10 @@ async function getRouteDest(
 	const destRoute = normalizePath(
 		join(
 			dirName,
-			fallback.fsPath.replace(/\.prerender-fallback(?:\.rsc)?(?:\.body)?/gi, '')
+			fallback.fsPath.replace(
+				/\.prerender-fallback(\.((rsc)|(body)|(json)))?/gi,
+				''
+			)
 		)
 	);
 	const destFile = join(outputDir, 'static', destRoute);
--- a/src/buildApplication/fixPrerenderedRoutes.ts
+++ b/src/buildApplication/fixPrerenderedRoutes.ts
@@ -83,7 +83,7 @@ async function getRoutePath(
  * // index.prerender-fallback.html -> index.html
  * // index.rsc.prerender-fallback.rsc -> index.rsc
  * // favicon.ico.prerender-fallback.body -> favicon.ico
- * // 1.json.prerender-fallback.json -> 1.json
+ * // data.json.prerender-fallback.json -> data.json
  * ```
  *
  * @param config.fallback Fallback file configuration.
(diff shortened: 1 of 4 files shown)


src/buildApplication/fixPrerenderedRoutes.ts
```diff
@@ -83,6 +83,7 @@ async function getRoutePath(
  * // index.prerender-fallback.html -> index.html
  * // index.rsc.prerender-fallback.rsc -> index.rsc
  * // favicon.ico.prerender-fallback.body -> favicon.ico
+ * // 1.json.prerender-fallback.json -> 1.json
```
dario-piotrowicz, line 86: why `1`? looks a bit random :sweat_smile: (maybe `X` of the name of some json file we always get?)

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -171,9 +171,19 @@ describe('generateFunctionsMap', async () => {
 			});
 		});
 
-		test('succeeds for prerendered favicon', async () => {
+		test('succeeds for prerendered favicon + json', async () => {
```
dario-piotrowicz, line 174: I think it would be nicer to have two different separated tests, I don't see any real reason/benefit for having these two things tested together (it makes the test just more complex and it can fail if either favicon or json handling is broken without clearly indicating which one actually is)

WDYK?

src/buildApplication/fixPrerenderedRoutes.ts
```diff
@@ -98,7 +99,10 @@ async function getRouteDest(
 	const destRoute = normalizePath(
 		join(
 			dirName,
-			fallback.fsPath.replace(/\.prerender-fallback(?:\.rsc)?(?:\.body)?/gi, '')
+			fallback.fsPath.replace(
+				/\.prerender-fallback(\.((rsc)|(body)|(json)))?/gi,
```
dario-piotrowicz, line 103: This regex feels unnecessarily complex to me, would something like this work instead?
```suggestion
				/\.prerender-fallback\.(?:rsc|body|json)?/gi,
```

dario-piotrowicz (commented):
Awesome! thanks a lot! :heart: 

I've left a couple of small suggestions :slightly_smiling_face:

src/buildApplication/fixPrerenderedRoutes.ts
james-elicx: that was the file i was basing it off of :p

src/buildApplication/fixPrerenderedRoutes.ts
james-elicx: No. For prerendered html files that would strip out the final dot.

tests/src/buildApplication/generateFunctionsMap.test.ts
james-elicx: sure i'll put them separate, idm

src/buildApplication/fixPrerenderedRoutes.ts
dario-piotrowicz: ah ok :sweat_smile::+1:

src/buildApplication/fixPrerenderedRoutes.ts
dario-piotrowicz: ah ok sorry you're right 😓 

than I think this should ok:
```suggestion
				/\.prerender-fallback(?:\.(?:rsc|body|json))?/gi,
```
(you can keep or remove the `?:`s I am a bit fine with both versions)

(or do you think that the extra parens in your version make the regex more clear?)

james-elicx: I think that the extra brackets make it a little more clear that it is that specific group of letters, but honestly, I don't really mind. I'll go with your suggestion.

dario-piotrowicz: ah ok thanks a lot, I prefer not to have them as it makes me wonder if they are there for a specific purpose

(but yeah I'd be ok with having them if you strongly preferred them :) )

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -171,7 +171,7 @@ describe('generateFunctionsMap', async () => {
...
```
dario-piotrowicz, line 174:
```suggestion
		test('succeeds for prerendered favicon', async () => {
```

james-elicx: derp i didn't change it when i copy pasted. oopsy

dario-piotrowicz (approved):
Thanks!!! ❤️

tests/src/buildApplication/generateFunctionsMap.test.ts
dario-piotrowicz: it happens :slightly_smiling_face::+1:

PR #253, opened 2023-05-16
Fix template copy

GregBrimble:


dario-piotrowicz (approved):
Looks good to me thanks! 😃 

Although I would add a changeset here (even though this is a very small change)

PR #259, opened 2023-05-18
add wrangler3 peer dependency

dario-piotrowicz:


petebacondarwin (commented):
Would this version of next-on-pages still work with 2.20?
If so, do we need to force bump people to 3.0.0?

dario-piotrowicz:
> Would this version of next-on-pages still work with 2.20? If so, do we need to force bump people to 3.0.0?

No it doesn't work with 2.20, we do need to force 3.0.0 😓

petebacondarwin (approved):

Issue #257, opened 2023-05-18
[🐛 Bug]: KV namespace bindings


willin:
### next-on-pages environment related information

```
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/--version: Not Found - 404

This error happened while installing a direct dependency of /Users/v0/Library/pnpm/store/v3/tmp/dlx-33292

--version is not in the npm registry, or you have no permission to fetch it.

An authorization header was used: Bearer npm_[hidden]

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:28 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6000
        CPU: (10) arm64 Apple M1 Max
        Memory: 32 GB
        Shell: /bin/zsh
Binaries:
        Node: 19.1.0
        Yarn: 3.3.0
        npm: 8.19.3
        pnpm: N/A
Package Manager Used: npm
Relevant Packages:
        @cloudflare/next-on-pages: 1.0.0
        vercel: N/A
        next: N/A
```

### Description

```ts
import { KVNamespace } from '@cloudflare/workers-types';
import { NextResponse } from 'next/server';

export async function GET(request: Request) {
  const { DOMAINS } = process.env as any as { DOMAINS: KVNamespace };
  await DOMAINS?.put('test', 'value');

  const data = await DOMAINS?.get('test');
  return NextResponse.json({ data });
}
```

local output: `{}`

pages `Internal Server Error`

```js
{
  "outcome": "ok",
  "scriptName": "pages-worker--1088967-preview",
  "exceptions": [],
  "logs": [
    {
      "message": [
        "Error: internal error"
      ],
      "level": "error",
      "timestamp": 1684379041372
    }
  ],
  "eventTimestamp": 1684379041372,
  "event": {
    "request": {
      "url": "https://domain.js.cool/api/hello/",
      "method": "GET",
      "headers": {
        "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8",
        "accept-encoding": "gzip",
        "accept-language": "zh-CN,zh;q=0.9",
        "cache-control": "max-age=0",
        "cf-connecting-ip": "240e:3ae:937:8ff0:159a:ef6b:7a4a:ae64",
        "cf-connecting-o2o": "1",
        "cf-ipcountry": "CN",
[...]

PR #260, opened 2023-05-18
fix: non-dynamic trailing slash matching when using `trailingSlash`

james-elicx:
This PR does the following:
- In the `rewrite` phase, if no path is found in the build output, check if there is a trailing slash. If so, remove it and check the build output again.

When using `trailingSlash: true` in `next.config.js`, it redirects requests to include a trailing slash. However, it doesn't always rewrite the paths in the build output config to let them match. For dynamic routes, there are rewrites present in the `rewrite` phase that account for the trailing slash. However, with non-dynamic routes, there typically isn't a rewrite to take care of them. So, if we weren't able to match a rewrite and there's a trailing slash, we should manually remove it and check if the file system contains the route.

Tested using the reproduction in https://github.com/cloudflare/next-on-pages/issues/257
That issue is not about trailing slashes, and therefore I am not marking it as being fixed by this PR, but it's a problem that I noticed when trying out their reproduction.

### DIFF
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -504,9 +504,9 @@ export class RoutesMatcher {
 		if (
 			phase === 'rewrite' &&
 			!pathExistsInOutput &&
-			/^\/(.+)\/$/.test(this.path)
+			this.path.endsWith('/')
 		) {
-			const newPath = this.path.replace(/^\/(.+)\/$/, '/$1');
+			const newPath = this.path.replace(/\/$/, '')
 			pathExistsInOutput = newPath in this.output;
 			if (pathExistsInOutput) {
 				this.path = newPath;
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -506,8 +506,11 @@ export class RoutesMatcher {
 			!pathExistsInOutput &&
 			/^\/(.+)\/$/.test(this.path)
 		) {
-			this.path = this.path.replace(/\/(.+)\//, '/$1');
-			pathExistsInOutput = this.path in this.output;
+			const newPath = this.path.replace(/^\/(.+)\/$/, '/$1');
+			pathExistsInOutput = newPath in this.output;
+			if (pathExistsInOutput) {
+				this.path = newPath;
+			}
 		}
 
 		// In the `miss` phase, set status to 404 if no path was found and it isn't an error code.
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -496,7 +496,19 @@ export class RoutesMatcher {
 			return 'done';
 		}
 
-		const pathExistsInOutput = this.path in this.output;
+		let pathExistsInOutput = this.path in this.output;
+
+		// If a path with a trailing slash entered the `rewrite` phase and didn't find a match, it might
+		// be due to the `trailingSlash` setting in `next.config.js`. Therefore, we should remove the
+		// trailing slash and check again before entering the next phase.
+		if (
+			phase === 'rewrite' &&
+			!pathExistsInOutput &&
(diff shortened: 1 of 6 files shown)


tests/templates/requestTestData/trailingSlash.ts
```diff
@@ -0,0 +1,146 @@
...
+			src: '/404/?',
+			status: 404,
+			continue: true,
+			missing: [{ type: 'header', key: 'x-prerender-revalidate' }],
+		},
+		{ src: '/500', status: 500, continue: true },
+		{ handle: 'resource' },
+		{ src: '/.*', status: 404 },
+		{ handle: 'rewrite' },
+		{ src: '^/(?<nxtPlang>[^/]+?)(?:/)?$', dest: '/[lang]?nxtPlang=$nxtPlang' },
```
dario-piotrowicz, line 33: does `nxtPlang` has any meaning? could you maybe use something like generic or just `x` instead?

dario-piotrowicz (approved):
Awesome! thanks! 😃 

Just left a couple of minor comments please have a look 🙏

tests/templates/requestTestData/trailingSlash.ts
james-elicx: That was just what was in the route config i copied it from. can change it to be more descriptive sure

templates/_worker.js/routes-matcher.ts
```diff
@@ -496,7 +496,19 @@ export class RoutesMatcher {
...
+
+		// If a path with a trailing slash entered the `rewrite` phase and didn't find a match, it might
+		// be due to the `trailingSlash` setting in `next.config.js`. Therefore, we should remove the
+		// trailing slash and check again before entering the next phase.
+		if (
+			phase === 'rewrite' &&
+			!pathExistsInOutput &&
+			/^\/(.+)\/$/.test(this.path)
+		) {
+			this.path = this.path.replace(/\/(.+)\//, '/$1');
```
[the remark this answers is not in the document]
james-elicx: Oops!!! I added it the first but completely forgot the second!

Hmm, yeah, that's a decent thought - I'm happy to go down that route

tests/templates/requestTestData/trailingSlash.ts
dario-piotrowicz: yeah no I just meant that that feels a bit random, anything like that makes it clear that it's a dummy/meaningless string would be better I think 🙂 

(I've seen a couple of this in your other tests, I guess this just stuck a bit with me because I don't what it could indicate 🤣 )

templates/_worker.js/routes-matcher.ts
```diff
@@ -496,7 +496,22 @@ export class RoutesMatcher {
...
+			phase === 'rewrite' &&
+			!pathExistsInOutput &&
+			/^\/(.+)\/$/.test(this.path)
+		) {
+			const newPath = this.path.replace(/^\/(.+)\/$/, '/$1');
+			pathExistsInOutput = newPath in this.output;
+			if (pathExistsInOutput) {
+				this.path = newPath;
+			}
+		}
```
dario-piotrowicz, line 514: sorry 😓 , last minute suggestion

Thinking about it I think that this code can be simplified a little?

```suggestion
		if (
			phase === 'rewrite' &&
			!pathExistsInOutput &&
			this.path.endsWith('/')
		) {
			const newPath = this.path.slice(0, -1);
			pathExistsInOutput = newPath in this.output;
			if (pathExistsInOutput) {
				this.path = newPath;
			}
		}
```
or:

```suggestion
		if (
			phase === 'rewrite' &&
			!pathExistsInOutput &&
			this.path.endsWith('/')
		) {
			const newPath = this.path.replace(/\/$/, '')
			pathExistsInOutput = newPath in this.output;
			if (pathExistsInOutput) {
				this.path = newPath;
			}
		}
```

?

dario-piotrowicz (approved):

PR #265, opened 2023-05-19
Add Utils sub-module and allow users to access cf and ctx

dario-piotrowicz:
To see how the utils are used see: https://github.com/dario-piotrowicz/next-apps-for-testing/commit/cb27ef60cf445331aea2b502a4c2ad088272bb67

And you can see the deployed application here: https://ac37c053.next-on-pages-test-5h3.pages.dev/

___

>**Note**
> We do need utils for bindings, etc... but I wanted to keep this PR simple and just do the
> basic setus with the cf and ctx utilities
>
> We'll also need to document all of these, again I think I'd leave the documentation for a different PR


src/cli/templates/_worker.js/index.ts
```diff
@@ -0,0 +1,58 @@
...
+				`Error: Could not access built-in Node.js modules. Please make sure that your Cloudflare Pages project has the 'nodejs_compat' compatibility flag set.`,
+				{ status: 503 }
+			);
+		}
+
+		const cloudflareGlobalContextAls = getCloudflareGlobalContextAls();
+
+		if (!cloudflareGlobalContextAls) {
+			return new Response(
+				`Internal Server Error: cannot retrieve the cloudflare global context AsyncLocalStorage`,
```
dario-piotrowicz, line 28: this should actually never occur, but I guess I think this makes the logic here a bit more robust

if preferred I don't mind using a non-null assertion on `cloudflareGlobalContextAls` or similar

src/cli/templates/_worker.js/index.ts
```diff
@@ -0,0 +1,58 @@
...
+			);
+		}
+
+		return cloudflareGlobalContextAls.run(
+			{
+				cf: request.cf,
+				ctx,
+			},
+			() => {
+				return envAsyncLocalStorage.run(
```
dario-piotrowicz, line 39: instead of this nesting `run` we could use `enterWith` but it seems like the latter is experimental: https://nodejs.org/api/async_context.html#asynclocalstorageenterwithstore (and I am not sure if we even support that) so I personally would not go with that for now (also the nesting is not too too bad in my opinion 🙂)

anyways I can experiment with `enterWith` if someone feels strongly about it (but possibly I'd do in a separate PR and keep this one just focused on the utils logic)

cc. @GregBrimble

src/cli/templates/_worker.js/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		const cloudflareGlobalContextSymbol = Symbol.for(
+			'cloudflare-global-context'
+		);
+
+		const cloudflareGlobalContextAls = (
+			globalThis as unknown as {
+				[cloudflareGlobalContextSymbol]?: AsyncLocalStorage<CloudflareGlobalContext>;
+			}
+		)[cloudflareGlobalContextSymbol];
```
dario-piotrowicz, line 31: unfortunately here we duplicate some code from `src/utils/utils/cloudflareGlobalContext.ts` but I couldn't share it because we don't include `src/utils` in the package (while we do include the `templates` code, which is used by the CLI on build)

I could do something like creating a new shared dir with this code and include it in the package, or move this code to the templates dir and import it from the utils... none of these options sounds very good to me, so I just went with having some code duplication

If anyone has a better idea (or would prefer some of the solutions above) please let me know 🙏

package.json
```diff
@@ -2,14 +2,30 @@
...
 		"prettier:check": "npm run prettier -- --check",
 		"prettier:fix": "npm run prettier -- --write",
 		"lint": "eslint src",
 		"types-check": "tsc --noEmit",
-		"build": "npx esbuild --bundle --platform=node ./src/index.ts --external:esbuild --external:chokidar --outfile=./dist/index.js",
-		"build:watch": "npm run build -- --watch",
+		"build-cli": "npx esbuild --bundle --platform=node ./src/cli/index.ts --external:esbuild --external:chokidar --outfile=./dist/index.js",
+		"build-utils": "npx esbuild --bundle --platform=node --format=esm ./src/utils/index.ts --external:esbuild --external:chokidar --outfile=./dist/utils/index.js",
+		"build-utils-types": "tsc  -p ./tsconfig.utils.json --emitDeclarationOnly --declaration --outDir ./dist/utils",
+		"build": "npm run build-cli ; npm run build-utils ; npm run build-utils-types",
```
GregBrimble, line 24: Just a suggestion :) https://www.npmjs.com/package/concurrently

package.json
```diff
@@ -30,15 +46,15 @@
 		"js-yaml": "^4.1.0",
 		"pcre-to-regexp": "^1.1.0",
 		"zod": "^3.21.4",
-		"zodcli": "^0.0.4"
+		"zodcli": "^0.0.4",
+		"@cloudflare/workers-types": "^4.20230404.0"
```
GregBrimble, line 50: Is it normal to move to proper deps when exporting types? I'm not familiar with the common practices here.

src/cli/templates/_worker.js/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		if (!cloudflareGlobalContextAls) {
+			return new Response(
+				`Internal Server Error: cannot retrieve the cloudflare global context AsyncLocalStorage`,
```
GregBrimble, line 35:
```suggestion
				`Internal Server Error: cannot retrieve the Cloudflare global context AsyncLocalStorage`,
```

src/cli/templates/_worker.js/index.ts
```diff
@@ -0,0 +1,65 @@
...
+				{ status: 500 }
```
GregBrimble, line 36:
```suggestion
				{ status: 503 }
```

So it's like the `nodejs_compat` error page?

tsconfig.utils.json
GregBrimble, line 1: I personally like to put these into their respective folders so you can keep them being named `tsconfig.json`. I think some tooling doesn't yet support arbitrarily named tsconfig files.

tsconfig.utils.json
```diff
@@ -0,0 +1,7 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
```
GregBrimble, line 3: https://www.typescriptlang.org/tsconfig#declarationMap is a nice option to enable for users.

package.json
dario-piotrowicz: yeah, I didn't want to introduce an extra dependency but this could be made a bit nicer

Since you're suggesting it I'll try something :slightly_smiling_face: 

I was actually thinking of using [npm-run-all](https://www.npmjs.com/package/npm-run-all) and do:
```json
  "build": "run-p build-*",
```

package.json
dario-piotrowicz: yeah I am not too familiar with the common practices either, but I thought that the consumer of next-on-pages will need access to the types, so this package is needed as a standard dependency so that they will get them

but I don't think I've tried to see if it works with the package set as a dev dependency, I'll check that and will let you know, thanks :smiley:

tsconfig.utils.json
dario-piotrowicz: ah ok, so you mean something like `src/utils/tsconfig.json`? while we keep the `tsconfig.json` at the root as is? (since it's also used for testing and whatnot)

tsconfig.utils.json
dario-piotrowicz: ah yeah! very nice! thanks! :smiley:

package.json
dario-piotrowicz: Ok, I've tried putting this back (see https://github.com/cloudflare/next-on-pages/pull/265/commits/0e7f67a291d9c8411ff34d6760827d002adb1355, [prerelease](https://prerelease-registry.devprod.cloudflare.dev/next-on-pages/runs/5031931405/npm-package-next-on-pages-265)) but as I expected it then stops working:

Screenshot 2023-05-20 at 12 34 04

Screenshot 2023-05-20 at 12 33 50

dario-piotrowicz: vs having is as a dependency (see https://github.com/cloudflare/next-on-pages/pull/265/commits/dd3e1ce138542fb674bd8e371b2f818118b3d1b2, [prerelease](https://prerelease-registry.devprod.cloudflare.dev/next-on-pages/runs/5031974631/npm-package-next-on-pages-265)):
Screenshot 2023-05-20 at 12 48 37

Screenshot 2023-05-20 at 12 48 55

package.json
dario-piotrowicz: I've added `npm-run-all` and made the `build` and `build:watch` commands simpler, please have a look 🙂

(note: I used a somewhat weird naming convention, the fact is that in the `run-p` pattern I needed to have distinct prefixes for the build and build+watch scripts)

src/cli/templates/_worker.js/index.ts
dario-piotrowicz: > So it's like the nodejs_compat error page?

yeah practically it is, but as I said this should never happen as I think that the `cloudflareGlobalContextAls` is always going to be defined here at this point (since it is after the `envAsyncLocalStorage` check so `AsyncLocalStorage` is indeed available)

but I figured I would add this check just to be extra sure that if something went very wrong then there is a clear error message indicating the issue (but I think/hope no one will see this error message in their apps)

so, I am totally fine removing the check if you feel is too unnecessary (and just assume that `cloudflareGlobalContextAls` is always defined)

tsconfig.utils.json
dario-piotrowicz: works great!!! thanks! 😍

tsconfig.utils.json
dario-piotrowicz: I did that, please have a look and let me know if that's what you had in mind 🙂

(I could add a separate `src/cli/tsconfig.json` file that imports from the root one or something, and then maybe I could omit a couple of lines from the root `tsconfig`... but I am not too sure if that's needed, maybe later if we see that the need arises? WDYT?)

package.json
GregBrimble: npm-run-all is no longer maintained (whereas concurrently has basically all the same functionality and is maintained :))

package.json
GregBrimble: Cool 👍

.changeset/khaki-books-rescue.md
```diff
@@ -0,0 +1,33 @@
...
+	getRequestExecutionContext,
+} from '@cloudflare/next-on-pages/utils';
+
+// ...
+
+const cf = getRequestCfProperties();
+// access cf.colo, cf.city, etc...
+
+const ctx = getRequestExecutionContext();
+// ctx.waitUntil(myPromise), etc...
```
GregBrimble, line 32: We should show this in a useful context (e.g. being used in an App Router page render).

src/utils/getRequestCfProperties.ts
GregBrimble, line 1: Can this actually be `undefined`? When does that happen?

src/utils/getRequestExecutionContext.ts
GregBrimble, line 1: Likewise. Ever `undefined` here?

GregBrimble (approved):

package.json
dario-piotrowicz: ok, concurrently it is then 🙂👍 

By the way, I wanted `npm-run-all` because you could use `*` I didn't think that concurrently supported that but it looks like it does, so yeah I'm totally ok with that 🙂 (actually their matching seems even more advanced 😄)

.changeset/khaki-books-rescue.md
dario-piotrowicz: This is just the changeset, I will look into docs later if that's ok 🙂 

(or do you feel strongly that the changeset example should be good?)

src/utils/utils/cloudflareGlobalContext.ts
```diff
@@ -0,0 +1,14 @@
...
+	'cloudflare-global-context-als'
+);
+
+export function getCloudflareGlobalContextAls():
+	| AsyncLocalStorage<CloudflareGlobalContext>
+	| undefined {
+	const global = globalThis as unknown as {
+		[cloudflareGlobalContextAlsSymbol]?: AsyncLocalStorage<CloudflareGlobalContext>;
+	};
+	return global?.[cloudflareGlobalContextAlsSymbol];
```
dario-piotrowicz, line 13:
```suggestion
	return global[cloudflareGlobalContextAlsSymbol];
```

dario-piotrowicz:
@GregBrimble I am not sure what you're referring to in your last two comments, it seems that they somehow don't point to any code 😓 🤷 

if this is the sort of thing you're referring to:
Screenshot 2023-05-20 at 21 32 16

no, the store cannot be undefined at this point, I just used optional chaining to make typescript happy 😓 (even though we incur a small runtime cost)

the alternative (which is definitely more correct) would be this:
Screenshot 2023-05-20 at 21 33 58

let's go with the non-null assertion since it's more correct and removes ambiguity 🙂 (although it is quite annoying that it forces us to add the eslint comment, I do have mixed feelings about the eslint non-null assertion rule 😅)

dario-piotrowicz:
@GregBrimble I ended up removing the `?.` but I actually changed `getCloudflareGlobalContextAls` to just `getCloudflareGlobalContext` (since the caller would anyways always try to get the store so we might as well just return that directly) and as a by product we already implicitly check the presence of the store, I hope this is fine please let me know 🙂

package.json
dario-piotrowicz: switched to `concurrently` it works great 😃

dario-piotrowicz:
I was thinking of what other functions we could have for bindings, Kvs, etc...

And it actually occurred to me that instead of having `getRequestCfProperties`, `getRequestExecutionContext`, `getKeyValueNamespaces`, etc...

It could be a much nicer API if we provided a single `getCloudflareContext` function which would return an object which would contain the `cf` and `ctx` alongside `kvs`, `D1s`, bindings, etc...

I think it would be nicer since developers would only need to be aware of/remember a single function (instead of a whole set of functions) and the return value of that function would anyway be fully typed so it should be rather straightforward to get the various values from that.

I think I'll experiment into creating such utility function, but if ok I would anyways merge this PR first (since it does set up things nicely) and look into updating the API in a separate followup PR (and anyway release only when a final API has been decided)

dario-piotrowicz:
I've converted the PR to draft as we first need to agree on the API we want to offer (and there are currently conflicting opinions on the team which we need to sort out)

.changeset/khaki-books-rescue.md
```diff
@@ -0,0 +1,33 @@
...
+
+Note that naturally both only work on the server, they throw an exception if run on the client,
+where there is not request to get the objects from.
+
+Usage example:
+
+```ts
+import {
+	getRequestCfProperties,
+	getRequestExecutionContext,
```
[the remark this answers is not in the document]
dario-piotrowicz: the `Properties` part of `getRequestCfProperties` was just to match the workers types type: `IncomingRequestCfProperties`, since it's included there I felt as it could make things clearer to have it here as well?

regarding `getRequestExecutionContext` I can't really remember for sure but I _think_ that I included `Request` so that that the two methods have a very similar format making their use a bit easier/streamlines 🤔 

I don't mind changing the names, but what do you think of the above?

dario-piotrowicz:
I'm closing this PR as I think we're not going to go with this implementation

shiqimei:
> I'm closing this PR as I think we're not going to go forward with this implementation

Hi @dario-piotrowicz . Do you have other plan to retrive `cf` in next.js edge function?

dario-piotrowicz:
@shiqimei I'm replying to [the other PR](https://github.com/cloudflare/next-on-pages/pull/233) you commented on 🙂👍

PR #273, opened 2023-05-23
refine /_next/static/* headers generation

### LINKED ISSUE #97: The headers generation needs to be improved
This is how we generate headers at build time: https://github.com/cloudflare/next-on-pages/blob/4a360000fbd80c682634a81faf13e564da71b7d8/src/index.ts#L495-L508 (introduced in #50)

These values are hardcoded and should instead use what provided by the vercel cli in the `.vercel/output/config.json` file (as mentioned in this [comment](https://github.com/cloudflare/next-on-pages/pull/50#discussion_r1134719235))

We already have [`routesMatcher`](https://github.com/cloudflare/next-on-pages/blob/4a360000fbd80c682634a81faf13e564da71b7d8/templates/_worker.js/index.ts#L42) which can parse this `config.json`, so we should pull out the matching logic into a utility function that can be used at both runtime and at build time.

<details>

<summary>routesMatcher</summary>

`routesMatcher` contains logic on how to parse config.json and extract routes info from it. We need to have it available both at runtime and build time. In its current state however it's not really suitable for use at build time (since it relies on the incoming request), so it needs to be refactored and used:
 - in the same way it currently is in the runtime (in [`templates/_worker.js/index.ts`](https://github.com/cloudflare/next-on-pages/blob/main/templates/_worker.js/index.ts)) 
 - to add headers for all the remaining routes besides `/_next/static/*` at build time (in [`generateHeaders`](https://github.com/cloudflare/next-on-pages/blob/4a360000fbd80c682634a81faf13e564da71b7d8/src/index.ts#L495))

</details>

dario-piotrowicz:
resolves #97

___

Before:
Screenshot 2023-05-23 at 18 31 27

After:
Screenshot 2023-05-23 at 18 32 59

(basically the only noticeable different is that we get the header lower cased)


dario-piotrowicz:
In the issue we specified that we'd want to collect the headers off all routes present in the Vercel config, but I don't think we should do that since the headers are Vercel specific and used by the routing mechanism anyways.

Actually I am pretty sure we serve all static files via the worker anyways and apply the correct headers, so I think that instead of my change we can just remove this `_headers` logic altogether...

(I'll check that now, @james-elicx might know on the top of his head 🙂)

james-elicx:
> In the issue we specified that we'd want to collect the headers off all routes present in the Vercel config, but I don't think we should do that since the headers are Vercel specific and used by the routing mechanism anyways.
> 
> Actually I am pretty sure we serve all static files via the worker anyways and apply the correct headers, so I think that instead of my change we can just remove this `_headers` logic altogether...
> 
> (I'll check that now, @james-elicx might know on the top of his head 🙂)

The `/_next/static` directory doesn't invoke the worker, so that doesn't get any headers from it, but for everything else, the relevant headers are applied during routing.

I would be opposed to collecting all headers and applying them in this config as during routing, it's possible to *override* the collected headers and discard them, etc. Only using this config with directories excluded from invoking the worker seems the most reasonable choice.

dario-piotrowicz:
yeah thanks @james-elicx sorry the `/_next/static` directory doesn't invoke the worker so this change is actually valid, never mind me 😅

src/buildApplication/buildMetadataFiles.ts
```diff
@@ -1,27 +1,41 @@
...
-		`
+async function buildNextStaticHeaders() {
+	const vercelConfig = await getVercelConfig();
+
+	const hitRoutes = getPhaseRoutes(vercelConfig, 'hit');
+
+	const nextStaticRoute = hitRoutes.find(route =>
+		route.src?.startsWith('/_next/static/')
+	);
+	const nextStaticHeaders = (nextStaticRoute as VercelSource).headers;
```
petebacondarwin, line 21: `nextStaticRoute` could be null here, right?

Or is it guaranteed to find one?

petebacondarwin: Perhaps this should be 

```suggestion
	const nextStaticHeaders = (nextStaticRoute as VercelSource)?.headers;
```

src/buildApplication/getVercelConfig.ts
```diff
@@ -28,6 +28,57 @@ export async function getVercelConfig(): Promise<VercelConfig> {
...
+	const routes: VercelRoute[] = [];
+
+	if (!config.routes) {
+		return [];
+	}
+
+	let currentRouteIdx = 0;
+
+	if (phase === 'none') {
+		while (
```
petebacondarwin, line 51: It feels like these while loops (checking for undefined items in the routes array) would be better coded as for-loops stopping on the index being longer than the array. Or is it possible to actually have undefined values in the array?

src/buildApplication/getVercelConfig.ts
```diff
@@ -28,6 +28,57 @@ export async function getVercelConfig(): Promise<VercelConfig> {
...
+			config.routes[currentRouteIdx] &&
+			!(config.routes[currentRouteIdx] as VercelHandler).handle
+		) {
+			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
petebacondarwin, line 55: Assigning the item in the array to a const variable and checking the truthiness of that variable would avoid these non-null assertions.

src/buildApplication/buildMetadataFiles.ts
dario-piotrowicz: I think it is always guaranteed to find one 🤔

but I don't mind having more defensive code here to make sure things are as they should be 🙂👍 

(it's been a while since I opened this PR I am not sure if I missed the potential `undefined` because of the casting or if I purposely omitted it since we should always find the headers 🤔)

src/buildApplication/getVercelConfig.ts
dario-piotrowicz: To remove the eslint comment I thought of two alternatives:
Screenshot 2023-06-06 at 12 58 07

any preference? (or a different idea?)

(PS: unfortunately both to me feel less clear than the current code 😓)

src/buildApplication/getVercelConfig.ts
petebacondarwin: Here is a suggestion of how this could look (and be more self documenting)...

```ts
export function getPhaseRoutes(
	config: VercelConfig,
	phase: VercelPhase
): VercelRoute[] {
	if (config.routes === undefined) {
		return [];
	}
	const phaseStart = findStartOfPhase(config.routes, phase);
	return getRoutesInPhase(config.routes, phaseStart);
}

function findStartOfPhase(routes: VercelRoute[], phase: string) {
	if (phase === 'none') {
		return 0;
	} else {
		const index = routes.findIndex(
			route => isVercelHandler(route) && route.handle === phase
		);
		return index === -1 ? routes.length : index + 1;
	}
}

function getRoutesInPhase(routes: VercelRoute[], phaseStart: number) {
	const collectedRoutes: VercelRoute[] = [];
	for (let i = phaseStart; i < routes.length; i++) {
		const route = routes[i];
		if (!route || isVercelHandler(route)) {
			break;
		}
		collectedRoutes.push(route);
	}
	return collectedRoutes;
}

function isVercelHandler(route: VercelRoute): route is VercelHandler {
	return 'handler' in route;
}
```

dario-piotrowicz: Thanks! 😄 

although I am really not sure about the two functions `getPhaseRoutes` and `getRoutesInPhase` they basically convey the same exact intent don't they?

what I mean is, if someone unfamiliar with this code would read the two functions names it would be very unclear what each does and how they are different

I've applied your suggestion but merged the two functions into one, I hope that's fine, please have a look 🙏

dario-piotrowicz: I've also tried improving things further and avoid iterative loops entirely (just using `findIndex`es + `slice` at the end)

please let me know what you think 😄

src/buildApplication/getVercelConfig.ts
```diff
@@ -28,6 +28,44 @@ export async function getVercelConfig(): Promise<VercelConfig> {
...
+	const index = routes.findIndex(
+		route => isVercelHandler(route) && route.handle === phase
+	);
+	return index === -1 ? routes.length : index + 1;
+}
+
+function getPhaseEnd(routes: VercelRoute[], phaseStart: number): number {
+	const index = routes
+		.slice(phaseStart)
+		.findIndex(route => isVercelHandler(route));
```
petebacondarwin, line 65: The slice is a bit of heavy way to skip the first few items. How about:

```suggestion
		.findIndex((route, i) => i >= phaseStart && isVercelHandler(route));
```

src/buildApplication/getVercelConfig.ts
petebacondarwin: Great! Made one more suggestion.

petebacondarwin (approved):

src/buildApplication/getVercelConfig.ts
dario-piotrowicz: sure 🙂

PR #275, opened 2023-05-24
Add option to ignore routes

dario-piotrowicz:


dario-piotrowicz:
This was simply an idea I had, it's probably not very valuable, not to have the PR lingering forever I am closing it now

We might come back to this if/when there's a need for such an option

PR #276, opened 2023-05-24
feat: custom output directory through the `--outdir` flag

### LINKED ISSUE #231: [⚡ Feature]: --outdir
### Description

Let me specify an output directory rather than always only ever exporting to `.vercel/output/static/_worker.js`.

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Implements support for specifying a custom output directory with `--outdir`.
- Deletes/cleans the directory if it already exists, and copies all static assets to the output directory and uses it as the target for the worker.
- Leaves the `.vercel/output/static` directory unaltered when using a custom output directory.
- Moves the CLI error handling to its own function so it's a bit easier to work with, and supports other error types.
- Implements a utility for tests to make mocking the console easier+simpler.

closes #231


tests/_helpers/index.ts
```diff
@@ -300,3 +302,29 @@ export function createPrerenderedRoute(
 		[`${file}.rsc.prerender-fallback.rsc`]: `${fileWithBase}.rsc.prerender-fallback.rsc`,
 	};
 }
+
+type ConsoleMethods = 'log' | 'warn' | 'error';
+export function mockConsole<TMethods extends ConsoleMethods>(
+	methods: TMethods[]
```
dario-piotrowicz, line 308: Not sure about the `TMethods[]` that means that the following is valid:
```ts
const mockedConsole = mockConsole(['warn', 'warn', 'warn']);
```
as well as:
```ts
const mockedConsole = mockConsole([]);
```

although they doesn't make sense

we could use `Set<TMethods>` but that feels like it'd makes callers more verbose... (and you could still have an empty set)

You could also define a complex type here to make sure that you always get between one and three methods... but it doesn't feel worth it...

What about `mockConsole(method: ConsoleMethod)` and when you need multiple methods you call the function multiple times?

dario-piotrowicz: also sorry but I'm really not convinced on the benefits here 😢, using this function requires you to specify the method on creation and then also in the expected calls, feels to me that using it for a single console method is as much work (if not more) as doing things manually (and in almost all cases we do specify a single method)

(the only real benefit I can see is the restoring of all the methods, but again that only makes sense if you have more then one method, which we usually don't)

james-elicx: i just wanted a more convenient way to mock the console than doing the full thing each time tbh. i can revert it if you want, or change it to `mockConsole(method: ConsoleMethod)`. i'd probably lean towards the latter rather than just undoing it, because IMO it would still be more convenient, but i dont mind.

dario-piotrowicz: mh... ok if you feel strongly about the utility I am fine with it 🙂 

as long as after the `mockConsole(method: ConsoleMethod)` you shouldn't have to specify that method again, as in:
```ts
const mockedWarn = mockConsole('warn');
// ...
mockedWarn.expectCalls([
	/.../,
]);
```

would that be ok? 🙂

james-elicx: yes that sounds good to me 👍🏻

src/cli.ts
```diff
@@ -31,11 +32,69 @@ const cliOptions = z
...
+			const label = `Unknown option${unknownKeys.length === 1 ? '' : 's'}`;
+
+			return {
+				msg: `${label}: ${unknownKeys.join(', ')}`,
+				showHelp: true,
+			};
+		}
+
+		if (
+			(issue?.code === 'custom' || issue?.code === 'invalid_type') &&
```
dario-piotrowicz, line 79:
```suggestion
			(issue.code === 'custom' || issue.code === 'invalid_type') &&
```

src/cli.ts
```diff
@@ -31,11 +32,69 @@ const cliOptions = z
...
 		info: flag,
+		outdir: z
+			.string()
+			.optional()
+			.default(join('.vercel', 'output', 'static'))
+			.transform(path => normalizePath(resolve(path)))
+			.refine(
+				path =>
+					path.startsWith(normalizePath(resolve())) &&
+					path !== normalizePath(resolve()),
```
dario-piotrowicz, line 43:
```suggestion
				path => {
					const currentWorkingDirectory = normalizePath(resolve());
					return (
						path.startsWith(currentWorkingDirectory) &&
						path !== currentWorkingDirectory
					);
				},
```

Also the check could be made using a regex, but I suspect you didn't do that for readability, which I'm ok with 🙂

tests/src/buildApplication/processVercelOutput.test.ts
```diff
@@ -407,3 +416,72 @@ describe('processVercelOutput', () => {
...
+			existing: { 'file.txt': 'hello world' },
+		});
+	});
+
+	afterEach(() => mockFs.restore());
+
+	const vercelDir = resolve('.vercel', 'output', 'static');
+
+	test('default vercel output dir deletes telemetry file only', async () => {
+		const staticAssets = await getVercelStaticAssets();
```
dario-piotrowicz, line 441: ideally I don't think that in these tests we should use some production code (basically here you are kind of testing both `getVercelStaticAssets` and `setupOutputDir`)

it's not a huge deal as I think we have much worse coupling in other tests but here it should be pretty simple to remove this

src/cli.ts
dario-piotrowicz: also is it a necessary limitation that the output dir needs to be inside the current working dir? couldn't the output dir be anywhere?

src/buildApplication/processVercelOutput.ts
```diff
@@ -27,6 +29,67 @@ export async function getVercelStaticAssets(): Promise<string[]> {
...
+ */
+export async function setupOutputDir(
+	outputDir: string,
+	staticAssets: string[]
+) {
+	const vercelDir = normalizePath(resolve('.vercel', 'output', 'static'));
+
+	// If the output directory is not the default Vercel one, delete it if exists and create a new one.
+	// Then, copy the static assets from the default Vercel output directory to the new one.
+	if (!outputDir.startsWith(vercelDir)) {
```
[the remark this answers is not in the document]
james-elicx: ngl, i dont remember - will change it 😅

src/buildApplication/processVercelOutput.ts
```diff
@@ -27,6 +29,67 @@ export async function getVercelStaticAssets(): Promise<string[]> {
...
+		cliLog(`Using custom output directory: ${relative('.', outputDir)}`);
+
+		await rm(outputDir, { recursive: true, force: true });
+		await mkdir(outputDir, { recursive: true });
+		await copyVercelStaticAssets(outputDir, vercelDir, staticAssets);
+
+		await deleteNextTelemetryFiles(outputDir);
```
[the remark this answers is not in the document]
james-elicx: The approach I'm taking is to leave Vercel's output completely unaltered when you use a custom output directory, hence why I changed the deletion to be after all static assets are copied instead of before.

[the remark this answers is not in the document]
james-elicx: That's a fair point, sure

tests/src/buildApplication/generateFunctionsMap.test.ts
```diff
@@ -470,7 +541,9 @@ const invalidFuncDir = {
 async function generateFunctionsMapFrom(
 	functions: DirectoryItems,
 	staticAssets: DirectoryItems = {},
-	disableChunksDedup = true
+	disableChunksDedup = true,
+	outputDir = resolve('.vercel', 'output', 'static'),
+	otherDirs: DirectoryItems = {}
```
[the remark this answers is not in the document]
james-elicx: yeah it feels like it's getting kinda messy 🙁

src/utils/fs.ts
```diff
@@ -120,10 +120,10 @@ export async function readPathsRecursively(dir: string): Promise<string[]> {
  * Copies a file from one location to another, it also creates the destination
  * directory if it doesn't exist
  *
- * @param sourceFile Original file path.
  * @param destFile Destination for the file.
+ * @param sourceFile Original file path.
  */
-export async function copyFileWithDir(sourceFile: string, destFile: string) {
+export async function copyFileWithDir(destFile: string, sourceFile: string) {
```
dario-piotrowicz, line 126: no 😭 , sorry I was asking the opposite, to change the `copyVercelStaticAssets` function oredering, not the order of these 😢 

as you can see now this function uses a different ordering compared to the standard nodejs one (like 128👇)

(yeah sorry for being annoying on this one but having different orders I think can really trip people, uhm... me, up 😅 😓)

src/cli.ts
james-elicx: > also is it a necessary limitation that the output dir needs to be inside the current working dir? couldn't the output dir be anywhere?

good question. i suppose, yeah, would rather i change it to accomodate that?

src/utils/fs.ts
james-elicx: OHHHHH. I'll take the blame for that one, and I'll hide behind the excuse of it being late at night 🙈🙈

src/cli.ts
dario-piotrowicz: well since we already offer the flexibility to have move the built output I guess we can make it as flexible as we can 🤷, basically I don't see any downside in allowing the user to put any possible location, do you?

(also I am curious about @GregBrimble's outdir usecase, maybe he could need to be able to have the outdir anywhere?)

dario-piotrowicz (approved):
great stuff thanks @james-elicx 😃

I'm approving as I am happy with the changes (and I've tested them locally, all seems to be working as expected)

but please update the outdir/srcdir ordering 🙇 😢 

Pinging @GregBrimble in case he'd also like to have a quick look 🙂

src/cli.ts
james-elicx: > basically I don't see any downside in allowing the user to put any possible location, do you?

my reasoning was kinda just, well, i don't want to let the user accidentally set it to their home directory and wipe all their files... so yeah. im happy to change it if y'all need it to be more flexible though.

src/utils/fs.ts
dario-piotrowicz: ahah no problem 😄

src/cli.ts
dario-piotrowicz: ah yeah that sounds like a valid point... whatever's fine by me, let's see what greg thinks about it 🙂

dario-piotrowicz:
Another thing 😅 

Currently at the end of the build process we get:
Screenshot 2023-05-25 at 12 01 32

But with your changes we get a full absolute path:
Screenshot 2023-05-25 at 11 58 44

could you make it relative to the current working directory or something like that?

packages/next-on-pages/src/buildApplication/processVercelOutput.ts
```diff
@@ -23,6 +27,8 @@ export async function getVercelStaticAssets(): Promise<string[]> {
 		return [];
 	}
 
+	await deleteNextTelemetryFiles(dir);
+
 	return (
```
GregBrimble, line 32: Why is `getVercelStaticAssets()` necessary? Can't I just deal with the entire directory rather than treeing all the files? This would make copying to the custom outdir much easier since I could just `cp -r .vercel/output/static <outdir>` rather doing it individually for each file.

packages/next-on-pages/src/cli.ts
```diff
@@ -4,7 +4,8 @@ import { z } from 'zod';
 import { argumentParser } from 'zodcli';
```
GregBrimble, line 4: Y'all might want to look at [Pastel](https://github.com/vadimdemedes/pastel) now as a replacement for zodcli. It's much better than my thing.

packages/next-on-pages/src/buildApplication/processVercelOutput.ts
james-elicx: The `getVercelStaticAssets` function is primarily used to collect a list of static assets that can be injected into the worker so that it knows what static assets are available during routing.

GregBrimble:
Two pieces of general feedback:

1. We might want to turn on the [`trailingComma: "all"`](https://prettier.io/docs/en/options.html#trailing-commas) option in Prettier to reduce churn on lines unrelated to the actual PR. This makes it much easier for reviewers to focus on the diffed lines that are actually substantive.
2. The output directory is passed to a lot of functions as a parameter. This is totally fine, but just as a matter of style, I personally prefer minimizing the number of parameters that an object takes and instead using a big ol' object which takes a number of parameters inside it. `processStaticFiles({ outputDir, files })` vs `processStaticFiles(outputDir, files)`. Generally, I find this works better for three reasons:

    1. The callee has to specify the variable by explicitly naming it 'correctly'. This gives any reader of the callee file additional context about what that variable is and how it's used by the called function. For example, `processStaticFiles({ outputDir: resolve(cwd(), '.vercel', 'output'), files: ['index.html'] })` vs. `processStaticFiles(resolve(cwd(), '.vercel', 'output'), ['index.html'])`. The former is much clearer about exactly what these parameters are.
    2. It eliminates the need to decide about parameter ordering since the order of keys in an object don't matter. This can reduce line diff churn for reviewers but more importantly, it reduces the likelihood of accidentally forgetting to reorder any callee's parameters.
    3. It's easier to create a object or type interface ahead of actually calling the function. For example:
    
    ```ts
    interface ProcessStaticFilesOptions {
      outputDir: string;
      files: string[]
    }

    function processStaticFiles(options: ProcessStaticFilesOptions) {
      // do stuff
    }

    // callee.ts
    const processStaticFilesOptions: ProcessStaticFilesOptions = {
      outputDir: resolve(cwd(), '.vercel', 'output'),
      files: ['index.html']
    }
    ```

    is easier to write and more legible than:

    ```ts
    function processStaticFiles(outputDir: string, files: string[]) {
      // do stuff
    }

    type ProcessStaticFilesOptions = Parameters<typeof processStaticFiles>

    // callee.ts
    const processStaticFilesOptions: ProcessStaticFilesOptions = [resolve(cwd(), '.vercel', 'output'), ['index.html']]
    ```

packages/next-on-pages/tests/_helpers/index.ts
```diff
@@ -309,3 +311,26 @@ export function createPrerenderedRoute(
 		[`${file}.rsc.prerender-fallback.rsc`]: `${fileWithBase}.rsc.prerender-fallback.rsc`,
 	};
 }
+
+type ConsoleMethods = {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	[Method in keyof Console]: Console[Method] extends (...args: any[]) => any
```
GregBrimble, line 317: Ah, apologies. You can use `unknown` here instead.

```suggestion
	[Method in keyof Console]: Console[Method] extends (...args: unknown[]) => unknown
```

GregBrimble:
Thank you @james-elicx for doing this! Much appreciated :)

packages/next-on-pages/src/buildApplication/processVercelOutput.ts
GregBrimble: Ah, okay! Yeah, that's totally fine. I saw it was used before this PR, but I was just wondering what for.

Out of curiosity, if we offered something like `env.ASSETS.exists(request.url)` and promised a fast response telling you if a file existed or not, would that be of interest? Would that simplify things for next-on-pages? Could you get rid of your own static asset manifest?

packages/next-on-pages/src/cli.ts
dario-piotrowicz: Pastel seems to have a bunch of stuff in it... if you're deprecating `zodcli` I think I'd just fallback to something minimalistic/simpler like jargs, etc...

Pastel feels excessive here, don't you think? (unless we were to redo our whole CLI structure)

packages/next-on-pages/src/buildApplication/processVercelOutput.ts
james-elicx: Yeah, an option like that would mean we don't need to inject all the static assets (which also saves a little bit of bundle size).

The crucial part is of course the response time/reliability of it though, because we check for the existence at multiple points, so there's the potential for it to get hit a lot of times depending on the size/complexity of a project.

packages/next-on-pages/tests/_helpers/index.ts
james-elicx: I did change it to unknown, but it was missing quite a few of the options (e.g. `table` that you mentioned in your previous comment), so I changed it back to any and put the comment disabling eslint for it.

GregBrimble: Weird! Okay, whatever :)

packages/next-on-pages/src/cli.ts
GregBrimble: Up to you — I'm just personally thinking of Pastel as the spiritual successor to zodcli :)

src/cli.ts
dario-piotrowicz: @GregBrimble any preference here?

GregBrimble: No preference. Happy to go with the more restricted option now and we can widen later if need be.

PR #277, opened 2023-05-24
bump to v0.10.2 and add deprecation warning

dario-piotrowicz:
Just opening this for testing purposes (to get the prerelease)


dario-piotrowicz:
This PR is no longer needed, as the v0.10.2 package has already been tested and published

PR #280, opened 2023-05-26
update runtime error message

### LINKED ISSUE #225: [🐛 Bug]: edge runtime error message needs updating
The following error message is no longer valid:
https://github.com/cloudflare/next-on-pages/blob/701f0c257ceed22c19ff85fce8597351a47ef69f/src/buildApplication/buildApplication.ts#L127-L139

Because:
 - the first url (https://nextjs.org/docs/api-routes/edge-api-route) points to a 404 page
 - the latest version of Next you export `runtime` directly (https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#runtime)
 - Next doesn't allow you to set the runtime globally in `next.config.js` anymore
 - the last url (https://nextjs.org/docs/advanced-features/react-18/switchable-runtime) seems no longer valid
 
 ____
 
 The error message should be rewritten so that:
 - the urls are updated or removed, alongside their sentences if needed (whatever is deemed more appropriate/helpful)
 - the `next.config.js` suggestion is removed (since it's not available anymore and I think that it was never recommended)
 - the currently used Next version is checked:
    - if the version is prior to when the `config` export was removed (I believe it was `13.3.1` but that needs to be checked) we keep suggesting to export `const config = { runtime: 'edge' }`
    - for newer Next versions we suggest to export `const runtime = 'edge'`
    
<details>
<summary>Extra notes regarding the version check</summary>

#### Why I think it's needed
Since we haven't yet decided how we want to support Next.js versions, this seems to me the best approach to keep the message correct for any version we currently support (instead of just always suggesting to export `const config = { runtime: 'edge' }`)

#### The version
As I mentioned the delimiter version should be that which removed the `config` export support, but if simpler I think we can instead check when the `runtime` export was introduced and use that version as the delimiter instead (basically what I think is important here is that we don't suggest to do export `const runtime = 'edge'` when that is going to break the build, if we suggest it, it should work).

</details>

dario-piotrowicz:
resolves #225

____

Before:
Screenshot 2023-05-26 at 09 56 18

After:
on a post 13.1.2 app:
Screenshot 2023-05-26 at 09 51 02

on a pre 13.1.2 app:
Screenshot 2023-05-26 at 09 50 45


src/buildApplication/buildApplication.ts
```diff
@@ -115,28 +119,39 @@ async function prepareAndBuildWorker(
...
 		`
 		ERROR: Failed to produce a Cloudflare Pages build from the project.
 
-			The following functions were not configured to run with the Edge Runtime:\n${invalidFunctions
-				.map(fn => `				- ${fn}`)
+			The following routes were not configured to run with the Edge Runtime:\n${invalidRoutes
+				.map(route => `			  - ${route}`)
 				.join('\n')}
 
-			If this is a Next.js project:
```
dario-piotrowicz, line 127: Currently (and for the foreseeable future) the adaptor only supports `Next.js` so I don't really think that we should include this (it just adds confusion, besides adding it back when/if we'll support other frameworks should be rather trivial)

petebacondarwin (approved):

PR #282, opened 2023-05-26
fix: avoid copying rsc funcs to output dir

james-elicx:
This PR does the following:
- Removes any copied rsc function files (`.rsc.func.js`) from the worker output directory (we do not use them during routing due to them being the same as non-rsc funcs)
- Filters the static assets list that we provide to the routing system to remove the contents of the `_worker.js` directory (this should never have happened in the first place).

This change is primarily to reduce the bundle size by removing files that we don't utilize.

The tiny app that I tested it with had 1 rsc function. The zipped (`tar -zcvf`) sizes were as follows:
- Before: 192451
- After: 179682
- Change: -12769

On my modified version of the [app dir playground](https://github.com/james-elicx/app-playground), the difference is very noticeable:
- Before: 1298780
- After: 858230
- Change: -440550

### DIFF
--- a/src/buildApplication/processVercelOutput.ts
+++ b/src/buildApplication/processVercelOutput.ts
@@ -68,7 +68,7 @@ export function processVercelOutput(
 			entrypoint: value.replace(/\.rsc\.func\.js$/i, '.func.js'),
 		});
 
-		// Remove the RSC functions from the dist directory as they are not needed.
+		// Remove the RSC functions from the dist directory as they are not needed since we replace them with non-rsc variants for the runtime routing
 		if (/\.rsc\.func\.js$/i.test(value)) {
 			rmSync(value);
 		}
--- a/src/buildApplication/processVercelOutput.ts
+++ b/src/buildApplication/processVercelOutput.ts
@@ -27,7 +27,7 @@ export async function getVercelStaticAssets(): Promise<string[]> {
 		(await readPathsRecursively(dir))
 			.map(file => addLeadingSlash(normalizePath(relative(dir, file))))
 			// Filter out the worker script output directory contents as those are not valid static assets.
-			.filter(path => !/^\/?_worker\.js\//.test(path))
+			.filter(path => !/^\/_worker\.js\//.test(path))
 	);
 }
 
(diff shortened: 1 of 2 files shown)


src/buildApplication/processVercelOutput.ts
```diff
@@ -22,8 +23,11 @@ export async function getVercelStaticAssets(): Promise<string[]> {
 		return [];
 	}
 
-	return (await readPathsRecursively(dir)).map(file =>
-		addLeadingSlash(normalizePath(relative(dir, file)))
+	return (
+		(await readPathsRecursively(dir))
+			.map(file => addLeadingSlash(normalizePath(relative(dir, file))))
+			// Filter out the worker script output directory contents as those are not valid static assets.
+			.filter(path => !/^\/_worker\.js\//.test(path))
```
dario-piotrowicz, line 30: I don't think that files inside the `_worker.js` directory are uploaded as assets, so this is probably non-necessary? 🤔 (not that it hurts anything either 🤔)

james-elicx: They are collected as assets and injected into the worker file in the list of different assets/functions available. Filtering this list is just trimming down a little bit of the unnecessary fat in the worker index.js file.

dario-piotrowicz: yeah sorry, wrangler's totally unrelated here! my bad 🙇

src/buildApplication/processVercelOutput.ts
```diff
@@ -63,6 +67,11 @@ export function processVercelOutput(
 			// `TypeError: Cannot read properties of undefined (reading 'default')`
 			entrypoint: value.replace(/\.rsc\.func\.js$/i, '.func.js'),
 		});
+
+		// Remove the RSC functions from the dist directory as they are not needed.
```
dario-piotrowicz, line 71:
```suggestion
		// Remove the RSC functions from the dist directory as they are not needed since we replace them with non-rsc variants for the runtime routing
```
or something along those lines, I think it would be valuable to explain the why here 🙂

dario-piotrowicz (approved):
Awesome thanks! 😃 

I just left a small suggestion 🙂

Issue #283, opened 2023-05-27
[🐛 Bug]: Returned NextResponse from middleware is ignored


bharatari:
### next-on-pages environment related information

System:
        Platform: linux
        Arch: x64
        Version: #1 SMP Fri Jan 27 02:56:13 UTC 2023
        CPU: (12) x64 12th Gen Intel(R) Core(TM) i7-1255U
        Memory: 8 GB
        Shell: /bin/bash
Binaries:
        Node: 18.15.0
        Yarn: 1.22.19
        npm: 9.5.0
        pnpm: 8.3.1
Package Manager Used: npm
Relevant Packages:
        @cloudflare/next-on-pages: 1.0.1
        vercel: N/A
        next: 13.4.4

### Description

When a `new NextResponse` is returned from middleware, the response seems to be ignored. Instead of rendering what is returned from middleware, the route itself is rendered. 

`next dev` (expected behavior):
![image](https://github.com/cloudflare/next-on-pages/assets/5521868/b559dd1f-3307-465a-8db7-b5fca3e920a2)

`wrangler` (incorrect):
![image](https://github.com/cloudflare/next-on-pages/assets/5521868/2f719622-ddb3-426f-8f9d-a57ea5b74de3)

As of Next 13.1, this form of returning a new NextResponse from middleware should be valid: 
https://nextjs.org/docs/app/building-your-application/routing/middleware#producing-a-response

### Reproduction

Here is a repo demonstrating the issue: https://github.com/bharatari/middleware-error. I generated the app with create-next-app, added the edge runtime config to the page and added a middleware file. 

If you run this app with `wrangler`, you'll see that the route is rendered instead of the response returned from the middleware.

With `next dev`, you'll see the middleware response rendered (as expected).

### Pages Deployment Method

None

### Additional Information

Additional context in case it is helpful (although not necessary to debug the issue):

I noticed this while using the Next.js middleware from `clerk/nextjs` (an authentication provider). The middleware returns an interstitial page to refresh an expired JWT token, but with next-on-pages, this interstitial page isn't rendered even when it's returned from middleware. The app goes on to render the route even though authentication hasn't actually completed yet.

### Would you like to help?

- [ ] Would you like to help fixing this bug?

james-elicx:
Thank you very much for providing a reproduction and opening this bug report.

It looks like we didn't quite implement a certain scenario properly with how we handled middleware 🙁

[...]

PR #284, opened 2023-05-28
fix: middleware not returning response with `new NextResponse`

james-elicx:
This PR does the following:
- Modifies the middleware handling to check for the `x-middleware-next` header and act as follows:
  - If the header exists, remove it and continue as normal.
  - If the header does not exist, and there was no rewrite or location header, this means the middleware response should be the final response we return to the client.
    - Therefore, update the final response body and status.
- Changes the middleware error check to only be for status 500 which is what occurs when an error is thrown in a middleware func.
- Checks if a response body was set during middleware processing and if so, bails out of routing and also uses it when returning a response to the client.
- Adds a test for the new logic.

fixes https://github.com/cloudflare/next-on-pages/issues/283

### DIFF
--- a/templates/_worker.js/handleRequest.ts
+++ b/templates/_worker.js/handleRequest.ts
@@ -49,6 +49,7 @@ async function findMatch(
 		status: matcher.status,
 		headers: matcher.headers,
 		searchParams: matcher.searchParams,
+		body: matcher.body,
 	};
 }
 
@@ -61,7 +62,7 @@ async function findMatch(
  */
 async function generateResponse(
 	reqCtx: RequestContext,
-	{ path = '/404', status, headers, searchParams }: MatchedSet,
+	{ path = '/404', status, headers, searchParams, body }: MatchedSet,
 	output: VercelBuildOutput
 ): Promise<Response> {
 	// Redirect user to external URL for redirects.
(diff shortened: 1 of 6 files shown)


templates/_worker.js/handleRequest.ts
```diff
@@ -76,12 +77,16 @@ async function generateResponse(
...
-	let resp = await runOrFetchBuildOutputItem(output[path], reqCtx, {
-		path,
-		status,
-		headers,
-		searchParams,
-	});
+	let resp =
+		body !== undefined
+			? // If we have a response body from matching, use it instead.
+			  new Response(body, { status })
```
james-elicx, line 83: This probably isn't the *best* way to deal with a body existing, but I guess it'll do for now, at least until this function eventually gets refactored

dario-piotrowicz: why do you think it isn't the best way to deal with it? what would be? is there a need to refactor this function?

templates/_worker.js/routes-matcher.ts
```diff
@@ -165,6 +167,17 @@ export class RoutesMatcher {
 			resp.headers.delete(rewriteKey);
 		}
 
+		const nextKey = 'x-middleware-next';
+		const nextHeader = resp.headers.get(nextKey);
+		if (nextHeader) {
+			resp.headers.delete(nextKey);
```
dario-piotrowicz, line 173: Nit

```suggestion
		const middlewareNextKey = 'x-middleware-next';
		const middlewareNextHeader = resp.headers.get(middlewareNextKey);
		if (middlewareNextHeader) {
			resp.headers.delete(middlewareNextKey);
```

templates/_worker.js/handleRequest.ts
james-elicx: > why do you think it isn't the best way to deal with it? what would be? is there a need to refactor this function?

because it would now be unnecessarily creating the response twice in this case. here and then just a little bit below.

dario-piotrowicz: ah yeah... I see 🤔

dario-piotrowicz (approved):
Awesome, thanks a lot @james-elicx 😃 

Before merging, could you please add a changeset? 🙂

PR #289, opened 2023-05-31
Update documentation around `next/cache`

leerob:
Hey folks! I want to clarify in the documentation here that it is possible to provide a custom cache handler, and backlink to the documentation for how this is possible. Using `next/cache` is not dependent on Vercel's infrastructure (you can use whatever cache you prefer, for example in the future I would imagine it could very easily support KV or Durable Objects).

Hope this helps!


docs/supported.md
```diff
@@ -100,7 +100,7 @@ Earlier and Later versions might be only partially supported, we don't fully kno
 
 Incremental Static Regeneration (ISR) is a rendering mode in Next.js that allows you to automatically cache and periodically regenerate pages with fresh data. Next.js [does not support](https://nextjs.org/docs/pages/building-your-application/rendering/incremental-static-regeneration) building ISR pages for the edge runtime, and as such, pages should be changed to use server side rendering (SSR) instead.
 
-ISR pages are built by the Vercel CLI to something called prerender functions. These are Node.js Lambda functions that can be called in the background while serving the page from the cache. It is not possible to use these with Cloudflare Pages and they are not compatible with the edge runtime.
+ISR pages are built by the Vercel CLI to generate Vercel Serverless Functions. These are Node.js Lambda functions that can be called in the background while serving the page from the cache. It is not possible to use these with Cloudflare Pages and they are not compatible with the [edge runtime](https://nextjs.org/docs/app/api-reference/edge) currently.
```
james-elicx, line 103: I'm not sure I agree with this change. I specifically referred to them as prerender functions here because there are two types of serverless functions, and ISR uses the prerender variant.
- [Prerender functions](https://vercel.com/docs/build-output-api/v3/primitives#prerender-functions) that come with a prerender config and fallback file.
- [Serverless functions](https://vercel.com/docs/build-output-api/v3/primitives#serverless-functions) that are always run.

Another distinction relevant to this library is that in the case of prerender functions, we use the fallback files, but with regular serverless functions, we error during the build.

Perhaps it would be better if we called them Prerender functions but included a link to the Vercel docs on them?

```suggestion
ISR pages are built by the Vercel CLI to generate Vercel [Prerender Functions](https://vercel.com/docs/build-output-api/v3/primitives#prerender-functions). These are Node.js Lambda functions that can be called in the background while serving the page from the cache. It is not possible to use these with Cloudflare Pages and they are not compatible with the [edge runtime](https://nextjs.org/docs/app/api-reference/edge) currently.
```

james-elicx:
> Hey folks! I want to clarify in the documentation here that it is possible to provide a custom cache handler, and backlink to the documentation for how this is possible. Using `next/cache` is not dependent on Vercel's infrastructure (you can use whatever cache you prefer, for example in the future I would imagine it could very easily support KV or Durable Objects).
> 
> Hope this helps!

Hey there, thanks for the PR clarifying our docs around the incremental cache 🙂

I have just been trying to use a custom cache handler, per the (new?) Next.js documentation that you referenced, and took the example code and added a `revalidateTag` method. Unfortunately, though, I wasn't able to get much luck - it seemed to have no effect on what happened when calling `revalidatePath` and `revalidateTag` 🙁.

I wonder if this might be because we use the build output from Vercel, instead of the standard Next.js server/build? When using the Next.js node server, it worked as you'd expect, but the next-on-pages build that used the Vercel build output did not go through the custom cache handler.

Regarding your comment on Vercel infra, I initially wrote that it was dependent on it due to how the default cache implementation that comes out-of-the-box works. The [line](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L38) in the code that I referenced was where an URL injected by Vercel to the request headers (or taken from an env variable) tells the fetch cache which endpoint to use for tag revalidation. I'll agree that that might not have been the best way I could have phrased it.

For reference, the default cache implementation used by `revalidateTag` causes the following error when deploying to Pages as we don't provide a suspense cache URL.
```
Failed to revalidate tag / TypeError: Fetch API cannot load: undefined/v1/suspense-cache/revalidate?tags=/
```

bluebeel:
> > Hey folks! I want to clarify in the documentation here that it is possible to provide a custom cache handler, and backlink to the documentation for how this is possible. Using `next/cache` is not dependent on Vercel's infrastructure (you can use whatever cache you prefer, for example in the future I would imagine it could very easily support KV or Durable Objects).
> > Hope this helps!
> 
> Hey there, thanks for the PR clarifying our docs around the incremental cache 🙂
> 
> I have just been trying to use a custom cache handler, per the (new?) Next.js documentation that you referenced, and took the example code and added a `revalidateTag` method. Unfortunately, though, I wasn't able to get much luck - it seemed to have no effect on what happened when calling `revalidatePath` and `revalidateTag` 🙁.
> 
> I wonder if this might be because we process and use the build output from Vercel, instead of the standard Next.js server/build? When using the Next.js node server, it worked as you'd expect, but the next-on-pages build that used the Vercel build output did not go through the custom cache handler.
> 
> Regarding your comment on Vercel infra, I initially wrote that it was dependent on it due to how the default cache implementation that comes out-of-the-box works. The [line](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L38) in the code that I referenced was where an URL injected by Vercel to the request headers (or taken from an env variable) tells the fetch cache which endpoint to use for tag revalidation. I'll agree that that might not have been the best way I could have phrased it.
> 
> For reference, the default cache implementation used by `revalidateTag` causes the following error when deploying to Pages as we don't provide a suspense cache URL.
> 
> ```
> Failed to revalidate tag / TypeError: Fetch API cannot load: undefined/v1/suspense-cache/revalidate?tags=/
> ```

I have a branch where I managed to make the next/cache works on Page by adding a new function to next-on-pages. I'll make the pr this weekend or beginning of next week.

My implementation make the fetch cache the data in the cache API of the datacenter. I added a optional binding that if the user set it up, it can even save the cache in a KV

james-elicx:
> I have a branch where I managed to make the next/cache works on Page by adding a new function to next-on-pages. I'll make the pr this weekend or beginning of next week.
> 
> My implementation make the fetch cache the data in the cache API of the datacenter. I added a optional binding that if the user set it up, it can even save the cache in a KV
> 

I look forward to reading more about your approach when you open a PR @bluebeel :)

Does it use the custom incremental cache handler that Lee referenced in his change suggestions, or is it hooking into the default cache instead?

bluebeel:
I tried the custom hook but I discarded it because it means that we need to patch the next config of the user to add an experimental config. Not the biggest fan of that 😅

Nextjs already has a cache handler that matches what we need, the FetchCache. The FetchCache is mainly for pages that have the runtime edge because they run in workers-land and don't have access to the possible EFS linked to the lambas in AWS and used by the FileSystemCache.

I even introduced a `NextFetch` function that is the equivalent of the patched fetch from nextjs so the user can use the next/cache feature.

I'm cleaning the code a bit and making the pr

leerob:
You are correct that the incremental cache handler works when using Node.js. For processing through the Vercel Build Output API, we don't (yet) have a spec for how to define that cache output. It would still be helpful to merge or update this to be more clear that it's not dependent on Vercel 👍

GregBrimble (approved):

docs/supported.md
```diff
@@ -139,9 +139,9 @@ export async function getStaticPaths() {
...
+Revalidation and `next/cache` are not supported on Cloudflare Pages. This is used by the default `fetch` cache, which forms part of the incremental cache for revalidating data inside the App Router. Revalidating tags and data for an entire path also uses `next/cache`.
 
-Revalidating tags and paths goes through this functionality and appears to be what `next/cache` is built around. Because of this, it is highly unlikely that the default `next/cache` functionality will be supported on Cloudflare Pages.
+Because of this, `next/cache` functionality currently is not supported on Cloudflare Pages. The Next.js cache does work when self-hosting, and can be optionally provided a [custom cache handler](https://nextjs.org/docs/app/api-reference/next-config-js/incrementalCacheHandlerPath). It's possible this could use Cloudflare KV or Durable Objects in the future.
```
dario-piotrowicz, line 144: The previous paragraph mentions that `next/cache` is not supported on Cloudflare Pages, and here in this paragraph we say "Because of this, `next/cache` functionality currently is not supported on Cloudflare Pages."
That looks a bit incorrect and redundant to me. 🤔 

```suggestion
The Next.js cache does however work when self-hosting by optionally providing a [custom cache handler](https://nextjs.org/docs/app/api-reference/next-config-js/incrementalCacheHandlerPath). It's possible this could use Cloudflare KV or Durable Objects in the future.
```

dario-piotrowicz (approved):
@leerob Thanks a lot for the doc change!

It's really appreciated and it does help! 😃

I've added a very small suggestion, nothing major just some wording tweaking, besides that everything looks perfect 😄

PR #291, opened 2023-06-02
fix: non-index pages with trailing slash rewriting to `/index` for rsc requests

james-elicx:
This PR does the following:
- Alter the regex for the special handling of rewriting `^/` to `/index` to check that the index path is at the start of the pathname.

We previously added special handling for the build output config rule rewriting `^/` to `/index` when the `rsc` header is present as it was causing requests to `/non-index-path` to be rewritten. The handling involved verifying that the previous path before the rewrite was actually an index path. However, the regex wasn't checking that it started with the index path, only that one existed in the path, i.e. `*/` or `*/index`.

This meant that it would also match requests with a trailing slash by accident (e.g. `/non-index-path/`, so this change alters the regex to ensure that it is matching at the start of the path.

For reference, the config rule that this relates to:
```ts
{
  src: '^/',
  has: [{ type: 'header', key: 'rsc' }],
  dest: '/index.rsc',
  headers: {
    vary: 'RSC, Next-Router-State-Tree, Next-Router-Prefetch',
  },
  continue: true,
  override: true,
}
```

### DIFF
--- a/.changeset/smooth-pumpkins-press.md
+++ b/.changeset/smooth-pumpkins-press.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix non-index pages with trailing slash rewriting to /index for rsc requests.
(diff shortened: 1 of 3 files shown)


dario-piotrowicz (approved):
Awesome, thanks! 😃

PR #299, opened 2023-06-07
fix: `set-cookie` headers overriding instead of appending

james-elicx:
This PR does the following:
- Appends headers when they are for `set-cookie`, instead of overriding the existing value.
- Adds a test for the fix.

This change fixes a bug where multiple `set-cookie` headers would result in only the final one being applied. With this change, they are all applied and included.

### DIFF
--- a/.changeset/red-crabs-cross.md
+++ b/.changeset/red-crabs-cross.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix headers overriding when more than one is set, instead of appending.
--- a/templates/_worker.js/utils/http.ts
+++ b/templates/_worker.js/utils/http.ts
@@ -16,12 +16,16 @@ export function applyHeaders(
 	const entries =
 		source instanceof Headers ? source.entries() : Object.entries(source);
 	for (const [key, value] of entries) {
-		target.set(
-			key.toLowerCase(),
-			pcreMatch?.match
-				? applyPCREMatches(value, pcreMatch.match, pcreMatch.captureGroupKeys)
-				: value
-		);
+		const lowerKey = key.toLowerCase();
+		const newValue = pcreMatch?.match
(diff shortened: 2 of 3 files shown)


.changeset/red-crabs-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix headers overriding when more than one is set, instead of appending.
```
james-elicx, line 5:
```suggestion
Fix `set-cookie` headers overriding when more than one is set, instead of appending.
```

templates/_worker.js/utils/http.ts
```diff
@@ -16,12 +16,16 @@ export function applyHeaders(
...
-		);
+		const lowerKey = key.toLowerCase();
+		const newValue = pcreMatch?.match
+			? applyPCREMatches(value, pcreMatch.match, pcreMatch.captureGroupKeys)
+			: value;
+
+		if (lowerKey === 'set-cookie') {
+			target.append(lowerKey, newValue);
+		} else {
+			target.set(lowerKey, newValue);
```
dario-piotrowicz, line 27: can't we alway use `append`?

dario-piotrowicz (approved):

templates/_worker.js/utils/http.ts
james-elicx: Afraid not, we seem to end up with multiple `X-Matched-Path` headers when doing so.

dario-piotrowicz: and naturally I can't see your comment 😅

joelket23:
Hey @james-elicx  and @dario-piotrowicz , This PR is working perfectly fine. However, when I modified the build configuration to use `"npx @cloudflare/next-on-pages@1"`, it started utilizing the latest version `1.0.2 of next-on-pages`, but I'm still `encountering the same cookie error`. Any idea why?

PR #303, opened 2023-06-09
convert next-on-pages to a monorepo using turborepo and add eslint plugin package

dario-piotrowicz:


jculvey (approved):

internal-packages/eslint-config-next-on-pages/index.js
```diff
@@ -8,7 +8,7 @@
 	"parser": "@typescript-eslint/parser",
 	"parserOptions": {
 		"sourceType": "module",
-		"project": "./tsconfig.json"
+		"project": "tsconfig.json"
```
petebacondarwin, line 11: Why this change and what does it mean to have no ./?

packages/next-on-pages/package.json
```diff
@@ -0,0 +1,75 @@
...
+		"dedent-tabs": "^0.10.3",
+		"eslint": "^8.35.0",
+		"mock-fs": "^5.2.0",
+		"p-limit": "^4.0.0",
+		"prettier": "^2.8.4",
+		"typescript": "^5.0.4",
+		"vitest": "^0.31.0",
+		"vitest-environment-miniflare": "^2.13.0",
+		"next-on-pages-tsconfig": "*",
+		"eslint-config-next-on-pages": "*"
```
petebacondarwin, line 73: In the workers-sdk repo, it was decided that these synthetic packages should use `@cloudflare/` prefix to avoid a future problem where someone actually publishes to that package name and messes us up

packages/next-on-pages/README.md
```diff
@@ -0,0 +1,3 @@
+# `@cloudflare/next-on-pages`
```
petebacondarwin, line 1: I think that npm will grab the README from here rather than the root of the monorepo.
I wonder if we should move the useful content about this package here?

petebacondarwin (commented):
I see we have some changeset dependencies and the CHANGELOG appears to use that format. But I can't see where we are using changeset to do publishing. Is that correct?

petebacondarwin (commented):
No test runs on CI? Or am I missing something?

petebacondarwin (approved):

packages/next-on-pages/package.json
dario-piotrowicz: sure I'll update them 🙂👍 

(PS: but the eslint-plugin one needs to stay as is for eslint convention 😓 (so that you can just refer to it as `next-on-pages` in your eslintrc without needing to specify the whole package name) )

petebacondarwin: See https://github.com/cloudflare/workers-sdk/blob/9fd12df8d65d50ba82b927eea48de679d622b9eb/packages/wrangler/.eslintrc.js#L3

packages/next-on-pages/README.md
dario-piotrowicz: yeah, thanks a lot for the comment this is exactly what I was wondering about

Would you like us to have the next-on-pages details here? but then what do we have in the root readme? just basic info and links to the packages readmes?

I wasn't sure about that since the repo would be a monorepo but in practice its still very much just for next-on-pages, so ideally I think it would be nice to have info there as well...

Basically I am thinking about people visiting the repository, I think ideally they should see all the various next-on-pages and getting started steps without having to go to that specific package... (and we'd definitely not want to duplicate the info...)

I'm a bit conflicted on this one (since all solution have some drawbacks unfortunately), but I think I'll give the above (basic info at the root) version a shot 🤔

petebacondarwin: Hmm...

So the Angular monorepo goes the other way and directs people back to the monorepo root for the README: https://www.npmjs.com/package/@angular/core

Remix points you at their docs website: https://www.npmjs.com/package/remix as does React.

So perhaps it is not so bad as it is - as long as there is a clear link from the package's README to the monorepo README.

internal-packages/eslint-config-next-on-pages/index.js
dario-piotrowicz: I'll have to have a look, I can't remember why I made the change honestly 😓 

I think it's either that it wasn't working as intended or that `./` doesn't make any difference 🤔

packages/next-on-pages/package.json
dario-piotrowicz: Yes, but I think that that's just an internal package that is then extended: https://github.com/search?q=repo%3Acloudflare%2Fworkers-sdk%20%2Feslint-config-worker%2F&type=code

What I meant is that if the package is called `eslint-plugin-next-on-pages` then in your eslintrc you can refer to its rules as `next-on-pages/rule_name` instead of having to specify the whole package name  (the `eslint-plugin` is inferred) so if i where to rename the package to `@cloudflare/eslint-plugin-next-on-pages` (or anything else that doesn't start with `eslint-plugin`) then users would have to specify rules as: `@cloudflare/eslint-pluging-next-on-pages/rule_name`

which seems much worse to me, isn't it?

<details>

<summary>qwik example</summary>

For example qwik also does the same.

They have [`@builder.io/qwik`](https://www.npmjs.com/package/@builder.io/qwik), [`@builder.io/qwik-city`](https://www.npmjs.com/package/@builder.io/qwik-city) and then without the `@builder.io` prefix [`eslint-plugin-qwik`](https://www.npmjs.com/package/eslint-plugin-qwik)

</details>

petebacondarwin: Oh sorry. Yes. The external facing package should be named appropriately.
But any internal packages should be `@cloudflare` prefixed.
Sorry for the confusion.

dario-piotrowicz: no problem, sorry maybe I caused the confusion 😅 

anyways I've updated the internal packages names 🙂👍

packages/next-on-pages/README.md
dario-piotrowicz: I've updated the two READMEs, @petebacondarwin and @GregBrimble please have a look and let me know what you think 🙂 🙏

internal-packages/eslint-config-next-on-pages/index.js
dario-piotrowicz: Yeah I just checked and it doesn't seem to make any difference, the value there is a glob path: https://typescript-eslint.io/linting/typed-linting#specifying-tsconfigs so I think that unless it starts with `/` it is considered relative

I can add back the `./` if you find that more explicit/clear

I could also set this to `true` and it should pick up the correct `tsconfig` anyways if you'd prefer that 🙂 

___

Actually I did set it to `true` since it seems the proper solution given that the package is shared and doesn't have a `tsconfig` itself. So with `true` it feels to me clearer that we just delegate the `tsconfig` lookup to the consuming package, what do you think?

PR #311, opened 2023-06-13
remove `next-on-pages-vercel-cli` package usage

dario-piotrowicz:


package.json
```diff
@@ -49,7 +49,7 @@
 		"zodcli": "^0.0.4"
 	},
 	"peerDependencies": {
-		"next-on-pages-vercel-cli": "29.3.0-next-on-pages-fix-3",
+		"vercel": "^30.0.0",
```
GregBrimble, line 52: Just checking: was this the same version we had before?

GregBrimble (approved):

package.json
dario-piotrowicz: No, this is the latest vercel version, I set it so that we are up to date with them, I did test it and seems to be working fine with next-on-pages, so it should be all good 🙂

PR #312, opened 2023-06-13
fix: intercept routes being unintentionally stripped

### LINKED ISSUE #310: [🐛 Bug]: Next.js app directory route interception doesn't work properly

System:
        Platform: linux
        Arch: x64
        Version: #1 SMP Fri Jan 27 02:56:13 UTC 2023
        CPU: (12) x64 12th Gen Intel(R) Core(TM) i7-1255U
        Memory: 8 GB
        Shell: /bin/bash
Binaries:
        Node: 18.15.0
        Yarn: 1.22.19
        npm: 9.5.0
        pnpm: 8.3.1
Package Manager Used: npm
Relevant Packages:
        @cloudflare/next-on-pages: 1.0.2
        vercel: [object Promise]
        next: [object Promise]

### Description

Route interception as described in the Next.js docs [here](https://nextjs.org/docs/app/building-your-application/routing/intercepting-routes) doesn't seem to work properly with `next-on-pages`. [Parallel routes](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes) **do** work, it's just route interception (directories with `(.), (..), (...)` prefixes) that don't work.

### Reproduction

Here is a minimal repro of the issue: https://github.com/bharatari/route-interception.

`next dev`
Route `/`
Screenshot 2023-06-13 103657
Route `/hello` after clicking the button (which is a Link to /hello). Note that interception happens as expected here, the modal is shown instead of rendering the actual `/hello` route.
Screenshot 2023-06-13 103721
Route `/hello` when loading the URL directly. This is the case with no interception, the actual route is rendered.
Screenshot 2023-06-13 103737

`wrangler`
Route `/`
Screenshot 2023-06-13 104034
Route `/hello` after clicking the button (which is a Link to /hello). The `/hello` route is loaded, interception **does not** work as expected here.
Screenshot 2023-06-13 104100
Route `/hello` when loading the URL directly. This is the case with no interception.
Screenshot 2023-06-13 104047

When run with `wrangler`, intercepted routes as defined in the `app` directory are completely ignored. The actual route is rendered every time.

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Fixes the regex used to strip route groups to not strip route intercepts.

Intercept routes use the format `/(.)path` which was being accidentally caught in the route group stripping, since route groups use `/(name)`. This PR changes that behaviour to exclude brackets that contain periods from matching the route group regex.

fixes #310

### DIFF
--- a/.changeset/young-beers-buy.md
+++ b/.changeset/young-beers-buy.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Prevent the route group stripping regex from removing intercept routes.
--- a/src/utils/routing.ts
+++ b/src/utils/routing.ts
@@ -10,7 +10,7 @@ import { normalizePath } from './fs';
  * @returns Path name with route groups stripped.
  */
 export function stripRouteGroups(path: string) {
-	return path.replace(/\/\(([^)]+)\)/g, '');
+	return path.replace(/\/\(([^).]+)\)/g, '');
 }
 
 /**
--- a/src/utils/routing.ts
+++ b/src/utils/routing.ts
@@ -5,6 +5,9 @@ import { normalizePath } from './fs';
  *
  * The build output config does not rewrite requests to route groups, so we need to strip route
  * groups from the path name for file system matching.
+ * 
+ * Does not strip brackets containing `.` characters, as these are used for route intercepts.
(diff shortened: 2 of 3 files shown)


dario-piotrowicz (approved):
Awesome stuff as always 😍

PR #313, opened 2023-06-13
feat: wildcard option support in the build output config

james-elicx:
This PR does the following:
- Introduces support for the [`wildcard`](https://vercel.com/docs/build-output-api/v3/configuration#wildcard) option in the Vercel build output config.
- Applies wildcard matches to the route destination.
- Adds a test for the new functionality.
- Removes the `prevMatch` optional input for the `RoutesMatcher` constructor since it is never used.

### DIFF
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -24,6 +24,8 @@ export class RoutesMatcher {
 	private url: URL;
 	/** Cookies from the request to match */
 	private cookies: Record<string, string>;
+	/** Wildcard match from the Vercel build output config */
+	private wildcardMatch: VercelWildCard | undefined;
 
 	/** Path for the matched route */
 	public path: string;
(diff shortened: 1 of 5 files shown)


templates/_worker.js/routes-matcher.ts
```diff
@@ -57,21 +59,21 @@ export class RoutesMatcher {
...
 		this.url = new URL(reqCtx.request.url);
 		this.cookies = parse(reqCtx.request.headers.get('cookie') || '');
 
-		this.path = prevMatch?.path || this.url.pathname || '/';
-		this.status = prevMatch?.status;
-		this.headers = prevMatch?.headers || {
-			normal: new Headers(),
-			important: new Headers(),
-		};
-		this.searchParams = prevMatch?.searchParams || new URLSearchParams();
```
dario-piotrowicz, line 71: was this not necessary? 🤔 😕

james-elicx: I think the `prevMatch` was just left over from the initial version before it was converted to a class and has never been used since. No point in keeping it around.

dario-piotrowicz: ah! we we're not even passing it to the function??

dario-piotrowicz: ah yeah we're not 😮👍

dario-piotrowicz (approved):
Awesome stuff! thanks! 😍👍

PR #315, opened 2023-06-14
fix: i18n locale index request not matching for ssr'd index route

james-elicx:
This PR does the following:
- Introduces yet another hack for i18n to make it work properly.
- Ensures that requests to i18n index routes (e.g. `/nl`) can be rewritten to `/` in the `miss` phase by modifying the relevant source route's regex.

As we know from before, the build output config is rather iffy when it comes to `next.config.js` internationalization. The support for it appears to be half-baked. This PR adds another hack to resolve another regex issue for i18n in the build output config. At the moment, our support for statically generated index routes works fine, however, it appears that there is an issue when you use an SSR'd index route that generates an edge function.

In the `miss` phase, there is a source route as follows that is designed to rewrite any requests from `/{locale}/...` to `/...`.
```ts
{
  "src": "^//?(?:en|fr|nl)/(.*)",
  "dest": "/$1",
  "check": true
}
```
The problem with this regex is that it does not hit requests to `/{locale}`. It only hits when there is a trailing slash present, but it should always hit since this is the `miss` phase. When it fails to find a file to use for a path with i18n, it should be checking it without the locale present. In the case of requests to `/{locale}`, it should then check for a file for `/` if it can't find one.

Therefore, this hack modifies the regex to make the `/(.*)` optional.

`^//?(?:en|fr|nl)/(.*)` -> `^//?(?:en|fr|nl)(?:/(.*))?$`

Ugh. Supporting `next.config.js` i18n isn't that nice.

Fixes the issue encountered in the reproduction in https://github.com/cloudflare/next-on-pages/issues/37#issuecomment-1591894407

### DIFF
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -399,7 +399,9 @@ export class RoutesMatcher {
 			return route;
 		}
 
-		if (/^\//.test(route.src) && route.src.slice(1) in this.locales) {
+		const isLocaleIndex =
+			/^\//.test(route.src) && route.src.slice(1) in this.locales;
+		if (isLocaleIndex) {
 			return { ...route, src: `^${route.src}$` };
 		}
 
--- a/templates/_worker.js/routes-matcher.ts
+++ b/templates/_worker.js/routes-matcher.ts
@@ -1,6 +1,6 @@
 import { parse } from 'cookie';
 import type { MatchPCREResult } from './utils';
-import { parseAcceptLanguage } from './utils';
+import { isLocaleTrailingSlashRegex, parseAcceptLanguage } from './utils';
 import {
 	applyHeaders,
 	applyPCREMatches,
@@ -405,14 +405,7 @@ export class RoutesMatcher {
 			return { ...route, src: `^${route.src}$` };
 		}
 
-		const trailingSlashRegex = /^\^\/\/\?\(\?:([a-zA-Z|]+)\)\/\(\.\*\)$/;
-		const trailingSlashMatch = trailingSlashRegex.exec(route.src);
-		if (
-			trailingSlashMatch?.[1]
-				?.split('|')
-				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-				?.every(locale => locale in this.locales!)
-		) {
+		if (isLocaleTrailingSlashRegex(route.src, this.locales)) {
 			return { ...route, src: route.src.replace(/\/\(\.\*\)$/, '(?:/(.*))?$') };
 		}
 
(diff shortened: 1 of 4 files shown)


templates/_worker.js/routes-matcher.ts
```diff
@@ -364,19 +369,26 @@ export class RoutesMatcher {
...
 	): VercelSource {
+		if (!this.locales || phase !== 'miss') {
+			return route;
+		}
+
+		if (/^\//.test(route.src) && route.src.slice(1) in this.locales) {
+			return { ...route, src: `^${route.src}$` };
+		}
+
+		const trailingSlashRegex = /^\^\/\/\?\(\?:([a-zA-Z|]+)\)\/\(\.\*\)$/;
```
dario-piotrowicz, line 380: A regex of a regex is kind of difficult to follow I think 😓, what about refactoring this in a more readable (although more verbose) way?

Something like this for example?
Screenshot 2023-06-15 at 11 24 09

If you prefer, it can be a method of the class, it can also still use regexes, etc... but I feel like at the very least separating it and adding comments explaining what's going on would actually help making it more understandable and maintainable

what do you think? 🙂

dario-piotrowicz (approved):
Awesome, thanks! 😄

tests/templates/requestTestData/i18n.ts
```diff
@@ -163,7 +168,10 @@ export const testSet: TestSet = {
 			...locales.reduce(
 				(acc, locale) => ({
 					...acc,
-					[`${locale}.html`]: `<html>${locale}</html>`,
+					// @ts-expect-error - static locales is a subset of locales so this is fine
```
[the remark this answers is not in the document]
james-elicx: i dont really see the point in adding a separate function, could just do the type assertion in the same place instead

templates/_worker.js/routes-matcher.ts
james-elicx: sure, i mean, i dont really mind. if you dont like the regex of regex i have no issues changing it to be more like that

tests/templates/requestTestData/i18n.ts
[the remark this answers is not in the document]
dario-piotrowicz: yeah...maybe I was going overboard, I'm fine with either 🙂 (I guess it also won't need to be reused anyways 🤔)

templates/_worker.js/routes-matcher.ts
dario-piotrowicz: I do like regexes so if you strongly prefer the regex of regex I am fine with it, although as I mentioned I feel like we need to add comments and make things clearer as the code itself in its current state seems to me too difficult/inconvenient to grasp.

I'm also suggesting to have this logic into a separate function/method so that the if guard can be a clear oneliner, I think it also helps in making things clearer, but if you dislike that I'm fine of having it all in the same method.

tldr; I just think we should try to make this a bit clearer if possible 🙂

PR #319, opened 2023-06-20
update next.config.js options in supported doc

dario-piotrowicz:


docs/supported.md
```diff
@@ -57,64 +57,81 @@ If you're application is using a package which relies on unsupported Node.js API
...
+| generateBuildId                  | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/generateBuildId), [app](https://nextjs.org/docs/app/api-reference/next-config-js/generateBuildId)                         | ✅                 |
+| generateEtags                    | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/generateEtags), [app](https://nextjs.org/docs/app/api-reference/next-config-js/generateEtags)                             | 🔄                 |
+| headers                          | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/headers), [app](https://nextjs.org/docs/app/api-reference/next-config-js/headers)                                         | ✅                 |
+| httpAgentOptions                 | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/httpAgentOptions), [app](https://nextjs.org/docs/app/api-reference/next-config-js/httpAgentOptions)                       | `N/A`              |
```
dario-piotrowicz, line 82: I couldn't really test this one for some reason, I wasn't able to get the `keepAlive` header with or without the option (in a standard `next build` + `next dev` on a non-edge api route 🤷)

I think this is anyways only applicable for node.js routes, so I think `N/A` here is ok

(if it's not and someone has issues with it hopefully they'll open a github issue about it)

docs/supported.md
```diff
@@ -57,64 +57,81 @@ If you're application is using a package which relies on unsupported Node.js API
...
+| typescript                       | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/typescript), [app](https://nextjs.org/docs/app/api-reference/next-config-js/typescript)                                   | ✅                 |
+| urlImports                       | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/urlImports), [app](https://nextjs.org/docs/app/api-reference/next-config-js/urlImports)                                   | ✅                 |
+| webpack                          | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/webpack), [app](https://nextjs.org/docs/app/api-reference/next-config-js/webpack)                                         | ✅                 |
+| webVitalsAttribution             | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/webVitalsAttribution), [app](https://nextjs.org/docs/app/api-reference/next-config-js/webVitalsAttribution)               | `N/A`              |
```
dario-piotrowicz, line 103: I couldn't get web vitals attributions.... they don't seem too well documented.... I have no idea if they work or not with next-on-pages 😓

dario-piotrowicz: sort of got to the bottom of it, I am pretty sure we support them

packages/next-on-pages/docs/supported.md
```diff
@@ -57,64 +57,81 @@ If you're application is using a package which relies on unsupported Node.js API
...
+| incrementalCacheHandlerPath      | [app](https://nextjs.org/docs/app/api-reference/next-config-js/incrementalCacheHandlerPath)                                                                                                  | 🔄<sup>5</sup>     |
+| mdxRs                            | [app](https://nextjs.org/docs/app/api-reference/next-config-js/mdxRs)                                                                                                                        | ✅                 |
+| onDemandEntries                  | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/onDemandEntries), [app](https://nextjs.org/docs/app/api-reference/next-config-js/onDemandEntries)                         | ❌<sup>6</sup>     |
+| output                           | [pages](https://nextjs.org/docs/pages/api-reference/next-config-js/output), [app](https://nextjs.org/docs/app/api-reference/next-config-js/output)                                           | ❌<sup>7</sup>     |
```
dario-piotrowicz, line 87: - [ ] change to `N/A`

jculvey (approved):

petebacondarwin (approved):
Looks good but I think we could change the red "X" to something like "N/A" in many cases since the footnotes explain that they are not applicable. And that would look less broken in terms of feature support.

petebacondarwin (approved):

dario-piotrowicz:
> > Looks good but I think we could change the red "X" to something like "N/A" in many cases since the footnotes explain that they are not applicable. And that would look less broken in terms of feature support.
> 
> yeah good point, I set `N/A` as an after thought, let me go back and check the existing `X`es and eventually replace some with `N/A` 🙂👍

done 🙂

petebacondarwin (approved):

PR #317, opened 2023-06-16
add missing `await`s for `getPackageVersion` calls

dario-piotrowicz:
### Before
Screenshot 2023-06-16 at 09 46 07

### After
Screenshot 2023-06-16 at 09 46 25

___

Also
resolves #305

PR #325, opened 2023-06-23
improve bundled assets handling

### LINKED ISSUE #39: [🐛 Bug]: `@vercel/og` doesn't work
I'm having issues getting my project to successfully deploy on Cloudflare because I am relying on a dependency (`@vercel/og`) which uses `satori/wasm`.

The Vercel building process goes fine but when `next-on-pages` gets to build the final worker bundle , `esbuild` fails with error:

```bash
✘ [ERROR] Could not resolve "/wasm/wasm_53adb396180a95eb122a0218c93e3d6474b1a27e.wasm"
```

Any idea on how this could get resolved?

dario-piotrowicz:
This PR allows us to handle vercel asset fetches such as the one present in this example: https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation/og-image-examples#using-a-custom-font

Which makes it possible to use `@vercel/og` 🚀 😄 

Result:
[simple-wasm-pages-13.3.1](https://github.com/dario-piotrowicz/next-apps-for-testing/tree/master/apps/simple-wasm-pages-13.3.1) deployed at: https://a9aeda25.next-on-pages-test-5h3.pages.dev

resolves #39

### DIFF
--- a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
+++ b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
@@ -92,7 +92,7 @@ export async function generateFunctionsMap(
 	if (processingResults.bundledAssetsInfo.size) {
 		await copyBundledAssetFiles(
 			nextOnPagesDistDir,
-			processingResults.bundledAssetsInfo
+			processingResults.bundledAssetsInfo,
 		);
 	}
 
@@ -204,7 +204,7 @@ async function processDirectoryRecursively(
 				nextJsManifests.set(key, value),
 			);
 			dirResults.bundledAssetsInfo?.forEach((value, key) =>
-				bundledAssetsInfo.set(key, value)
+				bundledAssetsInfo.set(key, value),
 			);
 		}
 	}
@@ -233,12 +233,12 @@ type FunctionConfig = {
 
 async function processFuncDirectory(
 	setup: ProcessingSetup,
-	directoryFilepath: string
+	directoryFilepath: string,
 ): Promise<Partial<DirectoryProcessingResults>> {
 	const relativePath = relative(setup.functionsDir, directoryFilepath);
 
 	const functionConfig = await readJsonFile<FunctionConfig>(
-		join(directoryFilepath, '.vc-config.json')
+		join(directoryFilepath, '.vc-config.json'),
 	);
 
 	if (functionConfig?.runtime !== 'edge') {
@@ -966,7 +966,7 @@ async function collectChunksConsumersInfosFromFuncDirectory(
  */
 async function copyBundledAssetFiles(
 	distDir: string,
-	bundledAssetsInfo: Map<string, BundledAssetInfo>
+	bundledAssetsInfo: Map<string, BundledAssetInfo>,
 ): Promise<void> {
 	const assetsDir = join(distDir, 'assets');
 	await mkdir(assetsDir);
--- a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
+++ b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
@@ -89,6 +89,13 @@ export async function generateFunctionsMap(
 		);
 	}
 
+	if (processingResults.bundledAssetsInfo.size) {
+		await copyBundledAssetFiles(
+			nextOnPagesDistDir,
+			processingResults.bundledAssetsInfo
+		);
+	}
+
 	return processingResults;
 }
 
@@ -139,6 +146,11 @@ async function tryToFixInvalidFunctions({
 	}
 }
 
+type BundledAssetInfo = {
+	filename: string;
+	originalFileLocation: string;
+};
+
 type WasmModuleInfo = {
 	identifier: string;
 	importPath: string;
@@ -155,6 +167,7 @@ async function processDirectoryRecursively(
 	const prerenderedRoutes = new Map<string, PrerenderedFileData>();
 	const wasmIdentifiers = new Map<string, WasmModuleInfo>();
 	const nextJsManifests = new Map<string, string>();
+	const bundledAssetsInfo = new Map<string, BundledAssetInfo>();
 
 	const files = await readdir(dir);
 	const functionFiles = await fixPrerenderedRoutes(
@@ -190,6 +203,9 @@ async function processDirectoryRecursively(
 			dirResults.nextJsManifests?.forEach((value, key) =>
 				nextJsManifests.set(key, value),
 			);
+			dirResults.bundledAssetsInfo?.forEach((value, key) =>
+				bundledAssetsInfo.set(key, value)
+			);
 		}
 	}
 
@@ -200,6 +216,7 @@ async function processDirectoryRecursively(
 		prerenderedRoutes,
 		wasmIdentifiers,
 		nextJsManifests,
+		bundledAssetsInfo,
 	};
 }
 
@@ -216,12 +233,12 @@ type FunctionConfig = {
 
 async function processFuncDirectory(
 	setup: ProcessingSetup,
-	filepath: string,
+	directoryFilepath: string
 ): Promise<Partial<DirectoryProcessingResults>> {
-	const relativePath = relative(setup.functionsDir, filepath);
+	const relativePath = relative(setup.functionsDir, directoryFilepath);
 
 	const functionConfig = await readJsonFile<FunctionConfig>(
-		join(filepath, '.vc-config.json'),
+		join(directoryFilepath, '.vc-config.json')
 	);
 
 	if (functionConfig?.runtime !== 'edge') {
@@ -239,7 +256,7 @@ async function processFuncDirectory(
 		functionConfig.entrypoint = 'index.js';
 	}
 
-	const functionFilePath = join(filepath, functionConfig.entrypoint);
+	const functionFilePath = join(directoryFilepath, functionConfig.entrypoint);
 	if (!(await validateFile(functionFilePath))) {
 		if (isMiddleware) {
 			// We sometimes encounter an uncompiled `middleware.js` with no compiled `index.js` outside of a base path.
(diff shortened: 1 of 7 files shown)


packages/next-on-pages/templates/_worker.js/index.ts
```diff
@@ -12,25 +15,29 @@ declare const __ENV_ALS_PROMISE__: Promise<null | AsyncLocalStorage<unknown>>;
...
 			return new Response(
 				`Error: Could not access built-in Node.js modules. Please make sure that your Cloudflare Pages project has the 'nodejs_compat' compatibility flag set.`,
 				{ status: 503 }
 			);
 		}
-		return envAsyncLocalStorage.run({ ...env, NODE_ENV: __NODE_ENV__ }, () => {
-			const adjustedRequest = adjustRequestForVercel(request);
+		return envAsyncLocalStorage.run(
+			{ ...env, NODE_ENV: __NODE_ENV__ },
+			async () => {
```
GregBrimble, line 28: Does this now need to be async?

packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
```diff
@@ -15,8 +15,14 @@ export function generateGlobalJs(): string {
 				env: new Proxy(
 					{},
 					{
-						get: (_, property) => Reflect.get(envAsyncLocalStorage.getStore(), property),
-						set: (_, property, value) => Reflect.set(envAsyncLocalStorage.getStore(), property, value),
+						get: (_, property) => {
+							if(!envAsyncLocalStorage.getStore()) return;
```
GregBrimble, line 19: Why is this and L23 necessary?

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -0,0 +1,40 @@
...
+function applyPatch() {
+	const originalFetch = globalThis.fetch;
+	globalThis.fetch = async (...args) => {
+		const request = new Request(...args);
+
+		if (request.url.startsWith('blob:')) {
+			try {
+				const url = new URL(request.url);
+				const binaryContent = (
+					await import(`./__next-on-pages-dist__/assets/${url.pathname}.bin`)
```
GregBrimble, line 25: Does `url.pathname` potentially need to be decoded? Check out what happens if the file has whitespace in it or whatever.

packages/next-on-pages/templates/_worker.js/index.ts
```diff
@@ -12,25 +15,29 @@ declare const __ENV_ALS_PROMISE__: Promise<null | AsyncLocalStorage<unknown>>;
 
 export default {
 	async fetch(request, env, ctx) {
+		patchFetchToAllowBundledAssets();
```
GregBrimble, line 18: Does this need to happen in the fetch handler? Couldn't this happen in the global scope?

packages/next-on-pages/templates/_worker.js/index.ts
dario-piotrowicz: in https://github.com/cloudflare/next-on-pages/pull/317 I've added some eslint Promises related rules, one of those require you to make async functions that return a promise: [promise-function-async](
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/promise-function-async.md)

As I touched this file vscode complained about it so I added the `async`, I think that our eslint config might not be picking up stuff in the `templates` directory, I'll address that in a separate PR

For the added `async` and `promise-function-async` I guess it's kinda nice that we get the extra indication that functions are asynchronous(-like) instead of having to read their content, but I am lukewarm about it, if you prefer I am totally ok in removing these.

packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
dario-piotrowicz: I need to check with the runtime team, I'm somewhat confident that it's a runtime bug

anyways this is why I put the PR in draft I need to dig a sec deeper into this

packages/next-on-pages/templates/_worker.js/index.ts
dario-piotrowicz: yes, I thought doing it here felt cleaner, I've also added a flag in the function so that we don't do multiple patches, so even though this can run multiple times I don't think it really introduces any overhead

but anyways I can just as well move it in the global scope if you prefer, I don't have a strong opinion on it 🙂 

(but I guess I'd leave the flag there anyways just in case? or would it become completely unnecessary in the global scope?)

packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
dario-piotrowicz: I've removed this.... it doesn't seem necessary anymore... 🤷 😓

dario-piotrowicz: I'm not too sure what's going on to be honest, but if this works I think we can just proceed with it this way and keep an eye out for it 🤔

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
dario-piotrowicz: interestingly that doesn't seem the case

at least for example I tried with whitespaces:
https://github.com/dario-piotrowicz/next-apps-for-testing/commit/a75ab62508c61cc5d3b604ca50d2929ccf89473b

and it seems to be working just fine:
https://6e9650dc.next-on-pages-test-5h3.pages.dev/api/og-custom

If I try to encode the url Next itself complains about it:
Screenshot 2023-06-27 at 11 40 03

So I think we should be safe not decoding it 🙂
(and if this turns out to be needed we can always add the decoding later on)

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -922,3 +956,22 @@ async function collectChunksConsumersInfosFromFuncDirectory(
...
+ * @param bundledAssetsInfo map containing all the bundled assets files info collected during the build process
+ */
+async function copyBundledAssetFiles(
+	distDir: string,
+	bundledAssetsInfo: Map<string, BundledAssetInfo>,
+): Promise<void> {
+	const assetsDir = join(distDir, 'assets');
+	await mkdir(assetsDir);
+	for (const { filename, originalFileLocation } of bundledAssetsInfo.values()) {
+		const newLocation = join(assetsDir, `${filename}.bin`);
```
GregBrimble, line 974: This saddens me :(

We need to give frameworks a way to emit a manifest with content-types which Wrangler can pick up cc. @RamIdeas .

packages/next-on-pages/templates/_worker.js/index.ts
GregBrimble: I'd move it. The work we're doing in this isn't dependent on the request, and it is supposed to be mutating the global from the onset. So to me, it makes sense to live in the global space.

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -0,0 +1,40 @@
...
+				).default;
+
+				return {
+					arrayBuffer() {
+						return binaryContent;
+					},
+				} as Response;
```
GregBrimble, line 32: Would be nice to expand this to include `.body`, `.text()`, `.json()`, and `.blob()`. But not a big deal.

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -0,0 +1,40 @@
...
+		const request = new Request(...args);
+
+		if (request.url.startsWith('blob:')) {
+			try {
+				const url = new URL(request.url);
+				const binaryContent = (
+					await import(`./__next-on-pages-dist__/assets/${url.pathname}.bin`)
+				).default;
+
+				return {
```
GregBrimble, line 28:
```suggestion
				// TODO: Replace with a real Response object once available
				return {
```

GregBrimble (approved):
Tentatively approved, pending comments in DM

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
dario-piotrowicz: Added 🙂 

I've tested it with the following: https://github.com/dario-piotrowicz/next-apps-for-testing/commit/d7191257b7cb71f57869fa10cb5b0845a32f7e53
(deployed at https://acb42753.next-on-pages-test-5h3.pages.dev/api/inline-assets/)

This is how it compares to `next dev`:
Screenshot 2023-07-13 at 16 31 29

It works pretty much the same but we do get different Object types logged, since I create them with the correct constructors I assume that this is a runtime inconsistency (either with `toString` or with the object's creation)

dario-piotrowicz: Please let me know what you think 🙂

dario-piotrowicz: PS: yeah I think that the object toString thing is related to the runtime
Screenshot 2023-07-13 at 16 47 01

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
dario-piotrowicz: should be ok as discussed IRL 🙂

GregBrimble (approved):

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -291,6 +308,21 @@ async function processFuncDirectory(
...
+	try {
+		const assetsDir = join(directoryFilepath, 'assets');
+		const files = await readdir(assetsDir);
+		files.forEach(file => {
+			bundledAssetsInfo.set(file, {
+				filename: file,
+				originalFileLocation: join(assetsDir, file),
+			});
+		});
+	} catch {
```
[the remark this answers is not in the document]
dario-piotrowicz: sure 🙂 

actually we do have a function that checks the existence of a directory, I might as well then just use that and not have any try catch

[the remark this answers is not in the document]
dario-piotrowicz: I've removed the try-catch completely 🙂👍

PR #328, opened 2023-06-25
fix: overwrite prerendered files that already exist in output dir

james-elicx:


.changeset/hot-jars-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+Overwrite prerendered files if they already exist in the output directory
```
dario-piotrowicz, line 5: Could you mention that this is so that `--skip-build` can cope with prerendered files? 🙏

packages/next-on-pages/src/buildApplication/fixPrerenderedRoutes.ts
```diff
@@ -109,8 +109,7 @@ async function getRouteDest(
 
 	// Check if a static file already exists at the new location.
 	if (await validateFile(destFile)) {
-		cliWarn(`Prerendered file already exists for ${destRoute}`);
-		return null;
+		cliWarn(`Prerendered file already exists for ${destRoute}, overwriting...`);
```
dario-piotrowicz, line 112: I think this is valid...

it's a bit annoying that with `--skip-build` I can get this sort of incorrectly... but I am not sure if there's anything we can do about this 😓

dario-piotrowicz (approved):
Awesome! thanks! 🤩

.changeset/hot-jars-remain.md
james-elicx, line 5:
```suggestion
Overwrite prerendered files if they already exist in the output directory.

When using `--skip-build`, it would fail if a prerendered file already existed, so we need to override the files so that the build process can continue like normal. This was problematic as after the first build, the prerendered files would now exist in the output directory as static assets, preventing any additional builds.
```

packages/next-on-pages/src/buildApplication/fixPrerenderedRoutes.ts
james-elicx: well the message is technically correct, the prerendered file does already exists in the output directory as a static asset, so it's not the end of the world :p

PR #336, opened 2023-06-26
fix: middleware redirects accidentally applying search params

### LINKED ISSUE #335: [🐛 Bug]: Redirecting in middleware causes query string parameters to be retained

Relevant Packages:
	@cloudflare/next-on-pages: 1.1.0

### Description

If you create a middleware and return a redirect like so:

```
return new Response('', {
        status: 302,
        headers: {
          Location: 'http://localhost:8788/',
        }
      })
```

If the original request URL has query string parameters like `?foo=bar`, the Location header will be `http://localhost:8788/?foo=bar` which is not the intended redirect URL.

Same behaviour is exhibited if using `Response.redirect` or `NextResponse.redirect`.

This seems to be causing the issue: https://github.com/cloudflare/next-on-pages/blob/2c8748176ffed5d50585bd86e14b1f2c974d0c96/packages/next-on-pages/templates/_worker.js/handleRequest.ts#L75

Commenting this out and just returning the unmodified Location header results in the correct Location header being returned.

### Reproduction

_No response_

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Stops applying the search parameters to the `location` header if it is from a middleware.

This appears to be how it behaves in `next dev` for middleware, so I'm happy for this logic to be removed.

fixes #335

### DIFF
--- a/.changeset/large-pillows-try.md
+++ b/.changeset/large-pillows-try.md
@@ -1,5 +1,7 @@
 ---
-"@cloudflare/next-on-pages": patch
+'@cloudflare/next-on-pages': patch
 ---
 
 Prevent middleware redirects applying search params.
+
+When a middleware function results in a redirect, the location header specified in the response is the full destination, including any search params, as written by the developer. Previously, we always applied search params to redirects found during the routing process, no matter what. This meant that we accidentally were applying search params to middleware redirects, which would alter the intended final destination. This change prevents that from happening.
--- a/.changeset/large-pillows-try.md
+++ b/.changeset/large-pillows-try.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+Prevent middleware redirects applying search params.
--- a/packages/next-on-pages/templates/_worker.js/handleRequest.ts
+++ b/packages/next-on-pages/templates/_worker.js/handleRequest.ts
@@ -71,13 +71,15 @@ async function generateResponse(
 	output: VercelBuildOutput
 ): Promise<Response> {
 	// Redirect user to external URL for redirects.
-	if (headers.normal.has('location')) {
-		// Apply the search params to the location header.
-		const location = headers.normal.get('location') ?? '/';
-		const paramsStr = [...searchParams.keys()].length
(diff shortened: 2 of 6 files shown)


packages/next-on-pages/templates/_worker.js/utils/routing.ts
```diff
@@ -6,10 +6,16 @@ import {
 	applySearchParams,
 } from './http';
 
+export type MatchedSetHeaders = {
+	normal: Headers;
+	important: Headers;
+	middlewareLocation?: string | null;
```
dario-piotrowicz, line 12: could you add a comment explaining that this is not standard and why (and how) we need to use it? 🙏

packages/next-on-pages/templates/_worker.js/handleRequest.ts
```diff
@@ -71,13 +71,15 @@ async function generateResponse(
...
-	if (headers.normal.has('location')) {
-		// Apply the search params to the location header.
-		const location = headers.normal.get('location') ?? '/';
-		const paramsStr = [...searchParams.keys()].length
-			? `?${searchParams.toString()}`
-			: '';
-		headers.normal.set('location', `${location}${paramsStr}`);
+	const locationHeader = headers.normal.get('location');
+	if (locationHeader) {
+		// Apply the search params to the location header if it was not from middleware.
```
dario-piotrowicz, line 76: Could you add a comment saying why it should not come from the middleware? 🙏

.changeset/large-pillows-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+Prevent middleware redirects applying search params.
```
dario-piotrowicz, line 5: Could you expand/clarify/show an example in the changeset? 🙏

dario-piotrowicz (approved):
Thanks a lot @james-elicx 😃 

I've just added some comments asking to add more comments/context to the changes as I think that later on it might be quite hard to understand/remember why these changes are here

PR #337, opened 2023-06-26
fix: middleware being invoked more than once in a phase

james-elicx:
This PR does the following:
- Introduces an array to track which middleware has been invoked during each run of a phase.
- If a middleware is tried to be invoked more than once, it skips.
- Adds a test.

In the build output config, there are sometimes two records created for middleware in the `none` phase, both of which can be hit and then accidentally lead to multiple invocations. This PR introduces an array that tracks which middleware paths have been invoked when running a phase. If it tries to use one that has already been invoked, it will skip doing so, fixing the issue of multiple invocations.

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
@@ -40,6 +40,8 @@ export class RoutesMatcher {
 
 	/** Counter for how many times the function to check a phase has been called */
 	public checkPhaseCounter;
+	/** Tracker for the middleware that have been invoked in a phase */
+	private middlewareInvoked: string[];
 	/** Locales found during routing */
 	public locales: Record<string, string> | undefined;
 
@@ -72,6 +74,7 @@ export class RoutesMatcher {
 		applySearchParams(this.searchParams, this.url.searchParams);
 
 		this.checkPhaseCounter = 0;
+		this.middlewareInvoked = [];
 
 		this.wildcardMatch = wildcardConfig?.find(
 			w => w.domain === this.url.hostname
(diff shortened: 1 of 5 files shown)


packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -40,6 +40,8 @@ export class RoutesMatcher {
 
 	/** Counter for how many times the function to check a phase has been called */
 	public checkPhaseCounter;
+	/** Tracker for the middleware that have been invoked in a phase */
+	private middlewareInvoked: string[];
```
dario-piotrowicz, line 44: should/cloud this be a flag instead?

as in
```ts
private middlewareInvoked: boolean;
```

and you set it to `true`/`false` as needed?
do we need to check/care that the `middlewarePath` is the same? shouldn't we just prevent multiple middlewares from running regardless?

dario-piotrowicz: mh... maybe if `middlewarePath` is different we should the different once 🤔

james-elicx: I wanted to do something that would allow multiple middlewares on different paths to still get invoked.

Will there ever be a case where there are multiple with different paths? Not sure, maybe not, but that's why I didn't just do a Boolean.

james-elicx: It was a fix for the same middleware running multiple times, so I feel like considering the path is somewhat important.

dario-piotrowicz: yes I was just asking, all of this feels a bit unclear to me in any case 😓

dario-piotrowicz (approved):
Thanks a lot @james-elicx  🙂

PR #339, opened 2023-06-26
update configs in no-unsupported-configs eslint rule

dario-piotrowicz:
updating the eslint `no-unsupported-configs` rule based on the supported table updated in https://github.com/cloudflare/next-on-pages/pull/319

### DIFF
--- a/packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
+++ b/packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
@@ -6,44 +6,64 @@ import { extractPaths } from '../utils/extract-paths';
 // Note: the rule now only checks for property name, it will probably need to also include case in which we do accept a property but not
 //       certain values for it
 
-/** configs that next-on-pages does support */
-const supportedConfigs: Readonly<Set<string>> = new Set([
-	'env',
-	'basePath',
-	'rewrites',
-	'redirects',
-	'headers',
-	'pageExtensions',
-	'webpack',
-	'generateBuildId',
-	'eslint',
-	'typescript',
-	'experimental/appDir',
-]);
+type Config = { name: string; support: '✅' | 'N/A' | '❌' | '🔄' };
 
-/** configs that next-on-pages does not support and likely never will */
-const indefinitelyUnsupportedConfigs: Readonly<Set<string>> = new Set([
-	'compress',
-	'serverRuntimeConfig',
-	'publicRuntimeConfig',
-	'httpAgentOptions',
-	'distDir',
-	'onDemandEntries',
-	'exportPathMap',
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -6,44 +6,64 @@ import { extractPaths } from '../utils/extract-paths';
...
-	'trailingSlash',
-	'reactStrictMode',
-	'i18n',
-	'experimental/turbo',
-]);
+type Config = { name: string; support: '✅' | 'N/A' | '❌' | '🔄' };
+
+// configs taken from https://github.com/cloudflare/next-on-pages/blob/main/packages/next-on-pages/docs/supported.md#nextconfigjs-properties
+// NOTE: to some we need to add the `experimental/` prefix and the Runtime Config is split into two
+const configs: Config[] = [
```
dario-piotrowicz, line 13: The use of emojis here is a bit unconventional 😅, but I thought of doing this to keep the correlation between this and the [next.config.js props table](https://github.com/cloudflare/next-on-pages/blob/main/packages/next-on-pages/docs/supported.md#nextconfigjs-properties) as clear as possible (so that it's then easier to keep them in sync)

That's also why I changes the creation of the three config sets to all be derived from this list, so that we can have a single list to compare with the docs table (having manually typed support sets made making sure that the rule was in sync with our docs much much much harder).

packages/eslint-plugin-next-on-pages/tests/rules/no-unsupported-configs.test.ts
```diff
@@ -123,7 +123,7 @@ describe('no-unsupported-configs', () => {
     /** @type {import('next').NextConfig} */
     const nextConfig = {
       assetPrefix: 'test',
-      reactStrictMode: true,
```
petebacondarwin, line 126: What happened to `reactStrictMode`?

petebacondarwin (approved):

packages/eslint-plugin-next-on-pages/tests/rules/no-unsupported-configs.test.ts
dario-piotrowicz: This test is for the _currently_ unsupported configs, but I updated `reactStrictMode` to be _indefinitely_ unsupported because I don't think we're likely to ever support it since it's a development mode only feature ([docs](https://nextjs.org/docs/app/api-reference/next-config-js/reactStrictMode)) and we don't suggest the use of `next dev` (but rebuild the whole app altogether each time).

(this might change at some point if we decide to rely on `next dev` for development, but if we keep the current implementation I don't think we'll ever be able to enable this)

PR #340, opened 2023-06-26
fix: entrypoint not being resolved correctly

james-elicx:


dario-piotrowicz (approved):
Thanks for the quick fix! 😃

PR #342, opened 2023-06-27
fix: middleware returns 404 with `new NextResponse` on invalid route

### LINKED ISSUE #341: [🐛 Bug]: new NextResponse() ignoring 2xx status codes, always returning 404

`"@cloudflare/next-on-pages": "0.0.0-f879ffd"`

### Description

Returning a new Response with a status code of 200 will always return status code 404.

E.g:

```
return new NextResponse(
      `Hello, world`,
      {
        status: 200,
        headers: {
          'content-type': 'text/html',
        },
      },
    )
```

If you return a status code of 400 it will work, but a 200 status code does not and will always result in 404.

The bug seems to be caused by this line:
https://github.com/cloudflare/next-on-pages/blob/f879ffd7f6031bdff6420ad53047aa31c3c39f97/packages/next-on-pages/templates/_worker.js/routes-matcher.ts#L564

Removing the `this.status < 400` check fixes the issue, but I assume this is here for other reasons.

### Reproduction

_No response_

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Ensures we bail out of routing when a middleware response is present

Previously, we added a fix to bail out of routing when running a middleware returned a response. However, this change didn't account for when the request was to a route that doesn't exist in the build output. This fix ensures that it always bails out when a user returns `new NextResponse`, regardless of if the target route is in the build output. This is the correct behaviour that should have happened before 🙈

fixes #341

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
@@ -537,11 +537,12 @@ export class RoutesMatcher {
 			}
 		}
 
-		// In the `hit` phase or for external urls/redirects, return the match.
+		// In the `hit` phase or for external urls/redirects/middleware responses, return the match.
 		if (
 			phase === 'hit' ||
 			isUrl(this.path) ||
-			this.headers.normal.has('location')
+			this.headers.normal.has('location') ||
+			!!this.body
 		) {
 			return 'done';
 		}
--- a/.changeset/rude-hotels-raise.md
+++ b/.changeset/rude-hotels-raise.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": patch
+---
+
+Fix middleware returning `new NextResponse` resulting in 404 for routes that don't exist in the build output.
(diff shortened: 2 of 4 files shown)


packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -537,11 +537,12 @@ export class RoutesMatcher {
...
 		}
 
-		// In the `hit` phase or for external urls/redirects, return the match.
+		// In the `hit` phase or for external urls/redirects/middleware responses, return the match.
 		if (
 			phase === 'hit' ||
 			isUrl(this.path) ||
-			this.headers.normal.has('location')
+			this.headers.normal.has('location') ||
+			!!this.body
```
dario-piotrowicz, line 545: so if `this.body` if defined here we always want to bail out, is that correct?

do you think that this can have unintended consequences? (i.e. can we get a truthy `this.body` and not want to bail out?)

james-elicx: That is correct. There should be no unintended consequences. The value for `this.body` is only ever set when a middleware returns a response that is supposed to be returned to the client.

dario-piotrowicz: awesome thanks 🙂

dario-piotrowicz (approved):
Thanks @james-elicx 😄

PR #343, opened 2023-06-27
add fallback detection check in no-unsupported-configs rule

dario-piotrowicz:


packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -225,9 +232,52 @@ function getExportedConfigName(ast: Program): string | null {
...
+	const nodeAfterNextConfigComment = getNodeAfterComment(
+		ast,
+		/@type\s+\{\s*import\((['"])next\1\)\.NextConfig\}/
+	);
+	try {
+		assert(nodeAfterNextConfigComment);
+		assert(nodeAfterNextConfigComment.type === 'VariableDeclaration');
+		assert(nodeAfterNextConfigComment.declarations.length === 1);
+		assert(
+			nodeAfterNextConfigComment.declarations[0]?.id.type === 'Identifier'
```
petebacondarwin, line 246: This seems a strange way to do type narrowing.
Why not just use an if statement?

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -225,9 +232,52 @@ function getExportedConfigName(ast: Program): string | null {
 			/*  empty */
 		}
 	}
+
+	// we get the node right after /** @type {import('next').NextConfig} */
+	const nodeAfterNextConfigComment = getNodeAfterComment(
+		ast,
+		/@type\s+\{\s*import\((['"])next\1\)\.NextConfig\}/
```
petebacondarwin, line 239: Is there not a way to parse the jsdoc comment here and avoid having to use a potentially fragile regular expression?

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -94,7 +94,7 @@ const rule: Rule.RuleModule = {
 	create: context => {
 		const code = context.getSourceCode();
```
petebacondarwin, line 95: ASIDE: according to this doc the `getSourceCode()` function is deprecated: https://eslint.org/docs/latest/extend/custom-rules#:~:text=references%20to%20variables.-,getSourceCode(),-%3A%20(Deprecated%3A

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -225,9 +232,52 @@ function getExportedConfigName(ast: Program): string | null {
 			/*  empty */
 		}
 	}
+
+	// we get the node right after /** @type {import('next').NextConfig} */
+	const nodeAfterNextConfigComment = getNodeAfterComment(
```
petebacondarwin, line 237: Could we flip this and use the eslint API for `getCommentsBefore(nodeOrToken)` on each node in the AST.
That would avoid having to write and maintain our own node after comment helper.

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: Yes thanks a bunch for the reminder, I noticed this too the other day, but doing it seemed like `sourceCode` wasn't a field of `context`... I might need to bump the eslint types package 🤔

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: ah wow, awesome, sorry I didn't think eslint provided something like this out of the box 😓

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: you mean why generally having `assert`s?

I think it's a cleaner solution than `if` statements as those can get quite verbose and difficult to follow, having a series of `assert`s feels like a much more streamlined approach ~(I think I saw it in other libraries as well so I think it's somewhat common when doing AST parsing?)~ (I've spent a bunch of time looking around I am sure I saw something like this in the wild a while back, but I couldn't actually find any, so yeah I suspect that it's pretty non-standard 😓)

If we were to use `if` statements and similar I think we could easily end up in situations in which we get verbose or difficult to understand/tweak code, as we previously had for example:

Screenshot 2023-06-27 at 18 26 11

I think that using `assert`s made my code much cleaner and simple to understand than what could have been...

___

Anyways since it seems not really common and can potentially cause confusion, do you think we should revert to more standard `if`s?

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: good point 🙂👍

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
petebacondarwin: Both approaches seem about the same level of verboseness. What through me was that you are using throwing and catching (but swallowing) errors to do basic logic, which just seems a bit weird. I prefer to use the exception mechanism for real "errors" (i.e. exceptional situations).

E.g. 
image

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: I've refactored the code to use that 🙂👍

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: I've applied the library you suggested 🙂 

However it still required some regex check unfortunately as the type we get can contains spaces and whatnot (that, although absurd don't seem to break the type import)

Screenshot 2023-06-27 at 21 52 48

Anyways the use of the library has reduced the amount of comment we check with the regex so it's definitely a plus 🙂

(PS: I had a quick look, `comment-parser` seems the most popular of such libraries I have seen a few others but haven't test them out, I suspect all of them will have this same behavior, but I will have a double check)

___

- [ ] check other comment parsing libraries to see if all preserve the spacing inside type imports

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: yes I appreciate the mechanism of errors, that's a totally valid point 😓

regarding the verboseness, for example the `&&`s you shared have two significant drawbacks for me:
 - one is that if I need to understand why that `if` passes or not during debugging, by adding `console.log`s or `debuggest` (which can be needed when dealing with complex ASTs) I'll have to break it apart and reconstruct when I'm done with debugging, which makes the code not really convenient to work with
 - the second point is that in this code we don't get it but in other AST checks the left sides of the equations can become very long making the whole thing quite messy, so I think that often intermediate variables can be quite handy there, with a single if as you showed you can't define intermediate variables, so you'd have to have different `if` which can increase the verboseness of the code and make it more difficult to read
 
 <details>
 
 <summary>Example regarding the second point</summary>

Regarding my second point have a look at this function:
https://github.com/cloudflare/next-on-pages/blob/daa32fd41dfb7eea9b5fe7cf337c6cbcf23d951a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L531-L567
 
 I feel like this `assert` pattern there really helps making the code much simpler and clearer, the only alternative I guess would be to have single `if` that return `[]` when not satisfied? 🤔 
 
 </details>
 
 ____
 
Anyways I'm not saying that I am determined to keep this pattern, I'm just saying this to justify why I thought it was pretty nice.
Anyways as I said I got the impression that it was a common pattern 🤦, I'm not a big fan of inventing quirky patterns that can confuse people, so if you're not really in favor of it I think I'm pretty ok just removing it and going back to `if`s and the like.

Although I'd probably prefer to do it in a followup PR generally 🤔

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: Upate:

I've tried a handful of packages, none seemed worth even trying 😓 

@jsdoc/parse - no typings, no documentation 
@types/jsdoc-parse - no typings 
doxx - unmaintained, no active users
jsdoctypeparser - no typings

I think `comment-parser` is as good as it gets 😕

petebacondarwin: Cool. I think this is looking pretty good. We could probably use something like the MS TypeScript API to parse the type expression but that would be overkill.

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
petebacondarwin: Definitely no need to change it for this PR. I can see the pattern is used throughout this code.
I am not entirely against it, I was just surprised by it as it goes against what I normally expect the use of `assert` and `try-catch` to do, which is to assert that we "expect" a particular expression to be true at that point and that if it is false then that is a programming error on our part.

Personally, I would say that if a type narrowing gets too complicated then it should be pulled out into a type checking function (i.e. `function foo(value: any): value is XXX`) that keeps the call site nice and clean (and offers some kind of self documentation). And then the function can be broken into multiple steps quite nicely.  For example:

```ts
VariableDeclaration: node => {
	if (!exportedConfigName) {
		// We could not find an exported config with an identifier so there no variable to check
		return;
	}
	const declaration = node.declarations[0];
	if (isExportedConfig(declaration, exportedConfigName)) {
		const nextConfigProps =
			declaration.init.properties.filter(isProperty);
		checkConfigPropsRecursively(nextConfigProps, context, options);
	}
},

...

function isExportedConfig(
	declaration: VariableDeclarator | undefined,
	exportedConfigName: string
): declaration is VariableDeclarator & { init: ObjectExpression } {
	if (declaration === undefined) {
		return false;
	}
	if (declaration.id.type !== 'Identifier') {
		return false;
	}
	if (declaration.id.name !== exportedConfigName) {
		return false;
	}
	if (declaration.init?.type !== 'ObjectExpression') {
		return false;
	}
	return true;
}

function isProperty(p: Property | SpreadElement): p is Property {
	return p.type === 'Property';
}
```

petebacondarwin (approved):

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: yeah that's a totally valid point, I'll try to refactor this in a followup PR then 🙏

PR #345, opened 2023-06-28
bump eslint packages and remove deprecated calls

dario-piotrowicz:
as part of this also clean up related packages dependencies and add an eslint rule to warn about deprecated calls

___

@petebacondarwin followup for https://github.com/cloudflare/next-on-pages/pull/343#discussion_r1244023677 🙂👍

### DIFF
--- a/internal-packages/eslint-config-next-on-pages/index.js
+++ b/internal-packages/eslint-config-next-on-pages/index.js
@@ -29,7 +29,7 @@ module.exports = {
 		"@typescript-eslint/no-floating-promises": "error",
 		"@typescript-eslint/promise-function-async": "error",
 
-		"deprecation/deprecation": "warn",
+		"deprecation/deprecation": "error",
 	},
 	ignorePatterns: ["vitest.config.ts"]
 };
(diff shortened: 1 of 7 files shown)


internal-packages/eslint-config-next-on-pages/index.js
```diff
@@ -28,6 +28,8 @@ module.exports = {
 		"@typescript-eslint/await-thenable": "error",
 		"@typescript-eslint/no-floating-promises": "error",
 		"@typescript-eslint/promise-function-async": "error",
+
+		"deprecation/deprecation": "warn",
```
dario-piotrowicz, line 32: usually I am for having all rules set to `error` (since `warn`s are a bit pointless and can easily be overlooked/ignored) but since sometimes you might be forced to call deprecated functions so I set this to `warn` (so that we can use this to check our code but it can't make linting fail) 🤔 

I'm totally happy to set it to `error` if preferred 🙂

petebacondarwin: My take is that we should always "try" to use `error` and then opt out on a case by case basis with eslint comments in the code.

petebacondarwin (approved):

internal-packages/eslint-config-next-on-pages/index.js
dario-piotrowicz: ok good point, let me just set this to `error` here then 🙂 (also we don't have currently any deprecated call 🙂👍)

PR #347, opened 2023-06-28
deduplicate Next.js manifests

### LINKED ISSUE #330: [🐛 Bug]: Cannot deploy large applications over 25MB

not relevant. 

### Description

I am currently trying to deploy my application on Cloudflare, but I am facing an issue due to the presence of over 30 routes and each function in my application, after being built, is approximately 1MB in size and the app size is more than 25MB. We  have a Pro/Business account, our request to increase the storage capacity has not received a response yet.
I think Cloudflare should make it easier to deploy large applications and at least there should be a difference between paid accounts and free accounts. I had discussion with staff on Discord and they made a ticket `PAGES-2748`. 

### Reproduction

_No response_

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

dario-piotrowicz:
<details>

<summary>Results on a very small app</summary>

Before
Screenshot 2023-06-28 at 16 37 44

After
Screenshot 2023-06-28 at 16 35 39

This doesn't seem to make a huge difference on small applications but I believe that the duplication of the manifest data in large applications can spin out of control

</details>

<details>

<summary>Results on a medium/small app</summary>

Before
Screenshot 2023-06-28 at 16 53 13

After
Screenshot 2023-06-28 at 16 55 30

Here I think we do start seeing how much this can impact the overall app size 😄 

</details>

<details>

<summary>Results on a large app</summary>

After checking with the creator of #330 it seems like their application went from over 25MB to 9.8MB 🙂

Before
248768424-e4b2dbee-df89-459c-8353-03a8adf4068f

After
image

I was hoping that the size of the app would reduce even further but this is still quite a sizeable improvement 🙂 

</details>

____

resolves #330

### DIFF
--- a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
+++ b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
@@ -435,37 +435,33 @@ type ManifestStatementInfo = {
 function extractManifestStatementInfo(
 	statement: AST.StatementKind
 ): ManifestStatementInfo | null {
-	try {
-		if (
-			statement.type !== 'ExpressionStatement' ||
-			statement.expression.type !== 'AssignmentExpression' ||
-			statement.expression.left.type !== 'MemberExpression' ||
-			statement.expression.left.object.type !== 'Identifier' ||
-			statement.expression.left.object.name !== 'self' ||
-			statement.expression.left.property.type !== 'Identifier'
-		)
-			return null;
-
-		const nextJsManifests = [
-			'__RSC_SERVER_MANIFEST',
-			'__RSC_MANIFEST',
-			'__RSC_CSS_MANIFEST',
-			'__BUILD_MANIFEST',
-			'__REACT_LOADABLE_MANIFEST',
-			'__NEXT_FONT_MANIFEST',
-		];
-		if (!nextJsManifests.includes(statement.expression.left.property.name))
-			return null;
-
-		const { start, end } = statement as unknown as Node;
-		return {
-			manifestIdentifier: statement.expression.left.property.name,
-			start,
-			end,
-		};
-	} catch {
+	if (
+		statement.type !== 'ExpressionStatement' ||
+		statement.expression.type !== 'AssignmentExpression' ||
+		statement.expression.left.type !== 'MemberExpression' ||
+		statement.expression.left.object.type !== 'Identifier' ||
+		statement.expression.left.object.name !== 'self' ||
+		statement.expression.left.property.type !== 'Identifier'
+	)
 		return null;
-	}
+
+	const nextJsManifests = [
+		'__RSC_SERVER_MANIFEST',
+		'__RSC_MANIFEST',
+		'__RSC_CSS_MANIFEST',
+		'__BUILD_MANIFEST',
+		'__REACT_LOADABLE_MANIFEST',
+		'__NEXT_FONT_MANIFEST',
+	];
+	if (!nextJsManifests.includes(statement.expression.left.property.name))
+		return null;
+
+	const { start, end } = statement as unknown as Node;
+	return {
+		manifestIdentifier: statement.expression.left.property.name,
+		start,
+		end,
+	};
 }
 
 /**
--- a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
+++ b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
@@ -149,33 +149,34 @@ async function processDirectoryRecursively(
 		setup.outputDir
 	);
 
-	await Promise.all(
-		functionFiles.map(async file => {
-			const filepath = join(dir, file);
-			if (await validateDir(filepath)) {
-				const dirResultsPromise = file.endsWith('.func')
-					? processFuncDirectory(setup, filepath)
-					: processDirectoryRecursively(setup, filepath);
-				const dirResults = await dirResultsPromise;
-				dirResults.invalidFunctions?.forEach(fn => invalidFunctions.add(fn));
-				dirResults.functionsMap?.forEach((value, key) =>
-					functionsMap.set(key, value)
-				);
-				dirResults.webpackChunks?.forEach((value, key) =>
-					webpackChunks.set(key, value)
-				);
-				dirResults.prerenderedRoutes?.forEach((value, key) =>
-					prerenderedRoutes.set(key, value)
-				);
-				dirResults.wasmIdentifiers?.forEach((value, key) =>
-					wasmIdentifiers.set(key, value)
-				);
-				dirResults.nextJsManifests?.forEach((value, key) =>
-					nextJsManifests.set(key, value)
-				);
-			}
-		})
-	);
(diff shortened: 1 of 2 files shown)


packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -703,3 +809,26 @@ async function copyWasmFiles(
...
+ */
+async function createNextJsManifestFiles(
+	distDir: string,
+	manifests: Map<string, string>
+): Promise<void> {
+	const manifestsDistDir = join(distDir, 'nextjs-manifests');
+	// TODO: the following `mkdir` should not be recursive (since at this point the distDir should always exist),
+	//       the current unit tests force us to make this recursive, when we get proper e2es we need to phase out
+	//       the use of mocking fs in our unit tests and this should hopefully allow us to remove unnecessary
+	//       "recursive" flags in `mkdir` calls such as the following
```
dario-piotrowicz, line 828: I put this as I thought of giving up on removing the `recursive`... I need to give it another shot before throwing the towel 😓

dario-piotrowicz: Update:

The main issue here is the test with the custom dir, it seems like it doesn't create the correct directory, and if I try to add the directory myself via mockFs it gets wiped here:
https://github.com/cloudflare/next-on-pages/blob/68d905ed16c92aa36ad4b376639eebcd969cd267/packages/next-on-pages/src/buildApplication/processVercelOutput.ts#L88-L94

ok so the only possible solution is to get the test to trigger the prod code to generate the correct directory... can we do it... 🤔

dario-piotrowicz: ok it was actually my bad 😅 

it seems like we might not generate the output directory (`__next-on-pages-dist__`) if we only have static/prerendered routes, which is what that test has

so instead I added a check to only create the manifest dir if we've actually found manifests to create (which I thought we'd always do, but on hindsight yeah if everything is static I guess no manifests will be needed 🤔)

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -349,6 +368,92 @@ async function extractAndFixWasmRequires(
...
+
+	const manifestLocations: {
+		identifier: string;
+		start: number;
+		end: number;
+	}[] = program.body.map(getManifestStartEnd).filter(Boolean) as {
+		identifier: string;
+		start: number;
+		end: number;
+	}[];
```
dario-piotrowicz, line 403: TODO: 
- [ ] cleanup

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -349,6 +368,92 @@ async function extractAndFixWasmRequires(
...
+			.replace(/;$/, '');
+		manifests.set(identifier, manifest);
+	});
+
+	await writeFile(functionFilePath, updatedContents);
+
+	return { manifests, updatedContents };
+}
+
+function getManifestStartEnd(
```
dario-piotrowicz, line 428: TODO:

- [ ] update the name of this function
- [ ] add a tsdoc comment

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -144,37 +152,42 @@ async function processDirectoryRecursively(
...
-				dirResults.prerenderedRoutes?.forEach((value, key) =>
-					prerenderedRoutes.set(key, value)
-				);
-				dirResults.wasmIdentifiers?.forEach((value, key) =>
-					wasmIdentifiers.set(key, value)
-				);
-			}
-		})
-	);
+	// Note: the following could be implemented via `await Promise.all` since different directories are independent to each other
```
dario-piotrowicz, line 155: A bit of a weird note to add here, but I am afraid that if I don't do so then at some point in the future someone (including me! 😁) could see this code, think of making it more efficient and swap it to the `await Promise.all` alternative

Doing so would break only large apps so the regression could even go unnoticed for a while and wreak havoc in the meantime 🤔

(PS: maybe we could somehow cover this with e2e tests? but that would require us to have some huge app or to potentially programmatically generate one on the spot? 🤔)

petebacondarwin: No, this is a very good use of a comment!

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -703,3 +823,22 @@ async function copyWasmFiles(
...
+ *
+ * @param distDir the __next-on-pages-dist__ directory's path
+ * @param manifests map mapping next manifest identifiers to the manifest objects
+ */
+async function createNextJsManifestFiles(
+	distDir: string,
+	manifests: Map<string, string>
+): Promise<void> {
+	const manifestsDistDir = join(distDir, 'nextjs-manifests');
+	await mkdir(manifestsDistDir);
```
petebacondarwin, line 839: Do you need to add `{recursive: true}` here to ensure that parent directories are created if missing?

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -349,6 +368,104 @@ async function extractAndFixWasmRequires(
 	return { wasmIdentifiers, updatedContents };
 }
 
+/**
+ * Given the path of a function file and its content update the content so that it doesn't include Next.js manifests but imports them
+ * from the __next-on-pages-dist__/nextjs-manifests directory.
+ *
+ * Note: Such manifests are always the same for all functions so we need to share them across the various function files instead of costly duplicating them in each file.
+ *
+ * As a side effect this function also updates the file with the new content.
```
petebacondarwin, line 377: Why does this function write to the disk while other functions that transform the content do not?
It feels a bit awkward that at the top level it is not obvious when the file on disk gets mutated.

petebacondarwin (approved):

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
dario-piotrowicz: This code only runs after the build process which creates the parent directories.

The parent directories might not be generated (for example if the whole output is static) but if they are not then as a result no manifest is collected and this code doesn't run, so in any case it should be all good.

But do you think we should add `{ recursive: true }` just in case?

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
dario-piotrowicz: I did it here to be consistent with this function:
https://github.com/cloudflare/next-on-pages/blob/c0ab05a460ca430a73495f14a39a6cda5c505e1b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L308

this might actually be wrong/unnecessary in both functions, I'll look into this separately (if that's ok 🙇 🙏 )

PR #349, opened 2023-06-29
Avoid "deduping" webpack chunks used by a single function

dario-piotrowicz:
Result on [app-playground-edge](https://github.com/dario-piotrowicz/next-apps-for-testing/tree/master/apps/app-playground-edge):

<p align="center">

| Before | After |
|--------|--------|
| Screenshot 2023-06-29 at 13 22 01 | Screenshot 2023-06-29 at 13 24 36

</p>

Not a huge saving in size but the the reduction of files needed could be important for large applications (I think it can potentially solve https://github.com/cloudflare/workers-sdk/issues/3391)

### DIFF
--- a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
+++ b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
@@ -42,13 +42,26 @@ export async function generateFunctionsMap(
 		'__next-on-pages-dist__'
 	);
 
-	const processingSetup = {
+	const processingSetup: ProcessingSetup = {
 		functionsDir,
 		distFunctionsDir: join(nextOnPagesDistDir, 'functions'),
 		distWebpackDir: join(nextOnPagesDistDir, 'chunks'),
 		disableChunksDedup,
+		chunksToDedup: new Set(),
 		outputDir,
 	};
+
+	if (!disableChunksDedup) {
+		const chunksConsumersInfos = await collectChunksConsumersInfoRecursively(
+			processingSetup,
+			functionsDir
+		);
+		processingSetup.chunksToDedup = new Set(
+			Array.from(chunksConsumersInfos.entries())
+				.filter(([, chunkConsumers]) => chunkConsumers.size > 1)
+				.map(([chunkNumber]) => chunkNumber)
+		);
+	}
 
 	const processingResults = await processDirectoryRecursively(
 		processingSetup,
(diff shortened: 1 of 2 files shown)


anonymouscatcher:
We tried this prerelease yesterday, functions size and bundle size were ok also we didn't get `JavaScript heap out of memory` error while building but we just tried this prerelease which doesn't work :( 

the prerelease which was working: 
https://prerelease-registry.devprod.cloudflare.dev/next-on-pages/runs/5403622001/npm-package-next-on-pages-347

anonymouscatcher:
ah my bad, you are right...this is another pr,

petebacondarwin (commented):
Let's check the content for duplicate chunks and error if they do not match.
I think we can track the `chunksToDedup` rather than `chunksNotToDedup` to avoid a double negative.
Finally in that case we could use code like this to simplify the tracking:

```
const chunksToDedup = new Set(Array.from(chunksInfos.keys()).filter(
	num => chunksInfos.get(num)!.usedBy.size > 1
))
```

dario-piotrowicz:
> Let's check the content for duplicate chunks and error if they do not match. I think we can track the `chunksToDedup` rather than `chunksNotToDedup` to avoid a double negative. Finally in that case we could use code like this to simplify the tracking:
> 
> ```
> const chunksToDedup = new Set(Array.from(chunksInfos.keys()).filter(
> 	num => chunksInfos.get(num)!.usedBy.size > 1
> ))
> ```

We actually already have a check for that (when we actually proceed with the chunks extraction):
https://github.com/cloudflare/next-on-pages/blob/c0ab05a460ca430a73495f14a39a6cda5c505e1b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L426-L438
so it should be fine 🙂 

(sorry when we chatted earlier this check slipped my mind)

packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -698,3 +713,77 @@ async function copyWasmFiles(
...
+async function collectChunksConsumersInfosFromFuncDirectory(
+	filepath: string
+): Promise<ChunksConsumersInfo> {
+	const chunksConsumersInfos: ChunksConsumersInfo = new Map();
+
+	const functionConfig = await readJsonFile<FunctionConfig>(
+		join(filepath, '.vc-config.json')
+	);
+
+	if (!functionConfig) {
```
james-elicx, line 770:
```suggestion
	if (!functionConfig || functionConfig.runtime !== 'edge') {
```

Fixes an issue causing the CLI to crash due to trying to read a non-existent entrypoint field in nodejs runtime function configs.

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

petebacondarwin: What does it mean in this case (where the runtime is not marked as "edge"?
Is this actually an error case, since we appear to be trying to process code not marked as edge?

james-elicx: > What does it mean in this case (where the runtime is not marked as "edge"? Is this actually an error case, since we appear to be trying to process code not marked as edge?

@petebacondarwin The build output can produce either `edge` or `nodejs` runtime functions, depending on the route. With `nodejs` functions, there are two types; prerender functions and ssr functions. With the first kind, we take the static assets and discard the function, and with the second kind, we log an error during processing the build output. So, whether the runtime not being `edge` is an error case, well, that depends on other factors + checks.

The only type of function code that we use is for `edge` functions, so we should not attempt to process entrypoints for non-edge functions, but returning an error at this point is premature for the reason explained above. Because of where the first part of the deduping logic is added, it occurs before we process all the build output, which means it has no knowledge about which functions we are discarding/ignoring, if we need to error on a function, or what edge functions we've collected. And then the chunk information it collected is used while processing the build output so that we know about the chunks when actually modifying the edge functions.

dario-piotrowicz: > What does it mean in this case (where the runtime is not marked as "edge"? Is this actually an error case, since we appear to be trying to process code not marked as edge?

Yes it is an actual error case because 'edge' is the only runtime we support as you can see here:
https://github.com/cloudflare/next-on-pages/blob/c0ab05a460ca430a73495f14a39a6cda5c505e1b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L201-L205

___

I didn't put the check here as I didn't want this function to have more context than necessary to how things work (and not checking the runtime here would just mean that we might be parsing a handful of extra files in failing builds so I thought it'd be ok)

(and I think didn't test it with a broken app 🤦)

Thanks @james-elicx 😄 , I think it is ok to add that check here

Actually one thing to also do would be to see if we can abstract the files processing done both in [processDirectoryRecursively](https://github.com/cloudflare/next-on-pages/blob/c0ab05a460ca430a73495f14a39a6cda5c505e1b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L128) and in the newly `collectChunksConsumersInfoRecursively` so that we can share the logic and these types of checks.

I'd imagine that can be done separately (it could be very good if we could do it when we have e2es in place as it feels like a bit of a risky refactoring)

petebacondarwin (approved):

PR #350, opened 2023-06-29
Remove asserts from production code

dario-piotrowicz:
@petebacondarwin I'm removing the `try`-`catch` `assert`s pattern I've been following

I've spent a bunch of time trying to think of how to not use `assert`s in a nice way but I've been struggling to find something I'd be happy with.

If I were to define many small/granular utility functions for AST nodes that would likely produce lots of functions which would just add lots of noise and wouldn't simply things too much (they would just make checks more readable).

If I were to define more chunky/specific utility functions they likely just one used once again, potentially adding noise without too much benefit, they would need to be alongside smaller `if`s which (could use or not more granular utility functions).
Chunky utility functions would also not provide great type safety and we'd be forced to do a bunch of casting (since they would only assert/infer the top level types without considering descendant nodes).

At the end I just ended up taking the existing code and replacing `assert`s with `if`s 😓 🤷 

Please have a look and let me know what you think, I'd totally welcome suggestions 🙏

### DIFF
--- a/packages/eslint-plugin-next-on-pages/src/rules/missing-image-loader.ts
+++ b/packages/eslint-plugin-next-on-pages/src/rules/missing-image-loader.ts
@@ -1,6 +1,5 @@
 import type { Rule } from 'eslint';
 import type { Node, Program } from 'estree';
-import assert, { AssertionError } from 'node:assert';
 
 const rule: Rule.RuleModule = {
 	create: context => {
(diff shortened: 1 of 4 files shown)


dario-piotrowicz:
<details>

<summary>Example of what I mean with the top-level types limitation</summary>

Example to clarify what I meant here:
> Chunky utility functions would also not provide great type safety and we'd be forced to do a bunch of casting (since they would only assert/infer the top level types without considering descendant nodes).

Let's take this function:
https://github.com/cloudflare/next-on-pages/blob/68dd9e97393f5499323a620f94ea87a07011d34b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L531-L567

one nice abstraction here would be to take this section:
https://github.com/cloudflare/next-on-pages/blob/68dd9e97393f5499323a620f94ea87a07011d34b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L535-L555
and abstract it in it's own function

the problem is that then here:
https://github.com/cloudflare/next-on-pages/blob/68dd9e97393f5499323a620f94ea87a07011d34b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L557

a the utility type guard/check couldn't help us with making sure that typescript knows that `expr.arguments[0]` is defined and it has at least two `elements` of which the second is an `ObjectExpression`, depending on the input it would at most just let us know that `statement` is an `ExpressionStatement` or that `expr` is a `CallExpression`.

(PS: as far as I can tell AST types aren't generic)

(PS: I'm pretty sure that with complex type definitions we could get the guard to inform typescript of a correct-enough structure that we could then use without needing casting, but that would add lots of complexity and noise, so I seriously doubt that it would be worth it) 
</details>

petebacondarwin (approved):
Love it!

PR #352, opened 2023-06-29
Pretty sure you don't need both push and pull_request triggers

GregBrimble:
Remove the `pull_request` trigger. I'm pretty sure you only need `push`.


GregBrimble:
Eh, pulled it into #351

PR #356, opened 2023-07-02
add docs-scraper

dario-piotrowicz:
Just an idea to maybe help us to keep our docs for `next.config.js` up to date with the Next documentation

This PR adds a scheduled workflow that runs twice a day (or however often we think is best) and checks (via puppeteer scraping) [ours](https://github.com/cloudflare/next-on-pages/blob/main/packages/next-on-pages/docs/supported.md) and Vercel's docs ([pages](https://nextjs.org/docs/pages/api-reference/next-config-js) and[ app](https://nextjs.org/docs/app/api-reference/next-config-js)) and see if there are discrepancies in regards to the `next.config.js` properties.

If there are, it then creates (or updates if it already exist) an issue ([example](https://github.com/dario-piotrowicz/next-on-pages/issues/10)) which tells us what the differences are, reminds us what files we need to update and also provides a template for the configs table that can be used to update the docs.

Please let me know if you think this would be beneficial, if you think it could require too much effort to maintain (if the Vercel folk update often their docs for example :shrug:) and whatever else, if this is not a good idea I'm totally happy to drop it :slightly_smiling_face:

____

Relevant notes:
 - the workflow is set to error/fail if it doesn't find any configs in either the Next docs or ours, that would indicate that there is some issue with the scraping (some network issue, or maybe they/we have changed the target pages), we should fairly simply spot if the workflow fails
 - The Vercel's docs ([pages](https://nextjs.org/docs/pages/api-reference/next-config-js) and[ app](https://nextjs.org/docs/app/api-reference/next-config-js)) do not include all the `next.config.js` options, others are scattered uncategorized across their docs, so I think that the somewhat only option for us is to ignore those are it'd be quite difficult to keep track of them (we could get them from their [ts interface](https://github.com/vercel/next.js/blob/ce69574c24526b5335f10704294023a3a1a0241e/packages/next/src/server/config-shared.ts#L315) but that would make the thing just marginally simpler :thinking:)


.github/workflows/docs-scraping.yml
```diff
@@ -0,0 +1,168 @@
+name: Docs Scraping
+on:
+  push:
+    branches:
+      - docs-scraper
```
dario-piotrowicz, line 5:
```suggestion
```
This is temporary just for testing :slightly_smiling_face:

internal-packages/docs-scraper/scripts/utils/scrapeConfigs.ts
```diff
@@ -0,0 +1,118 @@
...
+
+	const configs = [
+		...commonNextConfigs,
+		...appSpecificNextConfigs,
+		...pagesSpecificNextConfigs,
+	].sort((a, b) => a.configName.localeCompare(b.configName));
+	return configs.filter(({ configName }) => {
+		// for testing purposed let's pretend that trailingSlash and urlImports have been removed from the next documentation
+		return !['trailingSlash', 'urlImports'].includes(configName);
+	});
```
dario-piotrowicz, line 93:
```suggestion
	return configs;
```
This is temporary just for testing :slightly_smiling_face:

internal-packages/docs-scraper/scripts/utils/scrapeConfigs.ts
```diff
@@ -0,0 +1,118 @@
...
+		els =>
+			els.map(el => {
+				// Note: footnotes leave a number at the end of the text, so we remove it here
+				return el.textContent.replace(/\d*$/, '');
+			})
+	);
+	return nextOnPagesDocumentedNextConfigs.filter(config => {
+		// for testing purposed let's pretend that haven't documented appDir and pageExtensions
+		return !['appDir', 'pageExtensions'].includes(config);
+	});
```
dario-piotrowicz, line 47:
```suggestion
	return nextOnPagesDocumentedNextConfigs;
```
This is temporary just for testing :slightly_smiling_face:

.github/workflows/docs-scraping.yml
dario-piotrowicz, line 1: TODO:

- [ ] update the filename

.github/workflows/configs-docs-scraping.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: Update Issue
+        uses: actions/github-script@v6
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          undocumentedNextConfigs: ${{ needs.check_docs.outputs.undocumented_next_configs }}
+          documentedNonNextConfigs: ${{ needs.check_docs.outputs.documented_non_next_configs }}
+          configsTable: ${{ needs.generate-configs-table.outputs.table }}
+        with:
+          script: |
+            const label = 'outdated-config-docs';
```
jculvey, line 101: Is it possible to make this an actual script in the repo that the workflow just invokes instead of inlining it here like this?

jculvey (approved):
Rad idea! I'm a little curious to see how flaky this turns out to be in practice.

dario-piotrowicz:
> Rad idea! I'm a little curious to see how flaky this turns out to be in practice.

Thanks! ❤️ 

Regarding the flakiness, I have run the workflow a good number of times and I don't think I never see it failing to retrieve the info correctly, so hopefully that shouldn't be an issue 🙂

What I'm more curious about is whether Vercel will turn out to update often and or in very breaky ways their doc pages, that could make this quite inconvenient

either way, flakyness or other things, this is pretty isolated from everything else so if it turns out to be problematic/not a good idea I think it should be relatively straightforward to rip it off 🙂

.github/workflows/configs-docs-scraping.yml
dario-piotrowicz: ah, yeah awesome suggestion!

I used this as the action gives you these `github` and `context` globally available variables, but yeah this does look quite heavy handed and probably not the best way to do it

I think that I can take those `github` and `context` object from this: https://github.com/actions/toolkit/tree/main/packages/github
so maybe I can just install the package and call a script from here

I'll give that a shot! thanks a lot 😄

dario-piotrowicz: Updated :smiley:

PR #358, opened 2023-07-02
fix workflow not running on prs

james-elicx:


dario-piotrowicz (approved):
Awesome! thanks! :smiley:

GregBrimble:
This seems to make it run twice for me (which is why I removed it). Is this actually necessary?

james-elicx:
> This seems to make it run twice for me (which is why I removed it). Is this actually necessary?
> 
> image

I would assume that's because you work on a branch in this repo. The `pull_request` trigger is needed for everyone else who isn't working on a branch in this repo - the workflow wasn't actually starting on PRs with just the `push` trigger. You should be fine to remove the `push` trigger if you don't mind it only triggering on PRs and not your branches.

PR #364, opened 2023-07-03
Bump deps across the board

GregBrimble:


.changeset/eighty-donkeys-attack.md
```diff
@@ -0,0 +1,6 @@
+---
+'eslint-plugin-next-on-pages': patch
+'@cloudflare/next-on-pages': patch
+---
+
+chore: Bump dependencies
```
dario-piotrowicz, line 6: does dependency bumping require changesets? 😮

GregBrimble: Yes. Otherwise it won't trigger a release. And it's something that a user might care about because this could conceivably cause a bug in the next release, so this would help them narrow down where the problem was introduced.

dario-piotrowicz: ah ok, thanks 🙂, I'll keep that in mind, I think I haven't included them in the changesets till now 😓 👍

PR #366, opened 2023-07-03
Delete missing image loader rule

dario-piotrowicz:
Deleting the missing-image-loader eslint rule since no longer needed after #357

___

This is actually a sort of breaking change.... but since it's only for the eslint plugin I don't think it should be problematic, right? 🤔


GregBrimble:
If you want to just make the rule a no-op and deprecate it now, that would avoid you making this a breaking change. And then you can just remove it in the next major verison.

.changeset/slimy-lemons-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-next-on-pages': minor
+---
+
+remove eslint missing-image-loader rule (since it is no longer needed)
```
petebacondarwin, line 5: Please can a provide a lot more context here?
Can you say what the rule used to do and why it is no longer needed?
I see that in the template the commented out code has been replaced with:

```
// TODO: implement proper image resizing
```

What is the context here? Would the rule need to be added back if this were to be implemented?

petebacondarwin (approved):
Just a request for more context. Otherwise LGTM

.changeset/slimy-lemons-taste.md
dario-piotrowicz: Sure great points, I'll make this changeset more clear thanks! 🙂

dario-piotrowicz: Regarding the changes to the commented out code, that code was added with the `Image` fix (#357) but it didn't actually represent how an actually implementation is going to look like, so I am just making it more abstract/generic so that later we don't risk getting confused by it.

The eslint rule should never get be added back (if we do implement proper image resizing that's just going to make things work as intended instead of just serving the images from the app's assets as we currently do)

petebacondarwin (approved):
LGTM

PR #370, opened 2023-07-06
temp fix for chunk + manifest deduping in `--skip-build`

james-elicx:
This is a temporary fix and should be resolved ASAP in a refactor of how the logic for processing the output works.

These two fixes are for errors thrown when using the `--skip-build` mode when deduplication has already taken place before.

Chunk deduping error that is fixed:
image

Manifest deduping error that is fixed:
image

### DIFF
--- a/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
+++ b/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
@@ -51,7 +51,10 @@ export async function generateFunctionsMap(
 		outputDir,
 	};
 
-	if (!disableChunksDedup) {
+	// TODO: Remove need for no chunks directory check in refactor
+	const chunksDirExists = await validateDir(processingSetup.distWebpackDir);
+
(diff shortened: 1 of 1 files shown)


packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts
```diff
@@ -51,7 +51,10 @@ export async function generateFunctionsMap(
 		outputDir,
 	};
 
-	if (!disableChunksDedup) {
+	// TODO: Remove need for no chunks directory check in refactor
```
dario-piotrowicz, line 54:
```suggestion
	// TODO: Remove need for no chunks directory check after https://github.com/cloudflare/next-on-pages/issues/371 is done
```

dario-piotrowicz (approved):
Thanks so much! 

yeah these creeped in... I actually practically never use/try the `--no-build` flag 😓 


I've added a couple of suggestions just to make it simpler to track these

PR #372, opened 2023-07-06
fix: build process modifying vercel functions directory

### LINKED ISSUE #371: [📝 TODO]: Remove`.vercel` files modification as part of the building process
### Description

`.vercel` files shouldn't be modified as part of the building process

This happens in a couple of places:
https://github.com/cloudflare/next-on-pages/blob/505be1e668aac2d964b96c85aeb8919844e82dfe/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L401

https://github.com/cloudflare/next-on-pages/blob/505be1e668aac2d964b96c85aeb8919844e82dfe/packages/next-on-pages/src/buildApplication/generateFunctionsMap.ts#L346

This is incorrect as we should keep the vercel output untouched

___

I think I introduced the above, I am not sure if it was just a plain dumb mistake or if the files are re-read multiple times so there was a need for this, in either case we should remove such behavior and make sure that everything works as expected without modifying files inside `.vercel`

### Additional Information

_No response_

### Would you like to help?

james-elicx:
fixes #371 🤬🤬🤬

### DIFF
--- a/.changeset/curvy-boats-promise.md
+++ b/.changeset/curvy-boats-promise.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Prevent the build process from modifying the `.vercel/output/functions` directory.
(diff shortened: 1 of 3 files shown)


dario-piotrowicz (approved):
😃👍

PR #374, opened 2023-07-06
e2es setup

dario-piotrowicz:
Just working on the e2es, I figured I'd open a draft PR to make sure that the CI handles the tests correctly as I work on them 🙂


pages-e2e/features/simpleAppSsrRoutes/navigation.test.ts
```diff
@@ -0,0 +1,72 @@
...
+		await page.locator('a', { hasText: 'to page A' }).click();
+
+		await page
+			.locator('h2', { hasText: 'Server Side Rendered Page A' })
+			.waitFor();
+		expect(
+			await page
+				.locator('h2', { hasText: 'Server Side Rendered Page A' })
+				.isVisible(),
+		).toBe(true);
```
dario-piotrowicz, line 35: This is an ugly hack workaround, we should not need the `.waitFor` call but only the expect (or similar alternatives)

as playwright should do some some auto-waiting and reties
[Screenshot 2023-07-17 at 16 38 47](https://playwright.dev/docs/api/class-locator)
not forcing developers do manually wait for elements to appear in the dom

___

> **Note**
> The worst parse is that `waitFor` throws a generic `TimeoutError`, I think I need to add a utility for this sort of check and there catch that and rethrow it with a more clear message.

dario-piotrowicz: I've improved this with an `assertVisible` utility

___

### Before and after

#### Code
Screenshot 2023-07-18 at 12 21 03

#### Failure message

Screenshot 2023-07-18 at 12 27 27


Screenshot 2023-07-18 at 12 03 57

pages-e2e/features/_utils/copyWorkspaceAssets.ts
```diff
@@ -0,0 +1,21 @@
+/// <reference types="../setup-types.d.ts" />
+
+import { cp, readdir } from 'fs/promises';
+import { join } from 'path';
+
+/**
+ * Copies all the files present in the `./assets` folder into the
+ * WORKSPACE_DIR folder (merging potentially already existing directories)
+ */
+export async function copyWorkspaceAssets(): Promise<void> {
```
dario-piotrowicz, line 10: @GregBrimble I apply this to all my features, it's a more convenient way of copying things over without needing to care about file names etc...

I wonder if this could/should be baked in the framework itself 🤔 
(like we could have the convention that a folder named a certain way present in a feature just gets automatically copied into the fixture)

pages-e2e/fixtures/pages13.0.0/package.json
```diff
@@ -0,0 +1,25 @@
...
+	"version": "0.1.0",
+	"private": true,
+	"scripts": {
+		"dev": "next dev",
+		"build": "next build",
+		"start": "next start",
+		"lint": "next lint"
+	},
+	"dependencies": {
+		"next": "13.4.8",
```
dario-piotrowicz, line 12: I naively thought that create-next-app@13.0.0 would use next 13.0.0, but it's actually always using the latest next version 🤦 

so I need to fix these as we are just testing the latest next versions right now 😓

dario-piotrowicz:
> **Warning**
> CreateNextApp seems to always [only install the latest](https://github.com/vercel/next.js/blob/2b497413087c56dedb3e04f77ea0deebb5d12d0b/packages/create-next-app/templates/index.ts#L193C1-L201C4), it doesn't matter if I use create-next-app@12, it will still install the  latest 13
>
> I missed that and now I have to update apps which I thought were getting created with different Next versions
>
> This also means that I need to rename `app13Latest` and `pages13Latest` to just `appLatest` and `pagesLatest` and that we can only have the on-demand initialization only for the latest stable 😓
> Meaning that we can't have: `app13Latest` and `app14Latest` nor `app13LatestCanary`, anything expect the latest stable will need to be managed differently (scrapped, statically with scheduled updates, etc...)
>

dario-piotrowicz:
> **Note**
> After looking into it I think that the earliest Next version supporting the edge runtime is 12.2.0, so we cannot go lower than that ([resource](https://nextjs.org/blog/next-12-2))

pages-e2e/features/simplePagesSsrRoutes/navigation.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+		await assertVisible('h2', { hasText: 'Server Side Rendered Page B' });
+
+		await page.goBack();
+
+		await assertVisible('h2', { hasText: 'Server Side Rendered Page A' });
+
+		// Note: we don't check the hardNavigationRequests here as this is broken in Next 12.2
+		//       we can either: fix for 12.2 (if worth it), find a way to discern and avoid these
+		//       checks if we are in 12.2 or uncomment the lines below when/if we drop support for 12
```
dario-piotrowicz, line 32: I've looked into it and it's actually caused by the Vercel output not being properly produced, so I think we cannot actually fix this on our end, Next 12.3 fixes it so I am just going to bump the fixture's version

(so I'd imagine that we can say that 12.3 is the oldest version we fully support 🤔)

(PS: 12.2 still worked but here we would get full page reloads when they should not happen)

jculvey (approved):

jculvey:
FYI these don't run for me locally on mac. Happy to debug with you sometime if needed.

PR #377, opened 2023-07-10
fix: image resizing returning weird response

james-elicx:
This PR fixes a weird issue where every other refresh in the browser would not show the image.

The actual fix is making the response mutable and applying the headers to it, instead of creating a new response with the new headers from the resp body. Why? I don't know - I assume something on the response object being weird. Don't ask me why this fixes it, but it does.

Oh, and this PR also includes an unrelated change to fetch relative assets in the build output through the assets fetcher instead of a fetch request.

and the other thing this pr does is also respect the expected cache-control response header behaviour of falling back to the minimumcachettl config value

https://e572d143.image-resizing-test.pages.dev/


dario-piotrowicz (approved):
Awesome!!! 🤩 Thanks! ❤️

PR #380, opened 2023-07-11
fix: query params not appending with the same key

james-elicx:
This PR does the following:
- Changes the `applySearchParams` function to append on unique key-value combos, instead of overriding them.
- Ensures that internal Next.js search params from matches in the build output config always override so routes function properly.

From what I understand, this is common behaviour and is how it should have been implemented originally.

### DIFF
--- a/packages/next-on-pages/tests/templates/utils/http.test.ts
+++ b/packages/next-on-pages/tests/templates/utils/http.test.ts
@@ -101,6 +101,23 @@ describe('applySearchParams', () => {
 		);
 	});
 
+	test('multiple query params with the same key must be unique values', () => {
+		const source = new URL('http://localhost/page?foo=bar&foo=baz&foo=baz');
+		const target = new URL('http://localhost/page?other=value&foo=baz');
+
+		expect([...source.searchParams.entries()].length).toEqual(3);
+		expect([...target.searchParams.entries()].length).toEqual(2);
+
+		applySearchParams(target.searchParams, source.searchParams);
+
+		expect([...source.searchParams.entries()].length).toEqual(3);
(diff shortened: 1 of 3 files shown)


packages/next-on-pages/tests/templates/utils/http.test.ts
```diff
@@ -81,6 +81,61 @@ describe('applySearchParams', () => {
...
+			'http://localhost/page?other=value&foo=baz&foo=test',
+		);
+
+		expect([...source.searchParams.entries()].length).toEqual(1);
+		expect([...target.searchParams.entries()].length).toEqual(3);
+
+		applySearchParams(target.searchParams, source.searchParams);
+
+		expect([...source.searchParams.entries()].length).toEqual(1);
+		expect([...target.searchParams.entries()].length).toEqual(4);
```
dario-piotrowicz, line 97: I'm confused why we need to do this and can't just do `.size`, I did try it locally and `.size` is `undefined`... do you have any idea as to why? is this a weird vitest thing? 😕

dario-piotrowicz: is this related? 🤔  https://github.com/vercel/next.js/issues/49245

dario-piotrowicz (approved):
Looks great! thanks a lot! 😄 

the only thing is those tests that look a bit awkward 🤔

packages/next-on-pages/tests/templates/utils/http.test.ts
dario-piotrowicz: Seems like `size` has only recently been added in V8 so maybe workerd is simply lagging a bit behind 🤷 

I'm totally happy with this version then 🙂

PR #385, opened 2023-07-18
refactor fetch patching and add missing user-agent header setting

dario-piotrowicz:
Example of what this PR fixed:

Fetch to the GitHub REST API:

[source code](https://github.com/dario-piotrowicz/next-apps-for-testing/blob/master/apps/simple-app-dir-13.4.2/app/api/gh-nop/route.ts)

| Before | After |
|--------|--------|
| Screenshot 2023-07-18 at 16 39 54 | Screenshot 2023-07-18 at 16 38 54 | 
| <sup><sub>https://a3caa9a5.next-apps-for-testing.pages.dev/api/gh-nop/</sub></sup>|<sup><sub>https://2bdb5dc5.next-apps-for-testing.pages.dev/api/gh-nop/</sub></sup>|

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
@@ -86,7 +86,7 @@ async function handleInlineAssetRequest(request: Request) {
 
 			// Note: clone is necessary so that body does work
 			resp.clone = (): Response => {
-				return { ...resp };
+				return { ...resp } as Response;
 			};
 
 			return resp;
--- a/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
@@ -86,7 +86,7 @@ async function handleInlineAssetRequest(request: Request) {
 
 			// Note: clone is necessary so that body does work
 			resp.clone = (): Response => {
-				return resp;
+				return { ...resp };
 			};
 
 			return resp;
--- a/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
@@ -65,13 +65,9 @@ async function handleInlineAssetRequest(request: Request) {
 				get body(): ReadableStream<unknown> | null {
 					return new ReadableStream({
 						start(controller) {
-							return pump();
-							function pump() {
-								const b = Buffer.from(binaryContent);
-								controller.enqueue(b);
-								controller.close();
-								return pump();
-							}
+							const b = Buffer.from(binaryContent);
+							controller.enqueue(b);
+							controller.close();
 						},
 					});
 				},
--- a/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
@@ -1,79 +1,121 @@
-export function patchFetchToAllowBundledAssets(): void {
-	const flagSymbol = Symbol.for('next-on-pages bundled assets fetch patch');
+/**
+ * Patches the global fetch in ways necessary for Next.js (/next-on-pages) applications
+ * to work
+ */
+export function patchFetch(): void {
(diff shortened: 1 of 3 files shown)


packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -1,79 +1,121 @@
...
+ *  Note: this is done by the Vercel network, but also in their next dev server as you can see here:
+ * 		https://github.com/vercel/next.js/blob/6705c803021d3bdea7fec20e5d98f6899e49836d/packages/next/src/server/web/sandbox/context.ts#L318-L320)
+ * @param request the request to update
+ */
+function setRequestUserAgentIfNeeded(
+	request: Request<unknown, RequestInitCfProperties>,
+): void {
+	if (!request.headers.has('user-agent')) {
+		request.headers.set(`user-agent`, `Next.js Middleware`);
+	}
```
dario-piotrowicz, line 116: this is the only functional change in this PR, all the rest is just refactoring to make the code nicer

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -1,79 +1,121 @@
...
+				},
+				get body(): ReadableStream<unknown> | null {
+					return new ReadableStream({
+						start(controller) {
+							return pump();
+							function pump() {
+								const b = Buffer.from(binaryContent);
+								controller.enqueue(b);
+								controller.close();
+								return pump();
```
RamIdeas, line 73: I think this will cause a `Maximum call stack size exceeded` error

Since you have already enqueued all chunks and closed the controller, I don't think you need to recursively call pump

dario-piotrowicz: ah ok I did tested it and it is actually happens to be working fine

when it calls `pump` the function runs twice once normally and the second time it must be throwing or something (even though it doesn't get surfaced) when it does `controller.enqueue` with the closed controller:
Screenshot 2023-07-20 at 00 03 19

Screenshot 2023-07-20 at 00 03 48

but yeah the code is wrong (and it doesn't need a pump function 🤦) I've updated it, thanks again for noticing it! ❤️  (it was a stupid copy-paste mistake on my part 😓)

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -1,79 +1,117 @@
...
-					return resp;
-				};
+			// Note: we can't generate a real Response object here because this fetch might be called
+			//       at the top level of a dynamically imported module, and such cases produce the following
+			//       error:
+			//           Some functionality, such as asynchronous I/O, timeouts, and generating random values,
+			//           can only be performed while handling a request
+			//       this is a somewhat known workerd behavior (currently kept for security and performance reasons)
+			//
+			//       if the above issue/constraint were to change we should replace the following with a real Response object
```
RamIdeas, line 60: 👏

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -1,79 +1,117 @@
...
+					return JSON.stringify(b.toString());
+				},
+				async blob() {
+					return new Blob(binaryContent);
+				},
+			} as Response;
 
+			// Note: clone is necessary so that body does work
+			resp.clone = (): Response => {
 				return resp;
```
RamIdeas, line 89: nit: should we at least spread this into a new object so that it does indeed clone?

```suggestion
				return { ...resp };
```

RamIdeas (approved):

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
dario-piotrowicz: sure 🙂 

it's not really going to be really cloning as the developer could still mutate nested fields, etc...

but anyways I think that this should be enough as I don't even expect this to get actually used (I've just added it for the same of completeness as Greg suggested)

PR #386, opened 2023-07-18
fix: middleware redirect not always bailing out of routing

james-elicx:
This PR does the following:
- Ensures we always bail out of routing when a middleware returns a redirect.
- Adds a test.

Previously, it was still possible for the routing to find a match in a later entry in the build output config. In the case of RSC requests, this would lead to overriding the found headers from the middleware because of the override rule on RSC matching entries in the build output config, discarding the redirect header from the middleware. This PR changes it so that when we encounter a redirect in a middleware, we bail out of routing straight away, instead of proceeding through the stages.

This is probably what should have been done originally, to be honest.

This resolves an issue that was raised in the Cloudflare Discord with Clerk.

### DIFF
--- a/.changeset/seven-socks-roll.md
+++ b/.changeset/seven-socks-roll.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix middleware redirects not always bailing out of the routing stages.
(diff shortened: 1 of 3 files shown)


dario-piotrowicz (approved):
Looks good to me, thanks 😄

PR #393, opened 2023-07-21
add note in readme about yarn and pnpm compatibility with v2 build image

### LINKED ISSUE #171: [⚡ Feature]: Document that pnpm is not supported by pages
pnpm is not supported by Cloudflare Pages build system so, if you try deploying an app and it needs to be built by Cloudflare (meaning that it is not a direct upload) then if it uses pnpm the build will fail.

(note: you can use pnpm locally without issue)

There are some simple solutions to this: https://github.com/cloudflare/next-on-pages/issues/52#issuecomment-1480345217
(but ultimately pnpm should get built-in support in Cloudflare Pages at some point, hopefully soon, see https://github.com/cloudflare/pages-build-image/discussions/1)

We should probably document the above somewhere so that developers can see this clearly instead of seeing their build failing and getting confused by it.

dario-piotrowicz:
resolves #171


petebacondarwin (approved):
Let's do it and follow up with further `yarn pnp` testing

PR #394, opened 2023-07-22
fix: `next.config.js` URL rewrites unintentionally redirecting instead of rewriting

james-elicx:
This PR does the following:
- Fixes an issue where full-URL rewrites were actually redirecting instead of rewriting the URL.
- Fixes the relevant test.

Previously, when a path was rewritten to an URL and it didn't set a `location` header, we set one for it. This was incorrect and it should not have done this. Instead, we should have treated that as the new rewritten destination and sent a fetch request to *proxy* the request to the URL. The response from fetching the rewritten URL should then be used as the response to be returned to the client.

Note that I am currently using Jacob (on Pages team)'s site for the tests, but I am happy to change the URL if you wish.

### DIFF
--- a/packages/next-on-pages/tests/templates/requestTestData/configRewritesRedirectsHeaders.ts
+++ b/packages/next-on-pages/tests/templates/requestTestData/configRewritesRedirectsHeaders.ts
@@ -60,7 +60,7 @@ const rawVercelConfig: VercelConfig = {
 		{ handle: 'resource' },
 		{
 			src: '^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))?(?:/)?$',
-			dest: 'https://uuid.rocks/$1',
+			dest: 'https://external-test-url.com/$1',
 			check: true,
 		},
 		{ src: '/.*', status: 404 },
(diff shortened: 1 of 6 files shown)


james-elicx:
Deployed test app:
- `uuid.rocks/plain` -> https://2ee5af0e.nop-config-rewrites-test.pages.dev/plain
- `uuid.rocks/invalid` (404) -> https://2ee5af0e.nop-config-rewrites-test.pages.dev/invalid

packages/next-on-pages/tests/templates/requestTestData/configRewritesRedirectsHeaders.ts
```diff
@@ -60,7 +60,7 @@ const rawVercelConfig: VercelConfig = {
 		{ handle: 'resource' },
 		{
 			src: '^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))?(?:/)?$',
-			dest: 'https://my-old-site.com/$1',
+			dest: 'https://uuid.rocks/$1',
```
dario-piotrowicz, line 63: as you mentioned in the PR's description 😅 ...

I really don't like us relying to some external site for the tests (and also the tests should ideally be able to run locally without an internet connection I think)

so I'd prefer it if we could mock the fetch instead of it actually happening, something along these lines:
Screenshot 2023-07-24 at 12 35 14

What do you think? does this look ok to you? 🙂

james-elicx: i thought you might have been a bit unhappy about mocking fetch ngl, but if you are happy with it i am happy to do that

dario-piotrowicz (approved):
Awesome! thanks a lot @james-elicx 😄 🎉 

(PS: I'm surprised we didn't notice this earlier 😮)

PR #397, opened 2023-07-24
fix: `/_next/static` in subdir not excluded from routing

### LINKED ISSUE #396: [🐛 Bug]: With Edge Runtime, missing a resource when deployed to Cloudflare Pages or running preview locally

```zsh
System:
	Platform: darwin
	Arch: arm64
	Version: Darwin Kernel Version 22.3.0: Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000
	CPU: (8) arm64 Apple M1 Pro
	Memory: 16 GB
	Shell: /bin/zsh
Binaries:
	Node: 18.16.0
	Yarn: 1.22.19
	npm: 9.5.1
	pnpm: N/A
Relevant Packages:
	@cloudflare/next-on-pages: 1.3.0
	vercel: N/A
	next: N/A
```

### Description

We are experimenting with Edge Runtime for our blog and we noticed the following 404 responses when we have the necessary Edge Runtime changes deployed and when we run a preview (`wrangler pages dev`).

```
Failed to load resource: the server responded with a status of 404 () https://748d706d.reproducible-missing-file.pages.dev/blog/_next/static/chunks/pages/%5Blanguage%5D-95ff3c047aad677e.js

Refused to execute script from https://748d706d.reproducible-missing-file.pages.dev/blog/_next/static/chunks/pages/%5Blanguage%5D-95ff3c047aad677e.js because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```

The 404 responses above comes from the repo for reproducing this error. It is the similar 404 responses we are encountering in our blog.

Our blog uses components from internal AirAsia packages, and none of the JS and CSS from these packages are being loaded. I'm thinking the file that has the necessary JS and CSS aren't being loaded. They do exist when running locally via `yarn dev` and `yarn start`.

I can confirm that the **decoded** version of URL above, https://748d706d.reproducible-missing-file.pages.dev/blog/_next/static/chunks/pages/[language]-95ff3c047aad677e.js, does exist. For whatever reason, the **encoded** version does not exist.

I can confirm that the **encoded** version does exist when running locally via `yarn dev` and `yarn start`.

### Reproduction

Instructions are in the README.md file.

https://github.com/thealiilman/reproducible-missing-file

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

Here's a preview deployment of no errors logged into DevTools console:
https://6ff41122.reproducible-missing-file.pages.dev/whatever/

Here's a preview deployment of errors logged into DevTools console:
https://748d706d.reproducible-missing-file.pages.dev/blog/whatever/

### Would you like to help?

james-elicx:
This PR does the following:
- Grabs the `{...}/_next/static` path from the static assets instead of always using `/_next/static` for the `_routes.json` file.

This allows the `/_next/static` assets to be excluded from worker invocations when in a subdirectory, instead of always assuming they are at the root.

Fixes #396

### DIFF
--- a/packages/next-on-pages/src/buildApplication/buildApplication.ts
+++ b/packages/next-on-pages/src/buildApplication/buildApplication.ts
@@ -120,6 +120,8 @@ async function prepareAndBuildWorker(
 		!disableWorkerMinification,
 	);
 
+	await buildMetadataFiles(outputDir, { staticAssets });
+
 	printBuildSummary(staticAssets, processedVercelOutput, processedFunctions);
 	await writeBuildInfo(
 		join(outputDir, '_worker.js'),
@@ -129,6 +131,4 @@ async function prepareAndBuildWorker(
 	);
 
 	cliSuccess(`Generated '${outputtedWorkerPath}'.`);
-
-	await buildMetadataFiles(outputDir, { staticAssets });
 }
--- a/packages/next-on-pages/src/buildApplication/buildApplication.ts
+++ b/packages/next-on-pages/src/buildApplication/buildApplication.ts
@@ -63,7 +63,6 @@ export async function buildApplication({
 		disableChunksDedup,
 		disableWorkerMinification,
 	});
-	await buildMetadataFiles(outputDir);
 
 	const totalBuildTime = ((Date.now() - buildStartTime) / 1000).toFixed(2);
 	cliLog(`Build completed in ${totalBuildTime.toLocaleString()}s`);
@@ -130,4 +129,6 @@ async function prepareAndBuildWorker(
 	);
 
 	cliSuccess(`Generated '${outputtedWorkerPath}'.`);
+
+	await buildMetadataFiles(outputDir, { staticAssets });
 }
--- a/packages/next-on-pages/src/buildApplication/buildMetadataFiles.ts
+++ b/packages/next-on-pages/src/buildApplication/buildMetadataFiles.ts
@@ -44,14 +44,8 @@ ${Object.entries(nextStaticHeaders)
 	}
 }
 
-async function buildRoutes(
-	outputDir: string,
-	{ staticAssets }: BuildMetadataFilesOpts,
-) {
-	const nextStaticRegex = /^(.*\/_next\/static)\/.+$/;
-	const nextStaticAsset = staticAssets.find(a => nextStaticRegex.test(a));
-	const nextStaticPath =
-		nextStaticAsset?.match(nextStaticRegex)?.[1] ?? '/_next/static';
+async function buildRoutes(outputDir: string, opts: BuildMetadataFilesOpts) {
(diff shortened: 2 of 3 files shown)


packages/next-on-pages/src/buildApplication/buildApplication.ts
```diff
@@ -130,4 +129,6 @@ async function prepareAndBuildWorker(
 	);
 
 	cliSuccess(`Generated '${outputtedWorkerPath}'.`);
+
+	await buildMetadataFiles(outputDir, { staticAssets });
```
dario-piotrowicz, line 133: Is there a specific reason why you added this here? 
could you move this right below line 88? (`const staticAssets = await getVercelStaticAssets();`)

(or in any case I don't think this should be below the successful generation message)

packages/next-on-pages/src/buildApplication/buildMetadataFiles.ts
```diff
@@ -41,15 +44,23 @@ ${Object.entries(nextStaticHeaders)
...
 
-async function buildRoutes(outputDir: string) {
+async function buildRoutes(
+	outputDir: string,
+	{ staticAssets }: BuildMetadataFilesOpts,
+) {
+	const nextStaticRegex = /^(.*\/_next\/static)\/.+$/;
+	const nextStaticAsset = staticAssets.find(a => nextStaticRegex.test(a));
+	const nextStaticPath =
+		nextStaticAsset?.match(nextStaticRegex)?.[1] ?? '/_next/static';
```
dario-piotrowicz, line 54: Nit:
This is sort of clear to me but I feel like we could make the code more easy to read if we extracted this in a properly commented function, something like:
Screenshot 2023-07-24 at 13 32 58

But if you dislike extracting it this way (maybe it is unnecessary/unneeded?) I am also fine keeping the code as it, I leave it up to you 🙂

dario-piotrowicz (approved):
Awesome stuff! thanks a lot @james-elicx!

(such a quick turnaround on the issue!!! 🤩)

I'm approving but I'd really like it if we could move the `buildMetadataFiles` call before merging if possible 🙂 🙏

packages/next-on-pages/src/buildApplication/buildApplication.ts
james-elicx: > Is there a specific reason why you added this here?

consistenty with how it worked before i moved it into this function. it was previously after everything

james-elicx: i moved it, but not quite to the place you suggested

dario-piotrowicz: ok that works for me, thanks 😄

PR #398, opened 2023-07-24
feat: group identifiers by their consumers

james-elicx:
This PR does the following:
- Groups code block identifiers by their consumers.

This drastically reduces the number of unique files created for things like Webpack chunks by grouping them into the same file when they are consumed by the same function, which was causing issues in certain projects.

Fixes [#3391](https://github.com/cloudflare/workers-sdk/issues/3391) (right?)

### DIFF
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
@@ -46,10 +46,7 @@ export function collectIdentifiers(
  * @param identifierMaps The maps of found identifiers in the AST.
  */
 export function groupIdentifiers(identifierMaps: IdentifierMaps) {
-	// Webpack Chunks
-	groupIdentifiersByType(identifierMaps.webpack);
-	// Manifests
-	groupIdentifiersByType(identifierMaps.manifest);
+	Object.values(identifierMaps).forEach(map => groupIdentifiersByType(map));
 }
 
 type IdentifierMaps = Record<IdentifierType, IdentifiersMap>;
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
@@ -2,30 +2,31 @@ import type { Node } from 'acorn';
 import type * as AST from 'ast-types/gen/kinds';
 import type { ProcessVercelFunctionsOpts } from '.';
 import { cliError } from '../../cli';
+import { createHash } from 'crypto';
 
 /**
  * Collects the identifiers from the AST and adds them to the provided maps and lists.
  *
  * @param collectedIdentifiers The collected identifiers.
- * @param program The AST for the function file.
+ * @param programInfo The AST for the function file and its entrypoint.
  * @param opts The options for the function processing.
  */
 export function collectIdentifiers(
 	{ identifierMaps, programIdentifiers }: CollectIdentifiersOpts,
-	program: AST.ProgramKind,
+	programInfo: ProgramInfo,
 	{ disableChunksDedup }: Partial<ProcessVercelFunctionsOpts>,
 ): void {
 	// Wasm
 	collectIdentifiersByType(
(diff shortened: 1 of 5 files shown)


dario-piotrowicz:
> Fixes https://github.com/cloudflare/workers-sdk/issues/3391 (right?)

yes... hopefully 😅

packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
```diff
@@ -2,58 +2,72 @@ import type { Node } from 'acorn';
...
+/**
+ * Groups the identifiers in the provided maps by their consumers.
+ *
+ * @param identifierMaps The maps of found identifiers in the AST.
+ */
+export function groupIdentifiers(identifierMaps: IdentifierMaps) {
+	// Webpack Chunks
+	groupIdentifiersByType(identifierMaps.webpack);
+	// Manifests
+	groupIdentifiersByType(identifierMaps.manifest);
```
dario-piotrowicz, line 52: are you missing wasm? or are you intentionally skipping it?

if it needs to be included, what about doing:
```suggestion
	Object.values(identifierMaps).forEach(map => groupIdentifiersByType(map));
```

james-elicx: Wasm is just imported wasm files that are moved, so there is no point to include it here since grouping is used for code blocks.

dario-piotrowicz: sorry I am not sure I get it, are you saying that wasm files don't get grouped?

if not, it might not be too beneficial but I don't see a negative in treating them as the other type of identifiers and grouping them, no?

james-elicx: Well, it's just that it's completely pointless extra work to do. Only code blocks get grouped and written to grouped files. Including Wasm in the grouping logic is just making it do extra work at build time for no gain.

Wasm deduping is deduping the assets that are imported, not deduping code blocks.

dario-piotrowicz: yeah.. as I said that grouping wasms probably wouldn't be very beneficial, but also the extra work would be negligible, wouldn't it?

Feels to me that it would make sense to treat all identifier types consistently, slightly clearer code, very small benefit in the output, very small extra work done at build time

But it's not like it's going to make a huge difference so if you're totally against it I'm fine keeping it this way, but at least I would imagine that we should document somewhere that wasms are not grouped (to avoid possible future confusions) (as far as I can tell that is not clearly indicated anywhere in the code, is it?)

james-elicx: i will concede in this instance when it comes to that 99% of cases will have no wasm identifiers so there won't really be anything running

dario-piotrowicz:
| Before | After |
|--------|--------|
| Screenshot 2023-07-25 at 11 06 22 | Screenshot 2023-07-25 at 11 03 44 |

🚀🎉🤯 🎉🚀

packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
```diff
@@ -2,58 +2,69 @@ import type { Node } from 'acorn';
...
 	}
 }
 
+/**
+ * Groups the identifiers in the provided maps by their consumers.
+ *
+ * @param identifierMaps The maps of found identifiers in the AST.
+ */
+export function groupIdentifiers(identifierMaps: IdentifierMaps) {
+	Object.values(identifierMaps).forEach(map => groupIdentifiersByType(map));
```
dario-piotrowicz, line 49: sorry after reviewing better the code I think you had it right the first time around 😓
```suggestion
	// Webpack Chunks
	groupIdentifiersByType(identifierMaps.webpack);
	// Manifests
	groupIdentifiersByType(identifierMaps.manifest);
```

dario-piotrowicz (approved):
super duper awesome stuff! 🤩 ❤️ 🚀

PR #403, opened 2023-07-25
add back yarn pnp check with different error message

dario-piotrowicz:
This PR is reverting the change removal made in https://github.com/cloudflare/next-on-pages/pull/393 but also updating the error message showing that the issue lies with the vercel cli

> **Note**
> I haven't added a changeset as the change in 393 is not yet released, I'll tweak accordingly the changelog there before release


petebacondarwin (approved):

dario-piotrowicz:
> > I haven't added a changeset as the change in 393 is not yet released, I'll tweak accordingly the changelog there before release
> 
> In this case why not just revert the changeset too?

it just didn't occur to me to do that here instead of doing it in the release PR 🙈 

I've update the changeset, thanks! 😃

PR #408, opened 2023-07-28
chore: avoid running scraping workflow job in forks

james-elicx:


dario-piotrowicz (approved):
Thanks! sorry I missed this 😃 ❤️

PR #410, opened 2023-07-28
fix: specify framework in the generated project.json

james-elicx:


.changeset/two-bugs-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Specify the `framework` is `nextjs` in the Vercel `project.json` file.
```
dario-piotrowicz, line 5: does this need a changeset? can user get effected by this in any way?

dario-piotrowicz (commented):
what is this fixing? I think you told me the creation of extra middlewares? is it addressing this? https://github.com/cloudflare/next-on-pages/blob/c6621c684b43b7428bd8a348dd8e1322da1d466f/packages/next-on-pages/src/buildApplication/processVercelFunctions/edgeFunctions.ts#L96-L116

if that's the case, should this code also be updated by removing our extra handling of the uncompiled middleware?

james-elicx:
> what is this fixing? I think you told me the creation of extra middlewares? is it addressing this?
> 
> https://github.com/cloudflare/next-on-pages/blob/c6621c684b43b7428bd8a348dd8e1322da1d466f/packages/next-on-pages/src/buildApplication/processVercelFunctions/edgeFunctions.ts#L96-L116
> 
> if that's the case, should this code also be updated by removing our extra handling of the uncompiled middleware?

Yes, it should be fixing the need for that part, but I don't know if I want to remove that or leave it in still just to be on the safe side.

.changeset/two-bugs-sing.md
james-elicx: It's altering how the Vercel CLI might interpret a project, so I think a small changeset is appropriate. Previously the Vercel CLI might also detect the Node.js builder for middleware functions because it didn't know for sure that it was a Next.js project, so this alters its behaviour in a tiny way by telling it that it's Next.js

dario-piotrowicz:
> Yes, it should be fixing the need for that part, but I don't know if I want to remove that or leave it in still just to be on the safe side.

I think I'd remove it not to keep unnecessary stuff around (and if it were to cause issues we'd notice that fairly easily I'd imagine?), but if you want to keep it I'm fine with that as well 🙂

but if you want to keep it I think you should update the comment by removing the `// TODO: investigate when and where the file is generated. ` and potentially saying that it's likely resolved but kept around just in case

.changeset/two-bugs-sing.md
dario-piotrowicz: mh... I'm a bit hesitant by the fact that we do this under the hood, so it's not really a user facing thing? 🤔 

although not 100% sure about it I'm fine with it 🙂👍

dario-piotrowicz (approved):
looks good to me 😄

PR #411, opened 2023-07-28
add middleware e2e features for app and pages

dario-piotrowicz:
test middelwares via the appMiddelware and pagesMiddleware pages e2e features, as part of this change remove the no longer necessary middleware unit tests


pages-e2e/features/appMiddleware/middleware.test.ts
james-elicx, line 1: Please can you add tests back for a few of the edge cases that we implemented fixes for previously?

- Redirects when using the request header `rsc: 1`. It tests that redirects from middleware work when a later entry in the build output config is set to override the previous match.
- Returning `new NextResponse` with a status code >= 400.
- Returning `new NextResponse` on a request to a route that doesn't exist.

dario-piotrowicz: > Please can you add tests back for a few of the edge cases/mishaps that we implemented fixes for previously?

Sure, thanks for pointing them out 😄

> Returning redirects when using the request header rsc: 1. It tests that redirects from middleware work when a later entry in the build output config is set to override the previous match.

👍 

> Returning new NextResponse with an error status code (>= 400).

Sure 🙂👍 

> Returning new NextResponse on a request to a route that doesn't exist.

Sure, I didn't do that as I've already added a test for returning a new NextResponse and doing that for a non-existent route doesn't seem different enough, but I don't mind adding a test for that just in case 👍 


> Another that would be nice to add back but I'm not sure how feasible it would be for you to add is checking that middleware only gets invoked once when there are multiple entries in the build output config for the middleware. You can see what I mean when I say multiple entries in the hardcoded config used previously.

If possible I would very much rather not add anything that needs to be aware of the build output config (as hardcoding it is something I'm very not in favour of), the best thing I think would be to test that middlewares are called only once, especially in cases which previously caused them to be incorrectly called twice. I can't see any clear reproduction [here](https://github.com/cloudflare/next-on-pages/discussions/329) do you have some code that you used to fix the issue that caused the doubled calls? (if not I'll try to reproduce it with an older version of next-on-pages)
would this work for you?
 
___

PS: I've converted the PR to draft to address the above points

james-elicx: > If possible I would very much rather not add anything that needs to be aware of the build output config (as hardcoding it is something I'm very not in favour of), the best thing I think would be to test that middlewares are called only once, especially in cases which previously caused them to be incorrectly called twice. I can't see any clear reproduction https://github.com/cloudflare/next-on-pages/discussions/329 do you have some code that you used to fix the issue that caused the doubled calls? (if not I'll try to reproduce it with an older version of next-on-pages)
would this work for you?

I'm not quite sure what causes the build output config to include two entries for middleware. It's a bit unfortunate really, but I don't think there's a consistent way to reproduce it without hardcoding :/

dario-piotrowicz: > I'm not quite sure what causes the build output config to include two entries for middleware. It's a bit unfortunate really, but I don't think there's a consistent way to reproduce it without hardcoding :/

if that's the case I would just add tests to check that standard middlewares are called only once, hopefully it should be good enough and covert the vast majority of apps, if in the future something happens and the bug surfaces again we'll make sure to fix it and try to reproduce it and add proper testing for it

this is a bit of an optimistic approach but I think that:
 - the bug has been fixed so it's not an issue likely to resurface soon
 - even if it resurfaces it's not an app breaking bug and it's not the end of the world if it happens again
    (it shouldn't really cause any damage unless people perform DB operations or the like in their middlewares)
    
would this approach be ok with you?

james-elicx: Yeah, I think that'll be okay. Well, I'm not sure what else we could do really so yeah.

dario-piotrowicz: ### To do:

- [ ] add tests for: Returning redirects when using the request header rsc: 1. It tests that redirects from middleware work when a later entry in the build output config is set to override the previous match.
- [x] add tests for: Returning new NextResponse with an error status code (>= 400).
- [x] add tests for: Returning new NextResponse on a request to a route that doesn't exist.
___
- [ ] add tests for checking that middlewares are called only once

dario-piotrowicz:
### To do:
- [ ] test NextResponse.json
- [ ] test response headers modification

pages-e2e/features/appMiddleware/middleware.test.ts
dario-piotrowicz: Agreed with @james-elicx that the first todo can be changed to:
 - [ ] add tests to make sure middleware operations work as intended during client-side navigation (both in app dir and pages)

dario-piotrowicz:
Note: in the client side navigation tests I could not test for the absence of hard navigation 😓

I've done that in the ssr features by checking the requests that get sent to the app and make sure that those don't indicate a full page reload, but in this case I couldn't find anything that I could use to discern hard navigation requests from non-hard navigation requests

I've been trying to do other stuff to test that no hard navigation occurs like adding things to the window object (via [evaluate](https://playwright.dev/docs/api/class-frame#frame-evaluate)) and checking that they are still present when no hard-navigation occurs or trying to get [frames](https://playwright.dev/docs/api/class-frame) and see if they get detached on hard navigation, but nothing seems to work...

I've tried googling it but I couldn't find a clear way in playwright to check that no hard navigation occurs 😕, if someone has some ideas on how to do that please let me know 🙂

pages-e2e/features/appMiddleware/middleware.test.ts
```diff
@@ -166,4 +166,19 @@ describe('App Middleware', () => {
 			text: 'json response from middleware',
 		});
 	});
+
+	test('middleware to be invoked once and only once', async ({ expect }) => {
```
petebacondarwin, line 167: per request?

dario-piotrowicz: yeah, I can see the ambiguity, sorry I'll add that 🙇

pages-e2e/fixtures/pages12.3.0/next.config.js
```diff
@@ -5,4 +5,7 @@ module.exports = {
 		runtime: 'experimental-edge',
 	},
 	swcMinify: false,
+	env: {
+		frameworkVersion: '12',
```
petebacondarwin, line 9: Should [pages13.0.0/next.config.js](https://github.com/cloudflare/next-on-pages/blob/9dbefa1b9aaf0b86a11d3a166ae7868f35dee727/pages-e2e/fixtures/pages13.0.0/next.config.js) have an equivalent framework version?

petebacondarwin (approved):
I didn't get a chance to go through all the test cases in detail, but looks good to me.

pages-e2e/fixtures/pages12.3.0/next.config.js
dario-piotrowicz: this is sort of a hack, just because Next 12 doesn't seem to support some things (and/or we fail to support them, but it's not likely worth it to spend too much time trying to... I think?)

so I use this to change outputs slightly in the features so that I know when not to test stuff, truthfully more than a version value this is a flag, like `isNext12: boolean` (but I kept it generic as that might come in handy? 🤔)

adding it to `pages13.0.0` wouldn't change anything (since `process.env.frameworkVersion === '12'` would work identically) but I wouldn't mind adding it to all our fixtures just to make keep things consistent 🙂

dario-piotrowicz: I've added the `frameworkVersion` to all the fixture's config files 🙂

This doesn't include the ones we generate on the fly, for those I'd need to implement codemod logic to update the generated configs, it's not too too complex but not too trivial either so I figured I could skip that for now (since it wouldn't even have any effect) and do that if/when actually needed, hopefully that's ok with you 🙂

pages-e2e/features/appMiddleware/middleware.test.ts
dario-piotrowicz: text improved 🙂👍

PR #417, opened 2023-08-04
feat: support choosen of vercel version

Infrared96:


dario-piotrowicz:
Hi @Infrared96, thanks a lot for the PR! ❤️ 

I am not 100% sure about the change, I'm not saying it's wrong or that I'm against it, I'm just trying to understand better its need and usage 🙇

I can see it making sense if you just want to use `npx @cloudflare/next-on-pages` to build your application without installing next-or-pages (nor vercel), is that the use case you're trying to address?

The fact here is that:
 - such a workflow is not the recommended one (although we do support it)
 - if your project really really needs a specific Vercel CLI version, couldn't it be simply be added as a dev dependency?

If your use case instead isn't one in which you don't install the packages, is there a real need for this? are you having specific issues? (one thing I've noticed is that it seems like with `yarn (classic)` we always do install vercel, maybe that's been problematic for you?)

one thing that concerns me is that your flag would also override any potential vercel dev dependency, which looks to me like this could potentially add ambiguity, that's why I am not too sure about it (but if there is a compelling reason for its addition the slight ambiguity could be worth it)

In summary could you clarify why this change is needed/what use cases it addresses? 🙏

AlexandrNetsvetov:
> 

I use yarn v1.22.17.
The vercel v31.1.1 has been changed on v31.2.1

I'm not sure that I'll be able to fix this during this week. So, if you will have enough time to have a look I will be grateful!

dario-piotrowicz:
> he vercel v31.1.1 has been changed on v31.2.1

I see, not to worry I will have a look 🙂👍

Thanks for the info

dario-piotrowicz:
@AlexandrNetsvetov I've opened #422 which fixes vercel getting re-installed if already present, could you please try the [prerelease](https://github.com/cloudflare/next-on-pages/pull/422#issuecomment-1669950815) and let me know if that works for you? 🙂 🙏

AlexandrNetsvetov:
@dario-piotrowicz Hi, thanks! Now it works 🙂👍

dario-piotrowicz:
@AlexandrNetsvetov thanks for confirming! 😄 

if it's ok with you then we can close this PR and proceed with #422? 🙂

AlexandrNetsvetov:
> if it's ok with you then we can close this PR and proceed with #422? 🙂

@dario-piotrowicz Sure!

dario-piotrowicz:
Closing in favour of #422 as agreed with @AlexandrNetsvetov 🙂

PR #419, opened 2023-08-07
feat: fetch (suspense) cache handling, and `next/cache` support

### LINKED ISSUE #292: [⚡ Feature]: Why does Cloudflare Pages still not support OnDemandRevalidate, even after Next.js 13 introduced a new revalidate function that runs on the edge?
### Description

Next.js 13 introduced revalidatePath, a feature that will run on the edge and revalidate the page on demand.

```typescript
import { NextRequest, NextResponse } from 'next/server';
import { revalidatePath } from 'next/cache';

export const runtime = 'edge';

export async function GET(request: NextRequest) {
  const path = request.nextUrl.searchParams.get('path') || '/';
  revalidatePath(path);
  return NextResponse.json({ revalidated: true, now: Date.now() });
}
```

Please support this feature. It will help Next.js developers use Cloudflare Pages instead of Vercel.

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Introduces support for the suspense cache by intercepting the requests to our patched fetch.
- Support for the Cache API.
- Patches the built function contents to stop Next.js adding a `cache` property to `RequestInit` as it is not supported in `workerd`.

___

fixes #292

### DIFF
--- a/packages/next-on-pages/docs/caching.md
+++ b/packages/next-on-pages/docs/caching.md
@@ -4,49 +4,12 @@
 
 ## Storage Options
 
-There are various different bindings and storage options that `@cloudflare/next-on-pages` supports for caching.
+There are various different bindings and storage options that one could use for caching. At the moment, `@cloudflare/next-on-pages` supports the Cache API out-of-the-box.
 
-We recommend evaluating each option and choosing the one that best suits your use case, depending on latency and consistency requirements.
-
-### Workers KV
-
-[Workers KV](https://developers.cloudflare.com/workers/learning/how-kv-works/) is globally-distributed, low-latency storage option that is ideal for caching data. While it's designed for this use case, KV is an eventually-consistent data store, meaning that it can take up to 60 seconds for changes to propagate globally. This is fine for many use cases, but if you need to ensure that data is updated more frequently globally, you should consider a different storage option.
-
-1. Create a [new KV Namespace](https://dash.cloudflare.com/?to=/:account/workers/kv/namespaces).
-2. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
-3. Go to your Pages project Settings > Functions > KV Namespace Bindings.
-4. Add a new binding mapping `KV_SUSPENSE_CACHE` to your created KV Namespace.
-
-### Cloudflare D1
-
-[Cloudflare D1](https://developers.cloudflare.com/d1/) is a read-replicated, serverless database offering that uses SQLite. Unlike KV, it is strongly-consistent, meaning that changes will be accessible instantly, globally. However, while being read-replicated, it is not distributed in every data center, so there could be a minor impact on latency.
-
-1. Create a [new D1 Database](https://dash.cloudflare.com/?to=/:account/workers/d1) if you don't already have one.
-2. Create a new table in your database by clicking "Create table".
-   2.1. Give your table the name `suspense_cache`.
-   2.2. Add a row with the name `key` and type `text`, and set it as the primary key.
-   2.3. Add a row with the name `value` and type `text`.
-3. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
-4. Go to your Pages project Settings > Functions > D1 Database Bindings.
-5. Add a new binding mapping `D1_SUSPENSE_CACHE` to your D1 Database.
-
-If you would like to create the table with SQL instead, you can use the following query:
-
-```sql
-CREATE TABLE IF NOT EXISTS suspense_cache (key text PRIMARY KEY, value text NOT NULL);
-```
-
-### Cloudflare R2
-
-[Cloudflare R2](https://developers.cloudflare.com/r2/) is an S3-compatible object storage offering that is globally distributed and strongly consistent. It is ideal for storing large amounts of unstructured data, but is likely to experience higher latency that KV or D1.
-
-1. Create a [new R2 Bucket](https://dash.cloudflare.com/?to=/:account/r2/overview).
-2. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
-3. Go to your Pages project Settings > Functions > R2 Bucket Bindings.
-4. Add a new binding mapping `R2_SUSPENSE_CACHE` to your created R2 Bucket.
+In the future, support will be available for creating custom cache interfaces and using different bindings.
 
 ### Cache API
 
-The [Cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) is a per data-center cache that is ideal for storing data that is not required to be accessible globally. Due to this limitation, it is not a recommended storage option for Next.js caching and data revalidation - we suggest using one of the other options above.
+The [Cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) is a per data-center cache that is ideal for storing data that is not required to be accessible globally. It is worth noting that Vercel's Data Cache is regional, like with the Cache API, so there is no difference in terms of data availability.
 
-No additional setup is required to use the Cache API.
+Due to how the Cache API works, you need to be using a domain for your deployment for it to take effect.
--- a/packages/next-on-pages/docs/caching.md
+++ b/packages/next-on-pages/docs/caching.md
@@ -1,0 +1,52 @@
+# Caching and Data Revalidation
+
+`@cloudflare/next-on-pages` comes with support for data revalidation and caching for fetch requests. This is done in our router and acts as an extension to Next.js' built-in functionality.
+
+## Storage Options
+
+There are various different bindings and storage options that `@cloudflare/next-on-pages` supports for caching.
+
+We recommend evaluating each option and choosing the one that best suits your use case, depending on latency and consistency requirements.
+
+### Workers KV
+
+[Workers KV](https://developers.cloudflare.com/workers/learning/how-kv-works/) is globally-distributed, low-latency storage option that is ideal for caching data. While it's designed for this use case, KV is an eventually-consistent data store, meaning that it can take up to 60 seconds for changes to propagate globally. This is fine for many use cases, but if you need to ensure that data is updated more frequently globally, you should consider a different storage option.
+
+1. Create a [new KV Namespace](https://dash.cloudflare.com/?to=/:account/workers/kv/namespaces).
+2. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
+3. Go to your Pages project Settings > Functions > KV Namespace Bindings.
+4. Add a new binding mapping `KV_SUSPENSE_CACHE` to your created KV Namespace.
+
+### Cloudflare D1
+
+[Cloudflare D1](https://developers.cloudflare.com/d1/) is a read-replicated, serverless database offering that uses SQLite. Unlike KV, it is strongly-consistent, meaning that changes will be accessible instantly, globally. However, while being read-replicated, it is not distributed in every data center, so there could be a minor impact on latency.
+
+1. Create a [new D1 Database](https://dash.cloudflare.com/?to=/:account/workers/d1) if you don't already have one.
+2. Create a new table in your database by clicking "Create table".
+   2.1. Give your table the name `suspense_cache`.
+   2.2. Add a row with the name `key` and type `text`, and set it as the primary key.
+   2.3. Add a row with the name `value` and type `text`.
+3. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
+4. Go to your Pages project Settings > Functions > D1 Database Bindings.
+5. Add a new binding mapping `D1_SUSPENSE_CACHE` to your D1 Database.
+
+If you would like to create the table with SQL instead, you can use the following query:
+
+```sql
+CREATE TABLE IF NOT EXISTS suspense_cache (key text PRIMARY KEY, value text NOT NULL);
+```
+
+### Cloudflare R2
+
+[Cloudflare R2](https://developers.cloudflare.com/r2/) is an S3-compatible object storage offering that is globally distributed and strongly consistent. It is ideal for storing large amounts of unstructured data, but is likely to experience higher latency that KV or D1.
+
+1. Create a [new R2 Bucket](https://dash.cloudflare.com/?to=/:account/r2/overview).
+2. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
+3. Go to your Pages project Settings > Functions > R2 Bucket Bindings.
+4. Add a new binding mapping `R2_SUSPENSE_CACHE` to your created R2 Bucket.
+
+### Cache API
+
+The [Cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) is a per data-center cache that is ideal for storing data that is not required to be accessible globally. Due to this limitation, it is not a recommended storage option for Next.js caching and data revalidation - we suggest using one of the other options above.
+
+No additional setup is required to use the Cache API.
--- a/packages/next-on-pages/env.d.ts
+++ b/packages/next-on-pages/env.d.ts
@@ -6,6 +6,8 @@ declare global {
 			CF_PAGES?: string;
 			SHELL?: string;
 			KV_SUSPENSE_CACHE?: KVNamespace;
+			D1_SUSPENSE_CACHE?: D1Database;
+			R2_SUSPENSE_CACHE?: R2Bucket;
 			[key: string]: string | Fetcher;
 		}
 	}
--- a/packages/next-on-pages/env.d.ts
+++ b/packages/next-on-pages/env.d.ts
@@ -5,6 +5,7 @@ declare global {
 			npm_config_user_agent?: string;
 			CF_PAGES?: string;
 			SHELL?: string;
+			KV_SUSPENSE_CACHE?: KVNamespace;
 			[key: string]: string | Fetcher;
 		}
 	}
--- a/packages/next-on-pages/env.d.ts
+++ b/packages/next-on-pages/env.d.ts
@@ -5,9 +5,6 @@ declare global {
 			npm_config_user_agent?: string;
 			CF_PAGES?: string;
 			SHELL?: string;
-			KV_SUSPENSE_CACHE?: KVNamespace;
-			D1_SUSPENSE_CACHE?: D1Database;
-			R2_SUSPENSE_CACHE?: R2Bucket;
 			[key: string]: string | Fetcher;
 		}
 	}
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -454,6 +454,14 @@ function fixFunctionContents(contents: string): string {
 		'$1null$2null$3null$4',
 	);
 
+	// The workers runtime does not implement `cache` on RequestInit. This is used in Next.js' patched fetch.
+	// Due to this, we remove the `cache` property from those that Next.js adds to RequestInit.
+	// https://github.com/vercel/next.js/blob/269114b5cc583f0c91e687c1aeb61503ef681b91/packages/next/src/server/lib/patch-fetch.ts#L304
+	contents = contents.replace(
+		/"cache",("credentials","headers","integrity","keepalive","method","mode","redirect","referrer")/gm,
+		'$1',
+	);
+
 	return contents;
 }
 
--- a/packages/next-on-pages/templates/_worker.js/index.ts
+++ b/packages/next-on-pages/templates/_worker.js/index.ts
@@ -28,6 +28,7 @@ export default {
 		}
 
 		return envAsyncLocalStorage.run(
+			// NOTE: The `SUSPENSE_CACHE_URL` is used to tell the Next.js Fetch Cache where to send requests.
 			{ ...env, NODE_ENV: __NODE_ENV__, SUSPENSE_CACHE_URL },
 			async () => {
 				const url = new URL(request.url);
--- a/packages/next-on-pages/templates/_worker.js/index.ts
+++ b/packages/next-on-pages/templates/_worker.js/index.ts
@@ -1,6 +1,6 @@
+import { SUSPENSE_CACHE_URL } from '../cache';
 import { handleRequest } from './handleRequest';
 import {
-	SUSPENSE_CACHE_URL,
 	adjustRequestForVercel,
 	handleImageResizingRequest,
 	patchFetch,
--- a/packages/next-on-pages/templates/_worker.js/index.ts
+++ b/packages/next-on-pages/templates/_worker.js/index.ts
@@ -1,5 +1,6 @@
 import { handleRequest } from './handleRequest';
 import {
+	SUSPENSE_CACHE_URL,
 	adjustRequestForVercel,
 	handleImageResizingRequest,
 	patchFetch,
@@ -25,8 +26,9 @@ export default {
 				{ status: 503 },
 			);
 		}
+
 		return envAsyncLocalStorage.run(
-			{ ...env, NODE_ENV: __NODE_ENV__ },
+			{ ...env, NODE_ENV: __NODE_ENV__, SUSPENSE_CACHE_URL },
 			async () => {
 				const url = new URL(request.url);
 				if (url.pathname.startsWith('/_next/image')) {
--- a/packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
@@ -1,0 +1,178 @@
+export const SUSPENSE_CACHE_URL = 'INTERNAL_SUSPENSE_CACHE_HOSTNAME';
+
+/**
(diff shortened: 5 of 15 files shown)


packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
```diff
@@ -0,0 +1,178 @@
+export const SUSPENSE_CACHE_URL = 'INTERNAL_SUSPENSE_CACHE_HOSTNAME';
+
+/**
+ * Gets the cache interface to use for the suspense cache.
+ *
+ * @returns Interface for the suspense cache.
+ */
+export async function getSuspenseCacheInterface(): Promise<CacheInterface> {
+	if (process.env.KV_SUSPENSE_CACHE) {
+		return new KvCacheInterface(process.env.KV_SUSPENSE_CACHE);
```
james-elicx, line 10: I would like if we used a hardcoded name for bindings, e.g. `KV_SUSPENSE_CACHE`, that people can then add to their project. The only question is, what should they be called.

I am open to changing the name used here - I went with this one as Vercel call the API endpoint the Suspense Cache.

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -0,0 +1,176 @@
...
+
+	// Check if the cache entry is stale or fresh based on the tags.
+	const tags = getDerivedTags(data.value.data.tags ?? []);
+	const isStale = tags.some(tag => {
+		const tagEntry = cache.tagsManifest?.items?.[tag];
+		return (
+			tagEntry?.revalidatedAt && tagEntry?.revalidatedAt >= data.lastModified
+		);
+	});
+
```
bluebeel, line 84: are you not missing the following part of the [fetch-cache](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L246) implementation? 
```typescript
    if (data?.value?.kind === 'FETCH') {
      const innerData = data.value.data
      const derivedTags = getDerivedTags(innerData.tags || [])

      if (
        derivedTags.some((tag) => {
          return this.revalidatedTags.includes(tag)
        })
      ) {
        data = undefined
      }
    }
```
If a tag got revalidated, it skips the cached entry to force a new fetch

bluebeel:
Hey,

Your code look a hell more clean and better than mine 🤣 


What happens if a user make multiple binding?
```
export async function getSuspenseCacheInterface(): Promise<CacheInterface> {
	if (process.env.KV_SUSPENSE_CACHE) {
		return new KvCacheInterface(process.env.KV_SUSPENSE_CACHE);
	}

	const cacheApi = await caches.open('suspense-cache');
	return new CacheApiInterface(cacheApi);
}
```
Do we want to only let a user write in a single destination or multiple one?
If we enforce only one but multiple one are still set, which one should be consider as the "priority/best" one?

---
Is it possible to add R2 binding as a option too?

---
Is it possible to update your example to add the revalidate action so we can test it? 🙏🏾

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
bluebeel: I would leave it to follow the same naming as nextjs codebase 🤔

packages/next-on-pages/templates/_worker.js/utils/cache.ts
james-elicx: I actually didn't think about trying to implement that part because it would still run in the fetch cache, since this is just replacing the API endpoint.

After some testing though, it appears that the revalidated tags aren't properly tracked there, so I've just updated the code to track revalidated tags locally, so that server actions will be able to properly return the fresh data instead of requiring a refresh.

Thanks for getting me look at that part :)

james-elicx:
> What happens if a user make multiple binding?
> 
> ```
> export async function getSuspenseCacheInterface(): Promise<CacheInterface> {
> 	if (process.env.KV_SUSPENSE_CACHE) {
> 		return new KvCacheInterface(process.env.KV_SUSPENSE_CACHE);
> 	}
> 
> 	const cacheApi = await caches.open('suspense-cache');
> 	return new CacheApiInterface(cacheApi);
> }
> ```
> 
> Do we want to only let a user write in a single destination or multiple one? If we enforce only one but multiple one are still set, which one should be consider as the "priority/best" one?

A user should only have one binding set IMO, but yeah, it would be kind of a priority way I guess. I would probably say KV > D1 > DO > R2 > Cache API, but it shouldn't really matter though since people should not set multiple.

> Is it possible to add R2 binding as a option too?

Yes, it just needs another instance of the cache interface class for it, like I did with KV and the Cache API.

> Is it possible to update your example to add the revalidate action so we can test it? 🙏🏾

Updated the deployed demo with tag + path revalidation buttons 🙂

bluebeel:
Thanks it works 💯 

Would be interesting to see a benchmark of the different binding how they're behaving and when we access them from around the world.

I think personally I would go for the normal cache API or DO/R2 if I want a global unique store.
KV is the worst one for me because of the avg 60s update sync around the world... 😢

james-elicx:
> Thanks it works 💯
> 
> Would be interesting to see a benchmark of the different binding how they're behaving and when we access them from around the world.
> 
> I think personally I would go for the normal cache API or DO/R2 if I want a global unique store. KV is the worst one for me because of the avg 60s update sync around the world... 😢

Well, it's a trade-off between latency and availability. Personally, I would not advise people to use the Cache API, since that is different in each data centre.

While KV is eventually globally consistent, it's low latency and is designed for a high volume of reads, so it is more preferable and intended for this kind of use case - it's good for cases where the eventual consistency isn't that important. If one wanted relatively low latency and high consistency, D1 might be a good option, but that's still in alpha. I would assume that R2 probably has a higher latency, but that is also strongly consistent. And durable objects, well, that's the one thing I know nothing about.

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
```diff
@@ -0,0 +1,248 @@
...
+
+	if (process.env.D1_SUSPENSE_CACHE) {
+		return new D1CacheInterface(process.env.D1_SUSPENSE_CACHE);
+	}
+
+	if (process.env.R2_SUSPENSE_CACHE) {
+		return new R2CacheInterface(process.env.R2_SUSPENSE_CACHE);
+	}
+
+	const cacheApi = await caches.open('suspense-cache');
```
james-elicx, line 23: I wonder if it might be better to use `caches.default` here, instead of opening a new one. Thoughts?

packages/next-on-pages/docs/caching.md
```diff
@@ -0,0 +1,52 @@
...
+
+[Workers KV](https://developers.cloudflare.com/workers/learning/how-kv-works/) is globally-distributed, low-latency storage option that is ideal for caching data. While it's designed for this use case, KV is an eventually-consistent data store, meaning that it can take up to 60 seconds for changes to propagate globally. This is fine for many use cases, but if you need to ensure that data is updated more frequently globally, you should consider a different storage option.
+
+1. Create a [new KV Namespace](https://dash.cloudflare.com/?to=/:account/workers/kv/namespaces).
+2. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
+3. Go to your Pages project Settings > Functions > KV Namespace Bindings.
+4. Add a new binding mapping `KV_SUSPENSE_CACHE` to your created KV Namespace.
+
+### Cloudflare D1
+
```
dario-piotrowicz, line 21: I think we should warn that D1 is in alpha? (so not really stable and/or production ready)

packages/next-on-pages/docs/caching.md
```diff
@@ -0,0 +1,52 @@
...
+[Cloudflare R2](https://developers.cloudflare.com/r2/) is an S3-compatible object storage offering that is globally distributed and strongly consistent. It is ideal for storing large amounts of unstructured data, but is likely to experience higher latency that KV or D1.
+
+1. Create a [new R2 Bucket](https://dash.cloudflare.com/?to=/:account/r2/overview).
+2. Find your [Pages project](https://dash.cloudflare.com/?to=/:account/workers-and-pages) in the Cloudflare dashboard.
+3. Go to your Pages project Settings > Functions > R2 Bucket Bindings.
+4. Add a new binding mapping `R2_SUSPENSE_CACHE` to your created R2 Bucket.
+
+### Cache API
+
+The [Cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) is a per data-center cache that is ideal for storing data that is not required to be accessible globally. Due to this limitation, it is not a recommended storage option for Next.js caching and data revalidation - we suggest using one of the other options above.
```
dario-piotrowicz, line 50: `it is not a recommended storage option for Next.js caching` I'm not 100% sure about this, I'd be keen to hear what @GregBrimble thinks 🤔

Also could you add a note/warning that this isn't going to work on pages.dev and that this caching only works on custom domains?

dario-piotrowicz: Also you don't provide any instructions here, I don't think there's actually anything to do, but could you mention that so people are aware of it?

(PS: I haven't finished reviewing the PR, but I assume that this is the default and people get it out of the box if they don't have any of the other binding set?)

packages/next-on-pages/docs/caching.md
```diff
@@ -0,0 +1,52 @@
+# Caching and Data Revalidation
+
+`@cloudflare/next-on-pages` comes with support for data revalidation and caching for fetch requests. This is done in our router and acts as an extension to Next.js' built-in functionality.
+
+## Storage Options
+
+There are various different bindings and storage options that `@cloudflare/next-on-pages` supports for caching.
```
dario-piotrowicz, line 7: I would probably try to mention upfront that this relies on naming conventions (I think it is made clear in the next sections but I'd still clarify it here 🤔)

something like this:
```suggestion
There are various different bindings and storage options that `@cloudflare/next-on-pages` supports for caching. Besides the default behavior implemented via the [Workers Cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) the other options are based on bindings naming conventions, meaning that providing that they get enabled when bindings with the appropriate names are detected.
```

it's just an idea I'm ok with you don't want to.

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
dario-piotrowicz: I think it's totally fine to open a specific cache, would using the default one provide any benefit?

packages/next-on-pages/docs/caching.md
james-elicx: > Also you don't provide any instructions here, I don't think there's actually anything to do, but could you mention that so people are aware of it?

The next line says that no setup is required 😛. And yes, it's the default option if no binding is provided - probably worth clarifying that here.

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
james-elicx: > I think it's totally fine to open a specific cache, would using the default one provide any benefit?

I'm not really sure. I want to say it wouldn't

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
```diff
@@ -0,0 +1,248 @@
...
+		const value = await this.cache.get(key);
+		return value ? value.text() : null;
+	}
+
+	public override async delete(key: string) {
+		await this.cache.delete(key);
+	}
+}
+
+// TODO: DO Interface
```
dario-piotrowicz, line 241: ?

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
```diff
@@ -0,0 +1,248 @@
...
+
+type TagsManifest = {
+	version: 1;
+	items: { [tag: string]: TagsManifestItem };
+};
+
+type TagsManifestItem = { keys: string[]; revalidatedAt?: number };
```
dario-piotrowicz, line 248: what's are these types? do they come from Next/Vercel?

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -0,0 +1,199 @@
...
+	await cache.setTags(tags, { cacheKey });
+
+	getDerivedTags(tags).forEach(tag => revalidatedTags.delete(tag));
+
+	return new Response(null, { status: 200 });
+}
+
+type CacheEntry = { lastModified: number; value: NextCachedFetchValue };
+
+// https://github.com/vercel/next.js/blob/canary/packages/next/src/server/response-cache/types.ts
```
dario-piotrowicz, line 153:
```suggestion
// https://github.com/vercel/next.js/blob/fda1ecc/packages/next/src/server/response-cache/types.ts#L16
```

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -0,0 +1,199 @@
...
+		status?: number;
+		tags?: string[];
+	};
+	revalidate: number;
+};
+
+/**
+ * Derives a list of tags from the given tags. This is taken from the Next.js source code.
+ *
+ * @see https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/incremental-cache/utils.ts
```
dario-piotrowicz, line 169:
```suggestion
 * @see https://github.com/vercel/next.js/blob/1286e145/packages/next/src/server/lib/incremental-cache/utils.ts
```

dario-piotrowicz (commented):
Looks awesome! thanks a lot @james-elicx 😍 🚀 

I added a bunch of minor comments/suggestions please have a look 🙂 

I still need to run this locally and play a bit with it (but the code looks great 😄)

(Also I am still not sure if we should have built-in "adapters" for the caching 😓)

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -454,6 +454,14 @@ function fixFunctionContents(contents: string): string {
 		'$1null$2null$3null$4',
 	);
 
+	// The workers runtime does not implement `cache` on RequestInit. This is used in Next.js' patched fetch.
+	// Due to this, we remove the `cache` property from those that Next.js adds to RequestInit.
```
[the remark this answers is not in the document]
james-elicx: Yes. It throws an error because workerd doesn't support it.

packages/next-on-pages/env.d.ts
```diff
@@ -5,6 +5,9 @@ declare global {
 			npm_config_user_agent?: string;
 			CF_PAGES?: string;
 			SHELL?: string;
+			KV_SUSPENSE_CACHE?: KVNamespace;
+			D1_SUSPENSE_CACHE?: D1Database;
+			R2_SUSPENSE_CACHE?: R2Bucket;
```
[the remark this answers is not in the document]
james-elicx: What about NOP instead of NEXT_ON_PAGES?

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
```diff
@@ -0,0 +1,248 @@
+export const SUSPENSE_CACHE_URL = 'INTERNAL_SUSPENSE_CACHE_HOSTNAME';
```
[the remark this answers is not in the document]
james-elicx: That is what the environment variable is called in the Next.js code that this is used for.

packages/next-on-pages/templates/_worker.js/index.ts
```diff
@@ -25,8 +26,9 @@ export default {
 				{ status: 503 },
 			);
 		}
+
 		return envAsyncLocalStorage.run(
-			{ ...env, NODE_ENV: __NODE_ENV__ },
+			{ ...env, NODE_ENV: __NODE_ENV__, SUSPENSE_CACHE_URL },
```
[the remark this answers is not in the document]
james-elicx: Because that's the env variable used in the fetch cache so that it knows where to send the api requests for the cache. those api requests are the ones we then intercept.

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
james-elicx: For the future, adding a cache interface for Durable Objects. Can remove the TODO if you want, just thought it would be nice to leave in case someone wants to add a Durable Objects one.

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
james-elicx: Next.js yes - will add a link to the code above.

packages/next-on-pages/templates/_worker.js/utils/cache-interface.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah 😓

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah ok thanks 🙂 

I would have hoped/imagined that it would just be ignored

packages/next-on-pages/env.d.ts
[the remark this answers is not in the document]
dario-piotrowicz: sure 🙂

packages/next-on-pages/templates/_worker.js/index.ts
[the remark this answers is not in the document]
dario-piotrowicz: ahhhh, ok it's Next that it's accessing it I see! (https://github.com/vercel/next.js/blob/2cf5d3a8aa325f8bece1094c9e566311e604e114/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L28)

Could you add a comment regarding that so that we're sure to remember that it's used by them and not us? (otherwise this could be removed as it doesn't look like actually being used)

packages/next-on-pages/templates/cache/interface.ts
```diff
@@ -0,0 +1,244 @@
...
+
+					const tagEntry = this.tagsManifest?.items?.[tag];
+					return (
+						tagEntry?.revalidatedAt &&
+						tagEntry?.revalidatedAt >= (data.lastModified ?? Date.now())
+					);
+				});
+
+				// Don't return stale data from the cache.
+				return isStale ? null : data;
```
dario-piotrowicz, line 102: I think we discussed this and that even state data should be returned?

packages/next-on-pages/templates/cache/interface.ts
```diff
@@ -0,0 +1,244 @@
...
+			if (rawManifest) {
+				this.tagsManifest = JSON.parse(rawManifest) as TagsManifest;
+			}
+		} catch (e) {
+			// noop
+		}
+
+		if (!this.tagsManifest) {
+			this.tagsManifest = { version: 1, items: {} };
+		}
```
dario-piotrowicz, line 137:
```suggestion
		this.tagsManifest ??= { version: 1, items: {} };
```

dario-piotrowicz (commented):
Looks great! 🚀 😄 

I just left a few comments please have a look

I also still need to test this deployed, other that these it looks good for me 😄

packages/next-on-pages/templates/cache/interface.ts
```diff
@@ -0,0 +1,244 @@
+export const SUSPENSE_CACHE_URL = 'INTERNAL_SUSPENSE_CACHE_HOSTNAME';
+
+// Set to track the revalidated tags in requests.
+const revalidatedTags = new Set<string>();
+
+/** Generic interface for the Suspense Cache. */
+export class CacheInterface {
```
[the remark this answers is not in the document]
james-elicx: Yeah I like `CacheAdaptor`, good call 👍🏻

packages/next-on-pages/templates/cache/cache-api.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	public override async update(key: string, value: string) {
+		const cache = await caches.open('suspense-cache');
+
+		// Figure out the max-age for the cache entry.
+		const entry = JSON.parse(value) as IncrementalCacheValue;
+		const maxAge =
+			key === this.tagsManifestKey || entry.kind !== 'FETCH'
+				? '31536000'
+				: entry.revalidate;
```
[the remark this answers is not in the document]
james-elicx: Yeah will do a little later. Had some things to take care of this morning!

packages/next-on-pages/templates/cache/interface.ts
james-elicx: This part is different to what we discussed. Here, it is considered stale and returning null if a revalidation is triggered and the data is still old. What we discussed was returning data that is old but has not had a revalidation triggered for it yet.

I would also note that this is consistent with how Next.js' file system cache works - https://github.com/vercel/next.js/blob/baef523fb08e1a7d588026ba784247ebcfd07740/packages/next/src/server/lib/incremental-cache/file-system-cache.ts#L277

dario-piotrowicz: ahhh ok I see! ok, thanks! 🙏

packages/next-on-pages/templates/cache/interface.ts
```diff
@@ -0,0 +1,244 @@
+export const SUSPENSE_CACHE_URL = 'INTERNAL_SUSPENSE_CACHE_HOSTNAME';
```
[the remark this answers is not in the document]
dario-piotrowicz: @james-elicx could you add a TLD here?

```suggestion
export const SUSPENSE_CACHE_URL = 'INTERNAL_SUSPENSE_CACHE_HOSTNAME.com';
```

packages/next-on-pages/templates/cache/interface.ts
```diff
@@ -0,0 +1,244 @@
...
+export class CacheInterface {
+	/** The tags manifest for fetch calls. */
+	public tagsManifest: TagsManifest | undefined;
+	/** The key used for the tags manifest in the cache. */
+	public tagsManifestKey = 'tags-manifest';
+
+	/**
+	 * @param ctx The incremental cache context.
+	 */
+	constructor(protected ctx: Record<string, unknown> = {}) {}
```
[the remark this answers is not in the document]
dario-piotrowicz: I see... thanks for the explanation 🙂 

sorry for being annoying, but could you add a comment mentioning that we don't use it? 🙏

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -0,0 +1,79 @@
+import {
+	CacheAdaptor,
+	IncrementalCacheValue,
+	SUSPENSE_CACHE_URL,
+} from '../../cache';
```
dario-piotrowicz, line 5:
```suggestion
import type {
	CacheAdaptor,
	IncrementalCacheValue} from '../../cache';
import {
	SUSPENSE_CACHE_URL,
} from '../../cache';
```

(PS: these are casing my IDE's eslint to complain, I think it's valid, if I remember correctly we aren't including the templates dir in the eslint config, I need to fix that)

packages/next-on-pages/templates/cache/interface.ts
james-elicx:
```suggestion
	 * @param ctx The incremental cache context from Next.js. NOTE: This is not currently utilised in NOP.
	 */
	constructor(protected ctx: Record<string, unknown> = {}) {}
```

packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -0,0 +1,243 @@
...
+
+/** Generic adaptor for the Suspense Cache. */
+export class CacheAdaptor {
+	/** The tags manifest for fetch calls. */
+	public tagsManifest: TagsManifest | undefined;
+	/** The key used for the tags manifest in the cache. */
+	public tagsManifestKey = 'tags-manifest';
+
+	/**
+	 * @param ctx The incremental cache context.
```
james-elicx, line 15:
```suggestion
	 * @param ctx The incremental cache context from Next.js. NOTE: This is not currently utilised in NOP.
```

bluebeel:
Hello,
I see with the latest commit that you removed the env binding for the specific storage? The adapter will have to be added in user-land in the future?

james-elicx:
> Hello, I see with the latest commit that you removed the env binding for the specific storage? The adapter will have to be added in user-land in the future?

We decided to change the approach slightly where we are creating a generic adaptor that other classes can extend from and use to add additional storage mechanisms. By default, we will use the Cache API, however, if a user wants to use something else like a KV binding, they will need to specify in their `next.config.js` an import to the relevant code for that storage mechanism.

This is because part of our plan is to allow users to specify their own storage mechanisms in the future, so when we add support for other bindings, it will be done with the same approach that we will use for other custom user-provided solutions, and also will align with how Next.js expects a custom cache to be provided to itself.

We are adding back support for a KV binding in a follow-up PR, although that will be an export that the user will then have to use in their `next.config.js`, which is where we will then try to get it from, instead of doing it based on an environment variable existing.

dario-piotrowicz (approved):
This is awesome! thanks a lot @james-elicx for all the effort put into landing this! ❤️ 🚀

dario-piotrowicz:
And thanks @bluebeel for starting this effort and helping reviewing this PR! 🙂 

Sorry for not keeping you in the loop regarding the latest changes, it's as James mentioned the bindings (not sure if all) should be added back in a follow up PR 🙂 (+ the possibility for developers to provide their own adapter so that we don't force them what solution to use for caching)

PR #426, opened 2023-08-11
Improving e2es and removing some e2e-like unit tests

dario-piotrowicz:


pages-e2e/fixtures/pages12.3.0/main.fixture
```diff
@@ -3,11 +3,12 @@
 		"simplePagesApiRoutes",
 		"simplePagesSsrRoutes",
 		"simplePagesStaticRoutesAndAssets",
-		"pagesMiddleware"
+		"pagesMiddleware",
+		"pagesRouting"
```
dario-piotrowicz, line 7: I did not add the new `configRewritesRedirectsHeaders` to this fixture as it seems like such things don't work in Next 12.3.0 (not even with `next dev`)

I think it's fine not to include them in the tests, if later we end up having issues we might add 12.3.0 specific tests for whatever subset of such configs it supports

@james-elicx does Next 12.3.0 not supporting redirects and rewrites sounds correct to you? (especially with the `beforeFiles` and `afterFiles` options)

james-elicx: The docs indicate that it should work, but if it's not working in `next dev`, well, there's not much we can do about it...

dario-piotrowicz: well the docs don't specifically say when the `beforeFiles`, etc... were introduced 😕 🤷 

anyways even if they were I did saw other stuff that particularly in 12.3.0 is supposed to work but doesn't (in `next dev`)

so yeah I guess it's ok to leave this one be for now 🤷

dario-piotrowicz: > there's not much we can do about it...

as I said we could check subsets of the configs that work... or try bumping the version slightly to newer patch versions to see if it helps... but doesn't really seem worth the effort here 🤔

dario-piotrowicz:
I've converted this back into draft as I am trying to figure out why the e2es seem to still not properly run

jculvey (approved):

PR #439, opened 2023-08-24
Add vercel@32 as a peer dependency

Karibash:
[vercel@32](https://github.com/vercel/vercel/releases/tag/vercel%4032.0.0) has been released.
The only major change is that the Node.js minimum has been raised, so adding it to peer dependencies should not be a problem.


dario-piotrowicz (approved):
Looks to me like a safe change to make 🙂 

Thanks a bunch @Karibash ❤️

PR #445, opened 2023-08-27
fix: `process.env` not enumerable due to being a proxy

james-elicx:
This PR does the following:
- Allows `process.env` to be enumerated over.

This enables things like `Object.entries(process.env)` to work when it's a proxy.

### DIFF
--- a/packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
+++ b/packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
@@ -21,7 +21,7 @@ export function generateGlobalJs(): string {
 					{},
 					{
 						ownKeys: () => Reflect.ownKeys(envAsyncLocalStorage.getStore()),
-						getOwnPropertyDescriptor: () => ({ enumerable: true, configurable: true }),
+            getOwnPropertyDescriptor: (_, ...args) => Object.getOwnPropertyDescriptor(als.getStore(), ...args),
 						get: (_, property) => Reflect.get(envAsyncLocalStorage.getStore(), property),
 						set: (_, property, value) => Reflect.set(envAsyncLocalStorage.getStore(), property, value),
 				}),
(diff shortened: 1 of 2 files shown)


packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
```diff
@@ -20,6 +20,8 @@ export function generateGlobalJs(): string {
 				env: new Proxy(
 					{},
 					{
+						ownKeys: () => Reflect.ownKeys(envAsyncLocalStorage.getStore()),
+						getOwnPropertyDescriptor: () => ({ enumerable: true, configurable: true }),
```
dario-piotrowicz, line 24: I don't think that this is correct as it would return non-`undefined` for non existing properties, it also doesn't return the `value` of the property when it should, and in general this simply hardcodes a result for all properties

what about?
```suggestion
            getOwnPropertyDescriptor: (_, ...args) => Object.getOwnPropertyDescriptor(als.getStore(), ...args),
```

it does seem to work as expected: https://stackblitz.com/edit/stackblitz-starters-a2p5yp?file=index.mjs

dario-piotrowicz (commented):
Thanks @james-elicx :smiley: 

I've added a suggestion about one line which I think is not correct, please have a look :slightly_smiling_face:

packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
james-elicx: fine by me, i just found this from stack overflow 👍🏻

packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
```diff
@@ -20,6 +20,8 @@ export function generateGlobalJs(): string {
 				env: new Proxy(
 					{},
 					{
+						ownKeys: () => Reflect.ownKeys(envAsyncLocalStorage.getStore()),
+            getOwnPropertyDescriptor: (_, ...args) => Object.getOwnPropertyDescriptor(als.getStore(), ...args),
```
james-elicx, line 24:
```suggestion
						getOwnPropertyDescriptor: (_, ...args) =>
							Reflect.getOwnPropertyDescriptor(envAsyncLocalStorage.getStore(), ...args),
```

dario-piotrowicz (approved):

PR #447, opened 2023-08-30
fix: broken function entrypoints on windows

james-elicx:
This PR does the following:
- Normalizes the injected entrypoint for Windows platforms.

This fixes an issue where the function entrypoint we inject breaks as it might use double backslashes on Windows because, well, Windows has to ruin everything.

```
"/api/hello":{type:"function",entrypoint:"__next-on-pages-dist__\functionsapihello.func.js"}
```

### DIFF
--- a/.changeset/orange-swans-peel.md
+++ b/.changeset/orange-swans-peel.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Normalize the injected function entrypoints for Windows.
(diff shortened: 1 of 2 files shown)


dario-piotrowicz (approved):
Thanks @james-elicx 🙂 

I've got one small suggestion but if you want to keep things in the current way for now I'm on with it

______

Here you normalize the entrypoint path just as you create the build output item, but would't it be better to normalize all the paths up the chain at `processVercelOutput` as soon as they are generated? (instead of normalize things as needed/used)

james-elicx:
> Here you normalize the entrypoint path just as you create the build output item, but would't it be better to normalize all the paths up the chain at `processVercelOutput` as soon as they are generated? (instead of normalize things as needed/used)

I think we already do in quite a few places, but this is catching it at the final place so it removes all risk of a future change causing a regression.

Quite honestly, I thought we had already resolved these Windows path problems, so I think that addressing it at the final place where it's used is a good approach to remove risk.

dario-piotrowicz:
> a regression.
> 
> Quite honestly, I thought we had already resolved these Windows path problems, so I think that addressing it at the final place where it's used is a good approach to prevent accidental problems.

I think we should just do it at the very top and that's it instead of repeating the operation (not that it's a costly operation anyways...)

in order to avoid regressions and missing places we could test it thoroughly...  but yeah it might not be trivial.... 🤷

PR #449, opened 2023-09-01
fix: named captures in `has` fields not applying to route dest

### LINKED ISSUE #448: [🐛 Bug]: Rewrites with `has` query option doesn't behave as expected

<details>
  <summary>env info</summary>

```
  System:
	Platform: darwin
	Arch: arm64
	Version: Darwin Kernel Version 22.6.0: Wed Jul  5 22:17:35 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T8112
	CPU: (8) arm64 Apple M2
	Memory: 8 GB
	Shell: /opt/homebrew/bin/zsh
  Binaries:
	Node: 18.17.0
	Bun: N/A
	pnpm: N/A
	Yarn: 1.22.19
	npm: 9.6.7
  Relevant Packages:
	@cloudflare/next-on-pages: 1.6.0
	vercel: 32.1.0
	next: 13.4.19
```
</details>

### Description

Hi there,

During deployment we've noticed an issue which only seems to affect our project once deployed via next-on-pages. The issue pertains to Next's 'rewrites' config option which doesn't _appear_ to be supported properly at the moment.

On the [supported doc](https://github.com/cloudflare/next-on-pages/blob/main/packages/next-on-pages/docs/supported.md) it does state that "routes `has`" is supported; at present however I think some of the behaviour is not supported. For example, the following "rewrites" config:

```
async rewrites() {
    return {
      beforeFiles: [{
          source: '/user-query',
          has: [{
              type: 'query',
              key: 'name',
              value: '(?<name>.*)',
          }],
          destination: '/user/:name',
      }],
    };
  },
```

Which aims to rewrite "/user-query?name=bacon" to "/user/bacon". This does not appear to work correctly in next-on-pages. I'm not sure if this is a particularly unusual pattern or not but does seem to follow the docs on nextjs and does work when running in development or production build with next. It appears to only fail when locally using `wrangler` or when deployed to a preview URL on Cloudflare.

I also believe fixing this could fix some of the issues with `@sentry/nextjs` as their tunnelling option makes use of a very similar rewrite under-the-hood.

<details>
  <summary>@sentry/nextjs rewrite option</summary>

  ```
    {
      // Matched rewrite routes will look like the following: `[tunnelPath]?o=[orgid]&p=[projectid]`
      // Nextjs will automatically convert `source` into a regex for us
      source: `${tunnelPath}(/?)`,
      has: [
        {
          type: 'query',
          key: 'o', // short for orgId - we keep it short so matching is harder for ad-blockers
          value: '(?<orgid>.*)',
        },
        {
          type: 'query',
          key: 'p', // short for projectId - we keep it short so matching is harder for ad-blockers
          value: '(?<projectid>.*)',
        },
      ],
      destination: 'https://o:orgid.ingest.sentry.io/api/:projectid/envelope/',
    }
  ```
</details>

### Reproduction

https://github.com/emma-cw/next-pages-rewrites-test

Pulling this repo and running with `npm run dev`/`yarn dev` will run successfully and the rewrite will behave as expected. For example `http://localhost:3000/user/bacon` and `http://localhost:3000/user-query?name=bacon` will display the same very basic page.

However if running with `npx @cloudflare/next-on-pages@1` and then `npx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat` the rewrite no longer functions as expected. Instead,  `http://localhost:8788/user/bacon` will show correctly but `http://localhost:8788/user-query?name=bacon` will return `404` instead.

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR does the following:
- Changes the logic to use the PCRE match function for `has` entries as the regex is expected to be in the PCRE format.
- For PCRE matches in the `has` entries, attempts to update the route's destination with any named capture groups extracted.

Relevant docs: https://nextjs.org/docs/app/api-reference/next-config-js/rewrites#header-cookie-and-query-matching

fixes #448

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
@@ -1,7 +1,10 @@
+import { applyPCREMatches, matchPCRE } from './pcre';
+
 type HasFieldRequestProperties = {
 	url: URL;
 	cookies: Record<string, string>;
 	headers: Headers;
+	routeDest?: string;
 };
 
 /**
@@ -13,34 +16,46 @@ type HasFieldRequestProperties = {
  */
 export function hasField(
 	has: VercelHasField,
-	{ url, cookies, headers }: HasFieldRequestProperties,
-): boolean {
+	{ url, cookies, headers, routeDest }: HasFieldRequestProperties,
+): { valid: boolean; newRouteDest?: string } {
 	switch (has.type) {
 		case 'host': {
-			return url.hostname === has.value;
+			return { valid: url.hostname === has.value };
 		}
 		case 'header': {
 			if (has.value !== undefined) {
-				return !!headers.get(has.key)?.match(has.value);
+				return { valid: !!matchPCRE(has.value, headers.get(has.key)).match };
 			}
 
-			return headers.has(has.key);
+			return { valid: headers.has(has.key) };
 		}
 		case 'cookie': {
 			const cookie = cookies[has.key];
 
-			if (has.value !== undefined) {
-				return !!cookie?.match(has.value);
+			if (cookie && has.value !== undefined) {
+				return { valid: !!matchPCRE(has.value, cookie).match };
 			}
 
-			return cookie !== undefined;
+			return { valid: cookie !== undefined };
 		}
 		case 'query': {
 			if (has.value !== undefined) {
-				return !!url.searchParams.get(has.key)?.match(has.value);
+				const { match, captureGroupKeys } = matchPCRE(
+					has.value,
+					url.searchParams.get(has.key),
+				);
+
+				const newRouteDest =
+					match && captureGroupKeys.length && routeDest
+						? applyPCREMatches(routeDest, match, captureGroupKeys, {
+								namedOnly: true,
+						  })
+						: undefined;
+
+				return { valid: !!match, newRouteDest };
 			}
 
-			return url.searchParams.has(has.key);
+			return { valid: url.searchParams.has(has.key) };
 		}
 	}
 }
--- a/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
@@ -1,4 +1,3 @@
-import type { MatchPCREResult } from './pcre';
 import { applyPCREMatches, matchPCRE } from './pcre';
 
 type HasFieldRequestProperties = {
@@ -9,13 +8,14 @@ type HasFieldRequestProperties = {
 };
 
 /**
- * Checks if a Vercel source route's `has` record conditions match a request.
+ * Checks if a Vercel source route's `has` record conditions match a request, and whether the request
+ * destination should be updated based on the `has` record.
  *
  * @param has The `has` record conditions to check against the request.
  * @param requestProperties The request properties to check against.
- * @returns Whether the request matches the `has` record conditions.
+ * @returns Whether the request matches the `has` record conditions, and the new destination if it changed.
  */
-export function hasField(
+export function checkhasField(
 	has: VercelHasField,
 	{ url, cookies, headers, routeDest }: HasFieldRequestProperties,
 ): { valid: boolean; newRouteDest?: string } {
@@ -25,11 +25,11 @@ export function hasField(
 		}
 		case 'header': {
 			if (has.value !== undefined) {
-				const match = matchPCRE(has.value, headers.get(has.key));
-				return {
-					valid: !!match.match,
-					newRouteDest: tryApplyMatch(routeDest, match),
-				};
+				return getHasFieldPCREMatchResult(
+					has.value,
+					headers.get(has.key),
+					routeDest,
+				);
 			}
 
 			return { valid: headers.has(has.key) };
@@ -38,22 +38,18 @@ export function hasField(
 			const cookie = cookies[has.key];
 
 			if (cookie && has.value !== undefined) {
-				const match = matchPCRE(has.value, cookie);
-				return {
-					valid: !!match.match,
-					newRouteDest: tryApplyMatch(routeDest, match),
-				};
+				return getHasFieldPCREMatchResult(has.value, cookie, routeDest);
 			}
 
 			return { valid: cookie !== undefined };
 		}
 		case 'query': {
 			if (has.value !== undefined) {
-				const match = matchPCRE(has.value, url.searchParams.get(has.key));
-				return {
-					valid: !!match.match,
-					newRouteDest: tryApplyMatch(routeDest, match),
-				};
+				return getHasFieldPCREMatchResult(
+					has.value,
+					url.searchParams.get(has.key),
+					routeDest,
+				);
 			}
 
 			return { valid: url.searchParams.has(has.key) };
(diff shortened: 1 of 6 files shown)


emma-cw:
Would be amazing if we could proceed with this PR 🙏  it will help out a lot and should also help with support for @sentry/nextjs

james-elicx:
> Would be amazing if we could proceed with this PR 🙏 it will help out a lot and should also help with support for @sentry/nextjs

Hello @emma-cw, Dario is OOO at the moment and will take a look at this when he gets back 🙂. So, I'm afraid it might take a little longer than usual.

You can of course keep using the prerelease until this is merged though!

packages/next-on-pages/templates/_worker.js/utils/matcher.ts
```diff
@@ -13,34 +17,64 @@ type HasFieldRequestProperties = {
...
+			return { valid: url.hostname === has.value };
 		}
 		case 'header': {
 			if (has.value !== undefined) {
-				return !!headers.get(has.key)?.match(has.value);
+				const match = matchPCRE(has.value, headers.get(has.key));
+				return {
+					valid: !!match.match,
+					newRouteDest: tryApplyMatch(routeDest, match),
+				};
```
dario-piotrowicz, line 32: The current implementation seems to always have this `matchPCRE` + return object code, so there is a bit of code duplication, I've tried to remove the code duplication and ended up with something like this:
```diff
diff --git a/packages/next-on-pages/templates/_worker.js/utils/matcher.ts b/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
index 666ae895..f1e40aed 100644
--- a/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/matcher.ts
@@ -1,4 +1,3 @@
-import type { MatchPCREResult } from './pcre';
 import { applyPCREMatches, matchPCRE } from './pcre';
 
 type HasFieldRequestProperties = {
@@ -17,7 +16,7 @@ type HasFieldRequestProperties = {
  */
 export function hasField(
        has: VercelHasField,
-       { url, cookies, headers, routeDest }: HasFieldRequestProperties,
+       { url, cookies, headers, routeDest }: HasFieldRequestProperties
 ): { valid: boolean; newRouteDest?: string } {
        switch (has.type) {
                case 'host': {
@@ -25,11 +24,12 @@ export function hasField(
                }
                case 'header': {
                        if (has.value !== undefined) {
-                               const match = matchPCRE(has.value, headers.get(has.key));
-                               return {
-                                       valid: !!match.match,
-                                       newRouteDest: tryApplyMatch(routeDest, match),
-                               };
+                               return getHasFieldPCREMatchResult(
+                                       has.value,
+                                       routeDest,
+                                       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                                       headers.get(has.key)!
+                               );
                        }
 
                        return { valid: headers.has(has.key) };
@@ -38,22 +38,18 @@ export function hasField(
                        const cookie = cookies[has.key];
 
                        if (cookie && has.value !== undefined) {
-                               const match = matchPCRE(has.value, cookie);
-                               return {
-                                       valid: !!match.match,
-                                       newRouteDest: tryApplyMatch(routeDest, match),
-                               };
+                               return getHasFieldPCREMatchResult(has.value, routeDest, cookie);
                        }
 
                        return { valid: cookie !== undefined };
                }
                case 'query': {
                        if (has.value !== undefined) {
-                               const match = matchPCRE(has.value, url.searchParams.get(has.key));
-                               return {
-                                       valid: !!match.match,
-                                       newRouteDest: tryApplyMatch(routeDest, match),
-                               };
+                               return getHasFieldPCREMatchResult(
+                                       has.value,
+                                       routeDest,
+                                       url.searchParams.get(has.key) ?? ''
+                               );
                        }
 
                        return { valid: url.searchParams.has(has.key) };
@@ -62,19 +58,33 @@ export function hasField(
 }
 
 /**
- * Try to apply a PCRE match's named capture groups to a destination.
+ * Gets the has field result from a PCRE match
  *
- * @param dest Destination to apply match to.
- * @param match Matched PCRE result.
- * @returns The destination with the match applied, or `undefined` if there was no match.
+ * @param hasValue the value of the has field
+ * @param routeDest original route destination
+ * @param target target string to match
+ * @returns an object indicating if the match is valid and the new route destination if existent
  */
-function tryApplyMatch(
-       dest: string | undefined,
-       { match, captureGroupKeys }: MatchPCREResult,
-): string | undefined {
-       if (dest && match && captureGroupKeys.length) {
-               return applyPCREMatches(dest, match, captureGroupKeys, { namedOnly: true });
+function getHasFieldPCREMatchResult(
+       hasValue: string,
+       routeDest: string | undefined,
+       target: string
+): ReturnType<typeof hasField> {
+       const match = matchPCRE(hasValue, target);
+       let newRouteDest: string | undefined = undefined;
+
+       // if there are capture groups apply them to get a new route destination
+       if (routeDest && match.match && match.captureGroupKeys.length) {
+               newRouteDest = applyPCREMatches(
+                       routeDest,
+                       match.match,
+                       match.captureGroupKeys,
+                       { namedOnly: true }
+               );
        }
 
-       return undefined;
+       return {
+               valid: !!match.match,
+               newRouteDest,
+       };
 }
```

I __think__ something along these lines might be nicer but I am not completely sure, what do you think?

If you prefer the current implementation I'm totally fine to go with it as well 🙂

packages/next-on-pages/templates/_worker.js/utils/matcher.ts
```diff
@@ -13,34 +17,64 @@ type HasFieldRequestProperties = {
  */
 export function hasField(
```
dario-piotrowicz, line 18: Given the new return type I think the name of this function needs to be updated 🤔 
```suggestion
export function checkHasField(
```
(or something along these lines)

dario-piotrowicz: Also, could you please update the tsdoc comment?

the function description can clarify that performs a check and returns the new destination (if it exists) also the return says ``* @returns Whether the request matches the `has` record conditions.`` which is no longer totally correct

packages/next-on-pages/templates/_worker.js/utils/pcre.ts
```diff
@@ -43,9 +47,16 @@ export function applyPCREMatches(
 	rawStr: string,
 	match: RegExpMatchArray,
 	captureGroupKeys: string[],
+	{ namedOnly }: { namedOnly?: boolean } = {},
```
dario-piotrowicz, line 50: could you add this in the tsdoc comment? 👆

dario-piotrowicz (commented):
Thanks a lot @james-elicx 😃 

Sorry I left some minor comments, but nothing major just small naming bits and whatnot to amend 🙂

dario-piotrowicz:
I'm back, thanks a lot @james-elicx for the PR and sorry (both to you and @emma-cw) for the delay in reviewing it 🙇

packages/next-on-pages/templates/_worker.js/utils/matcher.ts
james-elicx: I could have bet money on you replying this!! I aaallllllmmmmoooossstttt made that change earlier this week in anticipation that you might say that. 😅

dario-piotrowicz: sorry for being too predictable 🙈 

anyways as I said it's just a suggestion, I'm totally ok with your solution too

my suggestion also isn't any shorter (I hoped it would be before writing it) so I don't think it's objectively better

dario-piotrowicz (approved):
Awesome! Thanks a lot @james-elicx  😄

PR #450, opened 2023-09-07
Add new `helpers` sub-module with `getRequestExecutionContext`

Karibash:
Add a helper function where ExecutionContext can be retrieved.
If you don't like this API, it's not a problem to close it.

We are currently using it in our project, and it's working well.
https://github.com/looks-to-me/looks-to-me/blob/3b5393ecc95e6e1af3e8bbd949ba5a246a23ec47/app/src/app/(main)/images/posts/%5Bid%5D/route.ts#L60 

This implementation is based on the following pull request for reference.
- #233
- #265

### DIFF
--- a/.changeset/happy-beds-talk.md
+++ b/.changeset/happy-beds-talk.md
@@ -1,0 +1,14 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Add new `helpers` sub-module with `getRequestExecutionContext`.
+
+Add a new `helpers` module that will provide utilities that will allow developer to utilize better next-on-pages and the cloudflare platform.
+
+The `getRequestExecutionContext` function has been introduced with helpers.
+This function provides an execution context object (`ctx`) to the developer.
+
+This function throws an error when executed on the client.
+This function also returns a mocked context when executed in a non pages environment.
+This is useful when executing unit tests, for example, as it prevents errors from occurring.
--- a/package-lock.json
+++ b/package-lock.json
@@ -7152,6 +7152,109 @@
 			"version": "0.0.1",
 			"license": "MIT"
 		},
+		"node_modules/concurrently": {
+			"version": "8.2.1",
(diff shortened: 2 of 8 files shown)


dario-piotrowicz:
Thanks a lot for the PR @Karibash 🙂 

Sorry I haven't updated you in #233, anyways as I mentioned there your solution seems valid but I think we might go a different direction, maybe your solution could be a temporary one if nothing else but it depends on how things proceed with the other alternatives (if we can soon go with the team's preferred alternative we might just go straight to that one)

On our side there aren't many updates (I've also been OOO for a couple of weeks) I'll need to chase the team up and try to get to the bottom of things, sorry for this taking a long time 😓

dario-piotrowicz:
Hi @dihmeetree, sorry no, not yet, I've still waiting to see what happens with https://github.com/cloudflare/workerd/pull/1213

Since, if that runtime API were to be released then any changes in next-on-pages would be unnecessary.

(But the PR has been stale for a while and I am not sure if we are going to land it :sweat:)

Let's wait a bit longer and see what happens.

Also the changes in this PR do not give access to the request object which is also something that users might need access to, so I am not too sure about it :thinking: 

I'll try to speed things up with the team and see if the runtime API is going to be added or not, if it's not let's discuss further here and adjust the API accordingly :slightly_smiling_face: (I also need to check with @GregBrimble as I think he's also interested in this and I think he's considering a potential generic/non next-on-pages spoecific approach).

(PS: so sorry this it taken so long :sweat:)

dario-piotrowicz:
Hi @Karibash, I wanted to let you know, eventually (finally! 😅) we've decided to implement a `getRequestContext` utility (https://github.com/cloudflare/next-on-pages/pull/659) that exposes `ctx` as well as `env` and `cf`.

The utility basically works the exact same way yours does and is implemented in a similar way.

Sorry for going ahead and opening a new PR for it, I hope you don't mind too much 🙇

Please have a look to see if like it 🙂

___

PS: since we've got #659 I'm closing this PR, I hope that's fine, if not please feel free to reopen it and discuss

Issue #453, opened 2023-09-08
[🐛 Bug]: dynamic segments containing an underscore don't work


gabro:
### next-on-pages environment related information

System:
        Platform: darwin
        Arch: x64
        Version: Darwin Kernel Version 22.2.0: Fri Nov 11 02:08:47 PST 2022; root:xnu-8792.61.2~4/RELEASE_X86_64
        CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
        Memory: 32 GB
        Shell: /bin/zsh
Binaries:
        Node: 18.15.0
        Bun: N/A
        pnpm: 8.6.3
        Yarn: 1.22.19
        npm: 9.5.0
Package Manager Used: pnpm
Relevant Packages:
        @cloudflare/next-on-pages: 1.6.0
        vercel: N/A
        next: 13.4.19

### Description

I'm using Next.js 13 with app directory. I set up a dynamic route using the following structure:

```ts
/app/patients/[patient_id]/page.tsx
```

and then `page.tsx` is something along the lines as:

```ts
export default async function PatientPage({ params }: { params: { patient_id: string } }) {
  // do stuff
}
```

Now if I navigate to `/patients/123` and log the params object, I get:

in development, using `next dev`:

```js
{ patient_id: 123 }
```

using `@cloudflare/next-on-pages` (both locally with wrangler or on a deployed env):

```js
{ patient_id: "_id" }
```

I banged my head against this for hours until I realized that avoiding the underscore fixed the issue.

I'm barely familiar with Next, let alone next-on-pages, but I can only assume there's a faulty regex somewhere when parsing the request path.

### Reproduction

I don't have a reproduction repo, unfortunately, but the issue can be consistently reproduced when using `_` in a dynamic segment.

### Pages Deployment Method

Pages CI (GitHub/GitLab integration)

### Would you like to help?

- [ ] Would you like to help fixing this bug?

james-elicx:
Hello, thank you for raising this issue and the additional detail. 🙂

I have opened a PR that should fix this behaviour, and I would welcome you to try out the prerelease available in https://github.com/cloudflare/next-on-pages/pull/454#issuecomment-1712228638.

gabro:
Hey @james-elicx thanks for the ultra fast turnaround on this! I'll give the prerelease a spin tomorrow and report back

PR #454, opened 2023-09-08
fix: dynamic params with underscore not applying correctly

james-elicx:
This PR does the following:
- Accounts for underscores when applying PCRE matches.

NOTE: Vercel does not include the underscore in the [code](https://github.com/vercel/vercel/blob/0a08e4b/packages/cli/src/util/dev/router.ts#L16) that I used as a reference when originally writing this. I am sceptical as to whether we should proceed with this, but as it works in `next dev`, I imagine we should go ahead with it.

fixes https://github.com/cloudflare/next-on-pages/issues/453

### DIFF
--- a/.changeset/old-sheep-visit.md
+++ b/.changeset/old-sheep-visit.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix dynamic parameters with an underscore leading to an incorrect param being provided.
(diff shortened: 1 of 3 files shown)


gabro (approved):
Tested the pre-released locally against the bug in #453 and it works now!

dario-piotrowicz (approved):
thanks a lot @james-elicx 😃

PR #460, opened 2023-09-14
update supported docs

dario-piotrowicz:


packages/next-on-pages/docs/supported.md
```diff
@@ -6,13 +6,11 @@
...
-Based on manual testing, it seems like the latest Next.js version (`13.4.2` at the time of writing) is supported.
+`@cloudflare/next-on-pages` supports all minor and patch version of Next.js 13. We regularly run manual and automated tests to ensure such compatibility.
 
-Earlier and Later versions might be only partially supported, we don't fully know.
+Next.js canary versions are not included.
 
-> The team is very soon going to test various different versions and make improvements to the documentation on which Next.js versions are supported and to what degree.
-
-## Supported Features
+> The edge runtime has been introduced as experimental since [Next.js 12.2](https://nextjs.org/blog/next-12-2#edge-server-rendering-experimental) so versions prior that that are inherently not compatible with the Cloudflare Pages runtime. Moreover the Next.js edge runtime has been significantly experimental early on, that's why we don't recommend using any Next.js version earlier than 13.0.0.
```
IgorMinar, line 13: Do we need to include this paragraph? I think the previous paragraphs seems sufficient to me. Let's be clear about what we support, and anything we don't mentioned is by default unsupported.

packages/next-on-pages/docs/supported.md
```diff
@@ -6,13 +6,11 @@
 
 ## Supported Next.js versions
 
-Based on manual testing, it seems like the latest Next.js version (`13.4.2` at the time of writing) is supported.
+`@cloudflare/next-on-pages` supports all minor and patch version of Next.js 13. We regularly run manual and automated tests to ensure such compatibility.
 
-Earlier and Later versions might be only partially supported, we don't fully know.
+Next.js canary versions are not included.
```
IgorMinar, line 11:
```suggestion
Next.js canary versions not actively being tested and we don't currently extend support to these versions.
```

packages/next-on-pages/docs/supported.md
```diff
@@ -28,12 +26,26 @@ The [Cloudflare workers runtime supports certain Node.js APIs](https://developer
 - `util`
 - `async_hooks`
 
-### External Packages
+## External Packages
 
 You can use any external npm package in your Next.js application, but keep in mind that many packages do rely on Node.js APIs, which, as presented in the previous section could not be supported.
 
+There are also packages which although not requiring node.js APIs don't work with `@cloudflare/next-on-pages` (and/or with Cloudflare's runtime in general) for example those that dynamically evaluate javascript code ([not allowed in the Workers runtime](https://developers.cloudflare.com/workers/runtime-apis/web-standards/#javascript-standards)) or that rely on features not supported by `@cloudflare/next-on-pages`.
```
IgorMinar, line 33: Can we write this a bit more succinctly? Perhaps combine the first paragraph and the second into saying:

You are free to use any external npm package with your Next application as long as it doesn't require unsupported Node.js APIs (see above), or the limited number of JavaScript APIs which Cloudflare disabled due to security concerns (link to https://developers.cloudflare.com/workers/runtime-apis/web-standards/#javascript-standards).

packages/next-on-pages/docs/supported.md
```diff
@@ -28,12 +26,26 @@ The [Cloudflare workers runtime supports certain Node.js APIs](https://developer
...
+
 If you're application is using a package which relies on unsupported Node.js APIs there is often little to be done, generally the only feasible solutions are to either look for an alternative edge-compatible package or see if the current package could be updated in an edge-compatible manner.
```
IgorMinar, line 35: this paragraph could also be shortened to just: "If you currently use an incompatible package, we suggest that you look for alternatives which often do exist on npm".

IgorMinar (approved):
a few suggestions, but overall looks good. thanks for documenting these things.

packages/next-on-pages/docs/supported.md
dario-piotrowicz: no I don't think it's strictly needed, I just thought of sort of providing the reason as to why we don't support earlier Next.js versions, but yeah if you think that this just adds extra fluff (and I can see why) I don't mind removing it 🙂👍

packages/next-on-pages/docs/supported.md
dario-piotrowicz: yes.... but I really wanted to also mention node.js APIs and dynamic evaluation of js code as those are two things that seem to somewhat often trip people over. To keep them around I'll try applying your first suggestion for this if it's ok 🙂 🙏

dario-piotrowicz: @IgorMinar I've applied your suggestion with some small tweaks

I've also removed the paragraph below as in hindsight it does feel as redundant as the Next.js version one we just removed above

Please have another look 🙂 🙏

packages/next-on-pages/docs/supported.md
```diff
@@ -28,11 +24,23 @@ The [Cloudflare workers runtime supports certain Node.js APIs](https://developer
...
 - `async_hooks`
 
-### External Packages
+## External Packages
+
+You are free to use any external npm package with your Next application as long as it doesn't:
+
+- require unsupported Node.js APIs (see above),
+- use [JavaScript APIs](https://developers.cloudflare.com/workers/runtime-apis/web-standards/#javascript-standards) disabled by Cloudflare due to security concerns,
+- nor it relies on features not supported by `@cloudflare/next-on-pages` (see below)
```
IgorMinar, line 33: I think this is a triple negative...  "as long as it doesn't" + "nor" + "not supported".

how about:

"- relies on Next features not suported by `@cloudflare/next-on-pages` (see below)"

IgorMinar (approved):
much better! simpler = better :) 

thanks

packages/next-on-pages/docs/supported.md
dario-piotrowicz: yeah good point 😅 , actually we have `require`, `use` and `relies`... I think this can be simplified further by just putting the verb in the prefix sentence 🙂

PR #462, opened 2023-09-16
only use the Vercel config hack for versions that don't support bun

james-elicx:
Vercel recently introduced proper support for Bun as a package manager in https://github.com/vercel/vercel/pull/10486 and https://github.com/vercel/vercel/pull/10511 (after rejecting my PR claiming they had no plans to support it).

Because of this, it makes sense to remove our logic where we create a custom Vercel config to offer full support for Bun.

Since we are removing this logic, I have added a check to ensure that people are using a version of the Vercel CLI that supports Bun, so that older projects don't suddenly start using the wrong package manager.

### DIFF
--- a/.changeset/lazy-toys-march.md
+++ b/.changeset/lazy-toys-march.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Avoid using the Vercel custom config hack for Bun package manager support when a version of the Vercel CLI that supports Bun is used.
--- a/.changeset/short-falcons-perform.md
+++ b/.changeset/short-falcons-perform.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Remove the Vercel custom config hack for Bun package manager support.
--- a/.changeset/short-falcons-perform.md
+++ b/.changeset/short-falcons-perform.md
@@ -1,5 +1,0 @@
----
-'@cloudflare/next-on-pages': patch
----
-
-Remove the Vercel custom config hack for Bun package manager support.
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,7 +69,6 @@
 		},
 		"node_modules/@ampproject/remapping": {
 			"version": "2.2.1",
-			"dev": true,
 			"license": "Apache-2.0",
 			"peer": true,
 			"dependencies": {
@@ -92,7 +91,6 @@
 		},
 		"node_modules/@babel/compat-data": {
 			"version": "7.22.5",
-			"dev": true,
 			"license": "MIT",
 			"peer": true,
 			"engines": {
@@ -101,7 +99,6 @@
 		},
 		"node_modules/@babel/core": {
 			"version": "7.22.5",
-			"dev": true,
 			"license": "MIT",
 			"peer": true,
 			"dependencies": {
@@ -131,7 +128,6 @@
 		},
 		"node_modules/@babel/core/node_modules/semver": {
 			"version": "6.3.0",
-			"dev": true,
 			"license": "ISC",
 			"peer": true,
 			"bin": {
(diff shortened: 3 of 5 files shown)


dario-piotrowicz (commented):
Interesting that Vercel added support for bun :slightly_smiling_face: 
(not too too suprising given the recent bun hype :smile:)

But I think that this PR introduces a breaking change so unfortunately I don't think we can lend it until the next major next-on-pages version...

The issue is that we currently allow people to use Vercel 30 and up your changes break this for all npm apps (because you update the peer dependencies and that if I remember correctly mismatched peer dependencies make the vercel command we run under the hood fail)

For bun applications specifically we actually even run `process.exit(1)`, so before someone could have been using bun with an older version of the Vercel CLI and your hack but now if they update their next-on-pages version all of a sudden their build would fail...

So I am definitely afraid that this is indeed a breaking change that can break projects and that we need to apply on a major bump.... what do you think? :cry:

james-elicx:
> The issue is that we currently allow people to use Vercel 30 and up your changes break this for all npm apps (because you update the peer dependencies and that if I remember correctly mismatched peer dependencies make the vercel command we run under the hood fail)

Ah, I see, I don't use npm normally, so I wasn't aware of that. In that case I guess I should probably change the peer dep back to how it was before. Maybe this can be the time for me to finally get my wish and change it to `>=30.0.0` instead :D

> For bun applications specifically we actually even run `process.exit(1)`, so before someone could have been using bun with an older version of the Vercel CLI and your hack but now if they update their next-on-pages version all of a sudden their build would fail...
> 
> So I am definitely afraid that this is indeed a breaking change that can break projects and that we need to apply on a major bump.... what do you think? 😢

personally, i dont think this is major bump worthy at all. i view it as a fix since it's changing things to use the proper process of dealing with bun instead of a custom way. If that means someone needs to update their devdeps, they can update them. they have to update their devdeps to to use a new version of next-on-pages anyway.

besides, none of that really matters. Bun support in Pages CI is only on a branch that you need to have your project opted into in the Discord if you want to help with testing it out, so there's only what, like 10 people at the maximum using Bun on Pages, and maybe half of them use Next.js (I'm one of them).

If this is going to be a deal breaker for ya though, i guess we can keep the hack and only use it when the vercel version is less than 32.2.4. lmk and will be happy to make that change.

dario-piotrowicz:
> Ah, I see, I don't use npm normally, so I wasn't aware of that. In that case I guess I should probably change the peer dep back to how it was before. Maybe this can be the time for me to finally get my wish and change it to `>=30.0.0` instead :D

mh... I still am not really sure about `>=30.0.0`, since it's rather permissive and we cannot know how the Vercel cli can change in the future... it does have it's advantages... if you're really set on `>=30.0.0` we can go with it for now :shrug: 

> besides, none of that really matters. Bun support in Pages CI is only on a branch that you need to have your project opted into in the Discord if you want to help with testing it out, so there's only what, like 10 people at the maximum using Bun on Pages, and maybe half of them use Next.js (I'm one of them).

ah well I wasn't following the bun support on Pages CI and didn't remember that it needs to be opted in... you do make a valid point.... however people could still have just npm installed bun in their build command right? :confused:

anyways a breaking change is still a breaking change[^1]... but I agree that it is very unlikely to affect many people, it could indeed likely fly under the radar without actually breaking any user...

keeping the hack in would be the safest option.. but it's a bit cumbersome to keep that lying around until the next major bump... (especially since as we're discussing could not really be needed)...

Let's then keep this a patch and I'll leave it up to you to decide if we want to keep the hack in or not (to be removed on our next major bump) :slightly_smiling_face: 

[^1]: breaking change for me is when someone has an application working fine with next-on-pages and updating the next-on-pages dependency breaks their app and requires them to change something (regardless on whether the change is big or small and involves code changes or not)

dario-piotrowicz (approved):
Awesome, thanks a lot for the changes 🙂

PR #468, opened 2023-09-24
use package-manager-manager library

### LINKED ISSUE #401: [🐛 Bug]: Relevant packages version detection still doesn't work repliably

```
System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:20 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T6000
        CPU: (10) arm64 Apple M1 Pro
        Memory: 32 GB
        Shell: /bin/zsh
Binaries:
        Node: 16.16.0
        Yarn: 1.22.19
        npm: 8.11.0
        pnpm: 8.6.9
Package Manager Used: pnpm
Relevant Packages:
        @cloudflare/next-on-pages: 1.3.1
        vercel: N/A
        next: 13.3.1
```

### Description

In my application I do have Vercel installed by I still get `N/A`, I have also seen this in various opened issues

### Reproduction

Any of my [testing apps](https://github.com/dario-piotrowicz/next-apps-for-testing/tree/master) causes this

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

This should have been solved in https://github.com/cloudflare/next-on-pages/pull/264 but it still seems like it's not working reliably 😓 

### Would you like to help?

### LINKED ISSUE #467: [🐛 Bug]: Command does not work with monorepos x yarn berry
Using version `@cloudflare/next-on-pages@1.6.1`
```
~/sites/edge-compute-poc$ yarn web next-on-pages --info
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: ENOENT: no such file or directory, open '.yarnrc.yml'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '.yarnrc.yml'
}
```

### Description

Currently `next-on-pages` assumes the `.yarnrc.yml` is always present in directory from where the script is ran. [See here](https://github.com/cloudflare/next-on-pages/blob/main/packages/next-on-pages/src/buildApplication/packageManagerUtils.ts#L43)

When having a monorepo (for example with Turborepo) we have the following structure:
```
- apps
 - web
  - {nextjs project}
- .yarnrc.yml
```

This will not work since the script is executed in `apps/web` and currently the file is 2 directories higher.
As a dirty hack, changing the patch to `../../.yarnrc.yml` makes the script run successfully.

### Reproduction

1. Create a monorepo using `npx create-turbo@latest`
2. Select `yarn` as package manager
3. Go into the created project folder
4. Run `yarn set version 3.3.1`
5. Run `yarn workspace web dlx @cloudflare/next-on-pages@1` to run the cloudflare pages build script
6. Be greeted with the following error ```⚡️ Watch mode unexpected error: Error: ENOENT: no such file or directory, open '.yarnrc.yml'```

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

I used `yarn patch @cloudflare/next-on-pages`, then follow the instructions in the terminal. The file you need to look for is the `dist/index.js` and then search for `".yarnrc.yml"` and change it to `"../../.yarnrc.yml"`.

After changing, "commit" the yarn patch as the terminal describes it and run `yarn install` again. You will now see the commands correctly work.

### Would you like to help?

dario-piotrowicz:
This PR takes out the package managers related logic present in next-on-pages and instead uses this [package-manager-manager](https://github.com/james-elicx/package-manager-manager) library which me and @james-elicx have been developing.

The library is focused on package managers so it has a smaller scope and has a number of unit tests and extra logic around package managers (for example handling monorepos which next-on-pages currently does not) so besides removing extra code from next-on-pages this should help making things work more correctly/reliably.

resolves #467
resolves #401
(also addresses usage of next-on-pages in monorepos/turborepo related to #304)


dario-piotrowicz:
Note:
Screenshot 2023-09-24 at 23 18 11

This doesn't seem to fix #468 (it's probably an incompatibility with pnpm monorepos)

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
```diff
@@ -1,16 +1,11 @@
...
-import type { PackageManager } from './packageManagerUtils';
-import {
-	getCurrentPackageManager,
-	getExecStr,
-	getPackageManagerInfo,
-	getPackageVersion,
-	waitForProcessToClose,
-} from './packageManagerUtils';
+import { getPackageManager } from 'package-manager-manager';
+import type { PackageManager } from 'package-manager-manager';
```
petebacondarwin, line 7: NIT: feels like these imports should be higher up the file now?

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
```diff
@@ -27,20 +22,24 @@ import {
  *
  */
 export async function buildVercelOutput(): Promise<void> {
-	const pm = await getCurrentPackageManager();
-	cliLog(`Detected Package Manager: ${pm}\n`);
+	const pm = await getPackageManager();
+	if (!pm) {
+		cliError('Error: Could not detect current package manager in use');
+		process.exit(1);
```
petebacondarwin, line 28: I generally try to avoid using `process.exit()` deep within code. It makes it difficult to test failure scenarios, as it is hard to mock out and can actually kill the test runner as well as the test subject.
I prefer to use exceptions to bubble up and cause a fatal error towards the top of the application.

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
```diff
@@ -27,20 +22,24 @@ import {
...
 
-	if (pm === 'bun') {
+	if (pm.name === 'bun') {
 		// Vercel introduced proper Bun support in 32.2.1 and 32.2.4 (for monorepos), therefore we should
 		// ensure the Vercel CLI has a config file telling it to use Bun for older versions. This is done
 		// to prevent a breaking change for users who are using an older version of the Vercel CLI.
-		const vercelVersion = await getPackageVersion('vercel', pm);
-		if (vercelVersion && vercelVersion < '32.2.4') {
+		const vercelInfo = await pm.getPackageInfo('vercel');
+		if (vercelInfo && vercelInfo.version < '32.2.4') {
```
petebacondarwin, line 42: This string comparison of semver strings is prone to problems. For example `"32.12.4" < "32.2.4"`.
Better to use a `semver` library to parse and compare.

petebacondarwin (approved):
Love the clean up - and that you have already contributed to the library to make sure it is in good shape for us.
A few minor concerns but generally approve.

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: I totally agree, [we do use semver](https://github.com/cloudflare/next-on-pages/blob/628f6bbbfa8251b703de273f2563567c467f01b8/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts#L1) and we should (have) here as well, sorry I'm not sure how we missed this 😓 , thanks a lot for noticing it 🙂

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: good points, sure I don't mind changing this, thanks 🙂

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: yeah thanks I'll move them 🙂

(as a followup we can add the eslint rule for the imports ordering 🤔)

dario-piotrowicz: done :)

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: I've updated this, thanks to your suggestion I noticed that we could actually pass the package manager from the caller (which already needed it anyways), so I moved the check there instead 🙂

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: updated 🙂

Issue #470, opened 2023-09-24
[🐛 Bug]:  Request fails when asset name start with 500 and 404


divyam234:
### Description

This happens only when  if any asset contains name which start with error status codes .For example  
 ```http://abc.com/files/500-abc.pdf```  this request fails. I recon it gets confused b/w error status pages.

### Reproduction

Use above route pattern will reproduce bug.

### Pages Deployment Method

None

### Would you like to help?

- [ ] Would you like to help fixing this bug?

dario-piotrowicz:
@divyam234 thanks so much for the issue 🙂 

it does look bizarre I'll have a look 👍 

But anyways one quick question.... it looks to me like this only happens with `500`s.... are you sure you've seen it with other numbers as well?

![Screenshot 2023-09-27 at 23 42 54](https://github.com/cloudflare/next-on-pages/assets/61631103/35c79977-2e90-4314-8412-fbf81f796859)

dario-piotrowicz:
Correction, I think it only applies to `500`s and `404`s 
![Screenshot 2023-09-27 at 23 49 05](https://github.com/cloudflare/next-on-pages/assets/61631103/fc093d86-83b9-46d5-9948-32e43a9290ad)

divyam234:
Yeah 500 and 404

PR #473, opened 2023-09-28
add $ to all route srcs

dario-piotrowicz:
issue #470 happens because Vercel generated route srcs don't contain appropriate `$` at the end, if can be fixed by adding ad-hoc `$` as we've done in the past, I am however trying to see, can we just always add `$`? is that how Vercel interprets such regexes anyways?

### DIFF
--- a/packages/next-on-pages/src/buildApplication/getVercelConfig.ts
+++ b/packages/next-on-pages/src/buildApplication/getVercelConfig.ts
@@ -84,7 +84,15 @@ export function processVercelConfig(
 
 	let currentPhase: VercelHandleValue | 'none' = 'none';
 	config.routes?.forEach(route => {
-		if (route.src) {
+		// The following src tweaks apply only on non-handler routes
+		// see: https://github.com/vercel/vercel/blob/ea5bc88/packages/routing-utils/src/index.ts#L58
+		const isHandler = isVercelHandler(route);
+
+		// routes pointing to '/' with a locale might need not to be tweaked (as suggested by one of our i18n tests)
+		// so just to be on the safe side we also skip those (note that this might actually be unnecessary)
+		const isLocaleIndex = !isHandler && !!(route.locale && route.src === '/');
+
+		if (route.src && !isHandler && !isLocaleIndex) {
 			// Route src should always start with a '^'
 			// see: https://github.com/vercel/vercel/blob/ea5bc88/packages/routing-utils/src/index.ts#L77
 			if (!route.src.startsWith('^')) {
--- a/packages/next-on-pages/src/buildApplication/getVercelConfig.ts
+++ b/packages/next-on-pages/src/buildApplication/getVercelConfig.ts
@@ -86,7 +86,7 @@ export function processVercelConfig(
 	config.routes?.forEach(route => {
 		// Vercel output routes sometimes include `$`s and sometimes they do not, but it seems
 		// like in either case Vercel behaves as if they're present, so we need to mimic such behavior
-		if(route.src && !route.src.endsWith('$')) {
+		if (route.src && !route.src.endsWith('$')) {
 			route.src += '$';
 		}
 
--- a/packages/next-on-pages/src/buildApplication/getVercelConfig.ts
+++ b/packages/next-on-pages/src/buildApplication/getVercelConfig.ts
@@ -70,7 +70,7 @@ export function processVercelConfig(
 	config: VercelConfig,
 ): ProcessedVercelConfig {
 	const processedConfig: ProcessedVercelConfig = {
-		...config,
+		...structuredClone(config),
 		routes: {
 			none: [],
 			filesystem: [],
@@ -84,11 +84,10 @@ export function processVercelConfig(
 
 	let currentPhase: VercelHandleValue | 'none' = 'none';
 	config.routes?.forEach(route => {
-		// Vercel generates `^/` as the regex instead of `^/$` for their root prefer route.
-		// This makes the route match everything and not only the actual root requests, so
-		// here we need to fix such erroneous regex.
-		if (route.src === '^/' && route.dest === '/index.prefetch.rsc') {
-			route.src = '^/$';
+		// Vercel output routes sometimes include `$`s and sometimes they do not, but it seems
+		// like in either case Vercel behaves as if they're present, so we need to mimic such behavior
+		if(route.src && !route.src.endsWith('$')) {
+			route.src += '$';
 		}
 
 		if (isVercelHandler(route)) {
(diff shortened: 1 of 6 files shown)


james-elicx:
For reference, here is Vercel's function to normalize source routes. Looks like this should be safe to do.

https://github.com/vercel/vercel/blob/ea5bc8806276abf5ba14bdb4a966267497e5d14d/packages/routing-utils/src/index.ts#L76-L85

image

packages/next-on-pages/src/buildApplication/getVercelConfig.ts
```diff
@@ -84,11 +84,15 @@ export function processVercelConfig(
...
-		// here we need to fix such erroneous regex.
-		if (route.src === '^/' && route.dest === '/index.prefetch.rsc') {
-			route.src = '^/$';
+		// Vercel output routes sometimes include `$`s and sometimes they do not, but it seems
+		// like in either case Vercel behaves as if they're present, so we need to mimic such behavior
+		if (
+			route.src &&
+			!route.src.endsWith('$') &&
+			route.src !== '/' &&
+			!route.src.endsWith('(.*)')
```
dario-piotrowicz, line 93: Is this needed?

dario-piotrowicz:
> For reference, here is Vercel's function to normalize source routes. Looks like this should be safe to do. Makes me wonder if there's a bug in their CLI somewhere.
> 
> https://github.com/vercel/vercel/blob/ea5bc8806276abf5ba14bdb4a966267497e5d14d/packages/routing-utils/src/index.ts#L76-L85
> 
> image

Out e2es and [next-apps-for-testing](https://github.com/dario-piotrowicz/next-apps-for-testing/actions/runs/6339870135) are looking promising, but I needed to add [more tweaks](https://github.com/cloudflare/next-on-pages/pull/473#discussion_r1340208608) to get our unit tests to pass, I need to investigate wether these are for some reason necessary or if our unit tests are checking things wrongly (but quickly looking at the tests nothing particularly stands out)

dario-piotrowicz:
Regarding the e2e changes I applied (which look unrelated to the rest of this PR), the e2es started failing because of my changes and by looking into it, it actually looks like my changes are fixing an inconsistency we had with Next and that the tests we had were basically checking the wrong behaviour 😮 

By running the testing app router app that throws on `/500-error` this is what we get with `next start`:
Screenshot 2023-10-03 at 00 09 40

And that is also what we __now__ get with next-on-pages.

But previously in next-on-pages we were getting the following:
Screenshot 2023-10-03 at 00 10 19

And the e2es were expecting that, so I needed to update them so actually expect the correct error page instead.

I do believe the new behaviour to be the correct one (as whatever happens on `next start` should be our source of truth)

> **Note**
> The `500- | Internal Server Error` page is actually the error page we get in the pages router, so what was happening before was that even though we were in the app router we were serving the pages router error page

packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -415,13 +415,16 @@ export class RoutesMatcher {
 		}
 
 		const isLocaleIndex =
-			/^\//.test(route.src) && route.src.slice(1) in this.locales;
+			/^\/$/.test(route.src) && route.src.slice(1) in this.locales;
```
petebacondarwin, line 418: Is this not the same as:

```suggestion
			route.src === "/" && "/" in this.locales;
```

petebacondarwin (approved):

packages/next-on-pages/templates/_worker.js/routes-matcher.ts
dario-piotrowicz: yes thanks 🙂, I missed this when updating the regex

Although `"/" in this.locales` is always going to return false since locales contains the actual locales (not paths)

So I might just completely remove this check I think 😕 🤔

dario-piotrowicz: Ah no ok, this is something I already looked into and tweaked before (we were adding `^` and `$` to specific cases here, but that's no longer needed as we now do it for all routes)

This code was already a noop but I forgot to remove it, thanks a lot for pointing it out! 🙏

dario-piotrowicz:
> **Note**
> this PR should be ready to merge but the change is rather broad, so I want to do a bunch of manual testing before merging it just to be extra sure

dario-piotrowicz:
Manual (brief/smoke-test) testing using the apps in the [next-apps-for-testing repo](https://github.com/dario-piotrowicz/next-apps-for-testing) 

(deployments taken from this [workflow](https://github.com/dario-piotrowicz/next-apps-for-testing/actions/runs/6398021014))

| Result | App | Url |
|--------|--------|--------|
| ? | `app-dir-not-founds` | https://fc181054.next-apps-for-testing.pages.dev/ |
| ? | `app-playground-edge` | https://0569e5fb.next-apps-for-testing.pages.dev/ |
| ? | `config-rewrites-redirects-headers-13.2.4` | https://7631f674.next-apps-for-testing.pages.dev/ |
| ? | `dynamic-static-app-dir-13.2.4` | https://3b5a28cf.next-apps-for-testing.pages.dev/ |
| ? | `empty-getstaticpaths-pages-13.4.4` | https://dd89c9fa.next-apps-for-testing.pages.dev/ |
| ? | `i18n-next-intl-13.4.1` | https://9f93b2e9.next-apps-for-testing.pages.dev/ |
| ? | `i18n-next-translate-13.4.1` | https://b489b4c5.next-apps-for-testing.pages.dev/ |
| ? | `i18n-pages-dir-13.4.1` | https://6b9d0e9f.next-apps-for-testing.pages.dev/ |
| ? | `kv-app-dir-13.4.4` | https://613d7375.next-apps-for-testing.pages.dev/ |
| ? | `middleware-13.2.4` | https://616957cf.next-apps-for-testing.pages.dev/ |
| ? | `simple-app-dir-13.0.0` | https://1b6f32d5.next-apps-for-testing.pages.dev/ |
| ? | `simple-app-dir-13.2.4` | https://5a3d9f45.next-apps-for-testing.pages.dev/ |
| ? | `simple-app-dir-13.3.1` | https://f6b6f558.next-apps-for-testing.pages.dev/ |
| ? | `simple-app-dir-13.4.1-server-actions` | https://3906a391.next-apps-for-testing.pages.dev/ |
| ? | `simple-app-dir-13.4.2` | https://d2e75644.next-apps-for-testing.pages.dev/ |
| ? | `simple-app-dir-13.4.2-images` | https://18403b4a.next-apps-for-testing.pages.dev/ |
| ? | `simple-pages-dir-13.2.4` | https://4fb77e82.next-apps-for-testing.pages.dev/ |
| ? | `simple-pages-dir-13.4.2` | https://9c0a6cac.next-apps-for-testing.pages.dev/ |
| ? | `simple-src-13.2.4` | https://9b44c8bf.next-apps-for-testing.pages.dev/ |
| ? | `simple-wasm-pages-13.3.1` | https://b9c47dc0.next-apps-for-testing.pages.dev/ |

dario-piotrowicz:
Based on the above comment I am sufficiently confident in the change to merge it 🙂

PR #478, opened 2023-10-01
Use pmm again

### LINKED ISSUE #467: [🐛 Bug]: Command does not work with monorepos x yarn berry

dario-piotrowicz:
fixes #467


packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
```diff
@@ -132,7 +132,7 @@ async function runVercelBuild(
 	additionalArgs: string[] = [],
 ): Promise<void> {
 	if (pm.name === 'yarn' && pm.version.startsWith('1.')) {
-		const vercelInfo = await pm.getPackageInfo('vercel');
+		const vercelInfo = await pm.getPackageInfo('vercel').catch(() => null);
```
petebacondarwin, line 135: It feels like it would be good to wrap this call+catch into a helper function, to avoid having to remember to do it everywhere that `pm.getPackageInfo()` is used, no?

petebacondarwin (approved):

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: Thanks! good call, I've added a helper function for that 🙂👍

PR #479, opened 2023-10-02
Introduce no nodejs compat flag error page

dario-piotrowicz:
@WalshyDev has pointed out to me the fact that our runtime error message: `Error: Could not access built-in Node.js modules. Please make sure that your Cloudflare Pages project has the 'nodejs_compat' compatibility flag set.` is probably not clear and developers don't get enough instructions on what to do, so to improve things I've updated next-on-pages to serve a more descriptive and actionable error page in case of this runtime issue.

> when we'll have configs for Pages projects I think this should become a build time error instead of a runtime one, but for now we can only deal with this at runtime 🥲 

Result: https://cede2e83.next-nodejs-compat-page-test.pages.dev/

Screenshot 2023-10-04 at 22 51 13

> because of the lack of better ideas I tried to mimic the error page developers get when a worker throws: https://f6e0ff92.next-nodejs-compat-page-test.pages.dev/


packages/next-on-pages/templates/_worker.js/index.ts
```diff
@@ -21,10 +21,13 @@ export default {
 	async fetch(request, env, ctx) {
 		const envAsyncLocalStorage = await __ENV_ALS_PROMISE__;
 		if (!envAsyncLocalStorage) {
-			return new Response(
-				`Error: Could not access built-in Node.js modules. Please make sure that your Cloudflare Pages project has the 'nodejs_compat' compatibility flag set.`,
-				{ status: 503 },
+			const noNodeJsCompatStaticPageRequest = await fetch(
+				'https://no-nodejs-compat-flag-static-error-page.pages.dev/',
```
WalshyDev, line 25: I understand why we do this but i also dislike it. I can't think of anything better though so it's cool.

WalshyDev (approved):

packages/next-on-pages/templates/_worker.js/index.ts
dario-piotrowicz: yeah.... it's kinda dirty... but I doubt people would be too happy to carry an extra (unnecessary) html + assets in their applications for no good reason 😅

let's hope that this is anyways a temporary patch until we can fail at build time (so hopefully we will remove this runtime check altogether in a future major)

internal-packages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,246 @@
...
+						>function settings page</a
+					>.
+				</p>
+
+				<p>
+					Then in the compatibility flags section for both the production and
+					preview deployments type <code> nodejs_compat </code> in the input box
+					and click on save to apply the changes:
+					<img
+						src="https://no-nodejs-compat-flag-static-error-page.pages.dev/img/input.png"
```
petebacondarwin, line 217: Note that you had to use the full URL here including host because this file is being rendered inside the broken next-on-pages deployment. But this means that if you decide to deploy it somewhere else, you need to remember to update these.
I wonder if you could reduce the duplication by setting a `base[href]` in the page?

petebacondarwin (approved):

internal-packages/no-nodejs-compat-flag-static-error-page/assets/index.html
james-elicx: What about hosting the image in this GitHub repo and using the raw link to it @dario-piotrowicz?

dario-piotrowicz: oh... yeah I didn't think of that! 😓

sounds like a good idea 🙂, thanks! 😄

dario-piotrowicz: applied, thanks @james-elicx 😄 

(PS: unfortunately I cannot test this change until I actually merge this PR 😓)

GregBrimble (changes requested):
The page we're using here has quite a few errors on it. I'd personally in-line it in the project so it's versioned with the package, but at the very least, we need to review the content of the page.

dario-piotrowicz:
> The page we're using here has quite a few errors on it. I'd personally in-line it in the project so it's versioned with the package, but at the very least, we need to review the content of the page.

yeah good point about including it in the output app itself, I didn't want to include the html in the project's assets but by chatting with @petebacondarwin I understood that the impact it has on the app assets limits is very negligible, so I'm including it there and not having an external deployed app for this.

Regarding the errors on the page, I did notice that once I swapped `compat` and `nodejs` 🤦 I've fixed that, thanks for pointing that out, besides I can't spot other problems there, could you let me know what other errors you saw there? 😓

internal-packages/no-nodejs-compat-flag-static-error-page/assets/index.html
dario-piotrowicz: Sorry @james-elicx after chats with Pete and the comment from Greg I'm going to actually include the html in the user's app, it should practically have no impact on the application 🙂

packages/next-on-pages/src/buildApplication/buildApplication.ts
```diff
@@ -145,3 +151,27 @@ async function prepareAndBuildWorker(
...
+	const noNodejsCompatFlagStaticErrorPagePath = join(
+		__dirname,
+		'..',
+		'no-nodejs-compat-flag-static-error-page',
+		'dist',
+		'index.html',
+	);
+	await cp(
+		noNodejsCompatFlagStaticErrorPagePath,
+		join(outputDir, htmlFileName),
```
GregBrimble, line 174: I'd consider putting this into `${outputDir}/cdn-cgi/errors/no-nodejs_compat.html` or something like that rather than just in the root of the build output directory, just so it's tucked away in something we clearly own.

pages-e2e/features/appNoNodeJsCompatFlag/no-nodejs_compat-flag.test.ts
```diff
@@ -0,0 +1,38 @@
+import { getAssertVisible } from '@features-utils/getAssertVisible';
+import { describe, test } from 'vitest';
+
+describe('no nodejs_compatibility flag set', () => {
```
GregBrimble, line 4: So cool to see these tests in use!

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+			#more-details-btn:hover {
+				text-decoration: underline;
+			}
+		</style>
+	</head>
+	<body>
+		<div>
+			<header>
+				<h1>
+					<span class="error">NodeJS Compatibility Error</span>
```
GregBrimble, line 158:
```suggestion
					<span class="error">Node.JS Compatibility Error</span>
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+					<span class="error-description"
+						>no <code>nodejs_compat</code> flag set</span
```
GregBrimble, line 160:
```suggestion
						>no <code>nodejs_compat</code> compatibility flag set</span
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+						>, but hasn't been properly configured.
+					</p>
+				</div>
+
+				<div class="what-can-i-do">
+					<h2>What can I do?</h2>
+					<p>
+						<strong>If you are the owner of this website:</strong>
+						<br />
+						You should go to the Pages Project's
```
GregBrimble, line 183:
```suggestion
						You should go to the Pages project's
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+						<a
+							href="https://dash.cloudflare.com/?to=/:account/pages/view/:pages-project/settings/functions#compatibility_flags_section"
+							target="_blank"
+							>compatibility flags settings section</a
+						>
+						and add the <code>nodejs_compat</code> flag to both your production
+						and preview deployments.
```
GregBrimble, line 190:
```suggestion
						and preview environments.
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+				<div class="what-can-i-do">
+					<h2>What can I do?</h2>
+					<p>
+						<strong>If you are the owner of this website:</strong>
+						<br />
+						You should go to the Pages Project's
+						<a
+							href="https://dash.cloudflare.com/?to=/:account/pages/view/:pages-project/settings/functions#compatibility_flags_section"
+							target="_blank"
+							>compatibility flags settings section</a
```
GregBrimble, line 187:
```suggestion
							>Compatibility Flags settings section</a
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+				<p>
+					In the
+					<a href="https://dash.cloudflare.com/" target="_blank"
+						>Cloudflare dashboard</a
+					>
+					go to your Pages project and navigate to the
+					<a
+						href="https://dash.cloudflare.com/?to=/:account/pages/view/:pages-project/settings/functions"
+						target="_blank"
+						>function settings page</a
```
GregBrimble, line 212:
```suggestion
						>Functions settings page</a
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+					preview deployments type <code>nodejs_compat</code> in the input box
+					and click on save to apply the changes:
+					<img
+						src="./img/input.png"
+						alt="screenshot showing how to input compatibility flags"
+					/>
+				</p>
+
+				<p>
+					At the end of the operation this is how your compatibility flags
```
GregBrimble, line 227:
```suggestion
					At the end of the operation this is how your Compatibility Flags
```

packages/next-on-pages/no-nodejs-compat-flag-static-error-page/assets/index.html
```diff
@@ -0,0 +1,250 @@
...
+					section should look like:
+					<img
+						src="./img/result.png"
+						alt="screenshot showing how to compatibility flags should be after having been set"
+					/>
+				</p>
+
+				<p>
+					<strong> Important! </strong>
+					After setting the compatibility flags you need to re-deploy your
```
GregBrimble, line 237:
```suggestion
					<strong>Important!</strong>
					After setting the Compatibility Flags, you need to re-deploy your
```

GregBrimble (approved):
Just some spelling/grammar nits. Thanks for shipping the page with the package directly rather than linking out!

packages/next-on-pages/src/buildApplication/buildApplication.ts
dario-piotrowicz: yeah I totally agree that we should make it clear that it's something from us, I thought it was clear enough given the file name, but yeah putting it into `cdn-cgi` (which, since you're suggesting it I guess is allowed/doesn't generate conflicts/issues) might just be cleaner simpler, I'll look into that, thanks! 😄

GregBrimble: https://developers.cloudflare.com/fundamentals/reference/cdn-cgi-endpoint/

This will actually mean that this will be inaccessible for anyone trying to directly access this asset and it can effectively only retrieved by the Pages Functions Worker :)

PR #496, opened 2023-10-07
remove 'watch mode' from unexpected error message

dario-piotrowicz:
pretty minor stuff but I noticed that this error says `watch mode` even though it a generic error that also applies in non-watch mode


dario-piotrowicz:
Closing in favour of #553

PR #509, opened 2023-10-19
bump package-manager-manager version

dario-piotrowicz:


packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
```diff
@@ -133,7 +133,7 @@ async function runVercelBuild(
 	pm: PackageManager,
 	additionalArgs: string[] = [],
 ): Promise<void> {
-	if (pm.name === 'yarn' && pm.version.startsWith('1.')) {
+	if (pm.name === 'yarn' && pm.version?.startsWith('1.')) {
```
petebacondarwin, line 136: Could we use `isYarnClassic()` function here or is it not available outside the package?

petebacondarwin: Oh I see it is not exposed. Shame.

petebacondarwin (approved):

packages/next-on-pages/src/buildApplication/buildVercelOutput.ts
dario-piotrowicz: I can expose it 🙂 

anyways if it's ok let's do it as a followup as anyways the current patch should help but it doesn't fix the whole issue (something is probably still a bit off with the version detection, I'll have to look more deeply into it and probably do a new bump in a few days anyways 🙂)

PR #517, opened 2023-10-24
fix: several issues with `next.config.js` `basePath` option

### LINKED ISSUE #516: [🐛 Bug]: Setting nextjs `basePath` breaks next-on-pages build

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:19 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6020
        CPU: (12) arm64 Apple M2 Max
        Memory: 64 GB
        Shell: /bin/zsh
Package Manager Used: pnpm (8.9.2)

Relevant Packages:
        @cloudflare/next-on-pages: 1.7.2
        vercel: 32.5.0
        next: 13.5.6

## Actual
As per the title, changing `apps/simple-app-dir-13.5.6/next.config.js` to

```
const nextConfig = {
  basePath: "/foo",
};
```

Results in borked build:

```
⚡️ Invalid prerender config for /foo
⚡️ Invalid prerender config for /foo.rsc

⚡️ ERROR: Failed to produce a Cloudflare Pages build from the project.
⚡️ 
⚡️      The following routes were not configured to run with the Edge Runtime:
⚡️        - /foo/_not-found
⚡️        - /foo
```

## Expected
Builds fine

### Reproduction

Reproduced on [next-apps-for-testing](https://github.com/dario-piotrowicz/next-apps-for-testing.git) repo as per the steps above.

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
This PR fixes the following issues relating to the `next.config.js` `basePath` option:
- infinite redirect for prerendered base path root.
- base path root index.rsc not generating a non-index rsc override.
- _not-found function hack not working with base paths.
- not discard an invalid unnecessary base path function at the root with a valid alternative.

fixes #516

### DIFF
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
@@ -10,7 +10,10 @@ import type { CollectedFunctions, FunctionInfo } from './configs';
 import { join } from 'path';
 import type { ProcessVercelFunctionsOpts } from '.';
 
-type Opts = Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'vercelConfig'>;
+type InvalidFunctionsOpts = Pick<
+	ProcessVercelFunctionsOpts,
+	'functionsDir' | 'vercelConfig'
+>;
 
 /**
  * Checks if there are any invalid functions from the Vercel build output.
@@ -25,7 +28,7 @@ type Opts = Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'vercelConfig'>;
  */
 export async function checkInvalidFunctions(
 	collectedFunctions: CollectedFunctions,
-	opts: Opts,
+	opts: InvalidFunctionsOpts,
 ): Promise<void> {
 	await tryToFixNotFoundRoute(collectedFunctions);
 
@@ -116,10 +119,7 @@ async function tryToFixNotFoundRoute({
  */
 async function tryToFixI18nFunctions(
 	{ edgeFunctions, invalidFunctions, ignoredFunctions }: CollectedFunctions,
-	{
-		vercelConfig,
-		functionsDir,
-	}: Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'vercelConfig'>,
+	{ vercelConfig, functionsDir }: InvalidFunctionsOpts,
 ): Promise<void> {
 	if (!invalidFunctions.size || !vercelConfig.routes?.length) {
 		return;
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
@@ -7,8 +7,10 @@ import {
 	stripFuncExtension,
 } from '../../utils';
 import type { CollectedFunctions, FunctionInfo } from './configs';
-import { join, resolve } from 'path';
+import { join } from 'path';
 import type { ProcessVercelFunctionsOpts } from '.';
+
+type Opts = Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'vercelConfig'>;
 
 /**
  * Checks if there are any invalid functions from the Vercel build output.
@@ -23,11 +25,13 @@ import type { ProcessVercelFunctionsOpts } from '.';
  */
 export async function checkInvalidFunctions(
 	collectedFunctions: CollectedFunctions,
-	opts: Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'vercelConfig'>,
+	opts: Opts,
 ): Promise<void> {
 	await tryToFixNotFoundRoute(collectedFunctions);
 
 	await tryToFixI18nFunctions(collectedFunctions, opts);
+
+	await tryToFixInvalidFuncsWithValidIndexAlternative(collectedFunctions);
 
 	if (collectedFunctions.invalidFunctions.size > 0) {
 		await printInvalidFunctionsErrorMessage(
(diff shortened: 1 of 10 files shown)


packages/next-on-pages/src/buildApplication/processVercelFunctions/prerenderFunctions.ts
```diff
@@ -13,6 +13,8 @@ import {
 } from '../../utils';
 import { cliWarn } from '../../cli';
 
+type Opts = Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'outputDir'>;
```
dario-piotrowicz, line 16: not a huge deal but I wonder if we could have a meaningful name for this type instead of `Opts` 🤔

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -7,9 +7,11 @@ import {
 	stripFuncExtension,
 } from '../../utils';
 import type { CollectedFunctions, FunctionInfo } from './configs';
-import { join, resolve } from 'path';
+import { join } from 'path';
 import type { ProcessVercelFunctionsOpts } from '.';
 
+type Opts = Pick<ProcessVercelFunctionsOpts, 'functionsDir' | 'vercelConfig'>;
```
dario-piotrowicz, line 13: as for the other `Opts` type I wonder if we could find a better more meaningful name for the type 😢

Also I think you're forgot to use it below:
Screenshot 2023-10-27 at 12 23 53

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -192,3 +216,37 @@ async function printInvalidFunctionsErrorMessage(
 		{ spaced: true },
 	);
 }
+
+/**
+ * Tries to fix potential invalid functions with a valid /index alternative from the Vercel build
+ * output.
+ *
+ * This deals with an edge case when using `basePath` and it creates an invalid function for the
+ * `/` route, but has a valid alternative at `/index`.
```
dario-piotrowicz, line 225:
```suggestion
 * This deals with an edge case when using `basePath` creates an invalid function for the
 * `/` route, but a valid alternative is created at `/index`.
```

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -55,28 +59,48 @@ export async function checkInvalidFunctions(
...
+	invalidFunctions,
+	ignoredFunctions,
+}: CollectedFunctions): Promise<void> {
+	for (const [fullPath, fnInfo] of invalidFunctions.entries()) {
+		const notFoundFuncName = '/_not-found.func';
+		const errorFuncName = '/_error.func';
 
-	const invalidNotFound = collectedFunctions.invalidFunctions.get(notFoundDir);
-	const invalidError = collectedFunctions.invalidFunctions.get(errorDir);
+		const invalidNotFound = fullPath.endsWith(notFoundFuncName);
```
dario-piotrowicz, line 70: I _think_ that the `_not-found.func` hack is for not-found functions at the root, so I'm not sure if checking this with simply `endsWith` is correct.

I have to have a think about it 🤔

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -192,3 +216,37 @@ async function printInvalidFunctionsErrorMessage(
...
+/**
+ * Tries to fix potential invalid functions with a valid /index alternative from the Vercel build
+ * output.
+ *
+ * This deals with an edge case when using `basePath` and it creates an invalid function for the
+ * `/` route, but has a valid alternative at `/index`.
+ *
+ * @param collectedFunctions Collected functions from the Vercel build output.
+ */
+async function tryToFixInvalidFuncsWithValidIndexAlternative({
```
dario-piotrowicz, line 229: this applies to all possible invalid funcs and not just the one at the root... I wonder if that's ok 🤔

packages/next-on-pages/templates/_worker.js/utils/http.ts
```diff
@@ -89,7 +89,7 @@ export function createRouteRequest(req: Request, path: string) {
 	applySearchParams(newUrl.searchParams, new URL(req.url).searchParams);
 
 	newUrl.pathname = newUrl.pathname
-		.replace(/^\/index.html$/, '/')
+		.replace(/\/index.html$/, '/')
```
dario-piotrowicz, line 92: as before, I wonder if not forcing this to apply at the root is the correct call 🤔

dario-piotrowicz (approved):

dario-piotrowicz:
(sorry approved by mistake I only meant to comment 😅)

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
james-elicx: Well, when using a base path, it creates one inside the base path, so....

packages/next-on-pages/templates/_worker.js/utils/http.ts
james-elicx: i would direct your complaints to the Pages team :)

https://developers.cloudflare.com/pages/platform/serving-pages/#route-matching

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
james-elicx: Hmm. Good question. I *think* it should be okay. Like, if you send a request to `/home`, that would be equivalent to sending a request to `/home/index.html` when using html files, so i think it should be okay to apply the same logic to invalid functions to try and fix them.

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
dario-piotrowicz: but I think that once we were trying to be smart about it and tried to detect when a base path was being used? (but checking all the routes and see if they have the same prefix?), did we scrap that? if not, could it be used here? :thinking:

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
dario-piotrowicz: ...maybe :thinking:

packages/next-on-pages/templates/_worker.js/utils/http.ts
dario-piotrowicz: ah! you mean this?

Screenshot at 2023-10-29 00-14-38


yeah good point :thinking:

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
james-elicx: > but I think that once we were trying to be smart about it and tried to detect when a base path was being used? (but checking all the routes and see if they have the same prefix?), did we scrap that? if not, could it be used here? 🤔

i'm not sure, i can't remember that... even if we did do that though, it might fail due to invalid functions being generated outside the base path

packages/next-on-pages/templates/_worker.js/utils/http.ts
james-elicx: yup...

dario-piotrowicz: ok then yeah I think it's fine to just leave it like this 🙂

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
dario-piotrowicz: although I am not 100% sure let's go with this as I feel like it's unlikely it's going to cause issues 👍

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
dario-piotrowicz: ok let's go with this, I feel like the risk of issues is hopefully pretty low :slightly_smiling_face:

dario-piotrowicz (approved):
Thanks @james-elicx :heart:  (sorry it took me a while to review the PR :sweat_smile:)

PR #519, opened 2023-10-28
feat: function and identifier sizes in the build log

james-elicx:
This PR does the following:
- Includes the raw byte length of code block identifiers in the build log.
- Includes the built function's size in the build log.
- Includes the list of consumer functions in the build log.


packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -201,6 +210,9 @@ async function buildFunctionFile(
 	const finalFileContents = `${functionImports}${fileContents}`;
 	const buildPromise = buildFile(finalFileContents, newFnPath, {
 		relativeTo: nopDistDir,
+	}).then(async () => {
+		const { size } = await stat(newFnPath);
+		fnInfo.outputSize = size;
```
dario-piotrowicz, line 215: I think output size is very generic, could you do `outputByteSize` or something like that?

(I know you're just following the fs naming, but this feels unclear to me and it's also inconsistent with `byteLength` where you clearly specify that the value is in bytes)

dario-piotrowicz (approved):
looks good to me, thanks James 😄

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
james-elicx: actually i did outputSize because we use outputPath, and im not using a buffer to get the byte length so i went with size instead of byteLength. It was just a coincidence that it's the same as what fs.stat gives. :P but yeah sure can change to outputByteSize

PR #534, opened 2023-11-10
temp-fix: rendering error in next.js 14.0.2 caused by next.js bug

james-elicx:
https://github.com/vercel/next.js/pull/57815 introduced a critical bug for us that breaks rendering due to the assumption that their internal request shape is always `NodeNextRequest` (and should have the `originalRequest` property on it), and is never `WebNextRequest` (which doesn't have the `originalRequest` property).

This fix is intended to be a temporary band-aid until the issue is fixed in Next.js. I raised an issue at https://github.com/vercel/next.js/issues/58265.

image

### DIFF
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -484,6 +484,14 @@ function fixFunctionContents(contents: string): string {
 		'$1',
 	);
 
+	// TODO: Remove once `vercel/next.js#58265` is fixed.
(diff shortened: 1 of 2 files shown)


packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -484,6 +484,14 @@ function fixFunctionContents(contents: string): string {
 		'$1',
 	);
 
+	// TODO: Remove once `vercel/next.js#58265` is fixed.
```
dario-piotrowicz, line 487:
```suggestion
	// TODO: Remove once https://github.com/vercel/next.js/issues/58265 is fixed.
```

dario-piotrowicz (approved):
Awesome! thanks a lot @james-elicx :heart: :pray: :heart:

Issue #535, opened 2023-11-10
[🐛 Bug]: NextJS fetch cache context lost when loading other edge routes 


tills13:
### next-on-pages environment related information

System:
        Platform: linux
        Arch: x64
        Version: #1 SMP Fri Jan 27 02:56:13 UTC 2023
        CPU: (12) x64 13th Gen Intel(R) Core(TM) i7-1365U
        Memory: 15 GB
        Shell: /bin/bash
Package Manager Used: npm (9.8.1)

Relevant Packages:
        @cloudflare/next-on-pages: 0.0.0-4572873 (beta/canary release)
        vercel: N/A
        next: 13.5.6

### Description

I'm not 100% sure of the conditions required, but it seems like if you have a sever action, and another edge worker script is loaded in a worker, [the AsyncLocalStorage context NextJS uses internally on fetch](https://github.com/vercel/next.js/blob/d1fe4d01c42c64154022a3156ea1b2ebe5fc4d54/packages/next/src/server/web/spec-extension/revalidate-tag.ts#L16) in the first edge worker is lost resulting in the following error when using `revalidatePath` or `revalidateTag`

```
Error: Invariant: static generation store missing in revalidateTag _N_T_/
```

This same behaviour happens in a production pages environment so it's not limited to Miniflare / this project.

### Reproduction

https://github.com/tills13/next-on-pages-bug

# Reproduction

1. npx --no-install @cloudflare/next-on-pages
1. npx wrangler d1 execute todos --local --file schema.sql
1. npx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat
1. go to http://localhost:8788
1. try to create a todo using the form

### Pages Deployment Method

Pages CI (GitHub/GitLab integration)

### Pages Deployment ID

I can DM you a larger app I have that exhibits this same behaviour lmk if you want that

### Would you like to help?

- [ ] Would you like to help fixing this bug?

james-elicx:
I was going to post here last night asking you to try the prerelease, but GitHub was having issues and took too long to start the workflow checks before I went to sleep. I just wanted to say thank you very much for providing a reproduction as it was very helpful when trying to find a fix :)

tills13:
I was desperate as this not getting fixed would mean having to use vercel instead of cloudflare. Really appreciate you taking the time. Glad I could help.

I looked around briefly but couldn't find one -- do you have a tip jar / buy-me-a-coffee link?

james-elicx:
[...]

PR #536, opened 2023-11-11
fix: patch fetch in the fetch handler instead of global scope

james-elicx:
fixes https://github.com/cloudflare/next-on-pages/issues/535

Honestly, this one kind of confuses me a bit, but from what I can tell, patching the fetch in the global scope results in it interacting with the static generation store from the previously run function.

You can see in the below screenshot that for some reason part way through running the function, it skips from the index.func.js to the other.func.js, even though there are no references in either file to use the webpack chunks inlined in the other (i am using disable-chunks-dedup here).

image

And my console.logs in each function file also indicate that it bizarrely jumps between executing code in one another.

image

<details>
<summary>index.func.js</summary>

image
image
image
</details>

<details>
<summary>other.func.js</summary>

image
image
image
</details>

So it really is quite strange that it is behaving like so when you patch the fetch in the global scope... Simply moving the fetch patch inside the fetch handler fixes the issue.

### DIFF
--- a/.changeset/twenty-eggs-lick.md
+++ b/.changeset/twenty-eggs-lick.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Patch the fetch function inside the fetch handler, instead of in the global scope.
--- a/packages/next-on-pages/templates/_worker.js/index.ts
+++ b/packages/next-on-pages/templates/_worker.js/index.ts
@@ -15,10 +15,10 @@ declare const __BUILD_OUTPUT__: VercelBuildOutput;
 
 declare const __ENV_ALS_PROMISE__: Promise<null | AsyncLocalStorage<unknown>>;
 
-patchFetch();
-
 export default {
 	async fetch(request, env, ctx) {
+		patchFetch();
+
 		const envAsyncLocalStorage = await __ENV_ALS_PROMISE__;
 		if (!envAsyncLocalStorage) {
 			const reqUrl = new URL(request.url);


tills13:
Amazing. 

Would a perhaps more permanent fix involve copying Next's internal patches over manually?

dario-piotrowicz (approved):
Looks good to me, thanks! :smiley:

dario-piotrowicz:
> Would a perhaps more permanent fix involve copying Next's internal patches over manually?

> You can find at least some of these in the link in the original Issue.

Hi @tills13 , could you clarify what you're suggesting? :slightly_smiling_face: 
And what's the original issue with the links? (#535 doesn't seem to have any such links... does it?) :slightly_smiling_face:

tills13:
I was thinking about this last night. 

Wouldn't the current solution result in a back-and-forth with Next, constantly patching and re-patching the global fetch since each side destroys the other's markers? Seems to me like it would effectively result in Next never hitting cache, as it would be recreated on every invocation. 

I cannot verify any of this, as I'm away from my computer this weekend, but this is based on my understanding of this project and Next. 

I'm happy with any solution that lets me use Next on Cloudflare, I'm just suggesting a more optimal outcome.

The link I'm referring to is https://github.com/vercel/next.js/blob/d1fe4d01c42c64154022a3156ea1b2ebe5fc4d54/packages/next/src/server/web/spec-extension/revalidate-tag.ts#L16

But you can see Next's own patches here: https://github.com/vercel/next.js/blob/bd2d8a3b23391b587efd3aeca81e79efa0455381/packages/next/src/server/lib/patch-fetch.ts#L644

Which would be steamrolled by next-of-page's patch.

james-elicx:
Hi @tills13, Next.js patches the fetch inside the edge function that it produces, and forwards the fetch calls to the original fetch, aka our patched fetch. It is worth noting that our patched fetch does a few important things; handles requests for bundled assets, ensures the user-agent header is present, and provides support for the suspense cache.

Your proposition to use Next.js patch function would be problematic in another way in that it would require us to import the function from Next.js, which would introduce other concerns and issues around implementation and our skip-build mode. In my view, we should not make the build process depend on being able to import something from Next.js. It just starts to get pretty messy, especially when you consider that it needs to work properly across various different versions of Next.js, which may differ. I would also consider copying the patch code instead of trying to import it to be unacceptable too.

If I understand what you're asking correctly and my thinking of the problem here is correct, that is actually necessary it seems. What appears to be happening is that we patch the fetch, then nextjs patches the fetch, and the next time it runs a function in the same worker instance, the function sees that their patch exists, they leave it and do not patch it again. this seems to cause problems with how it determines the static generation store instance to use, and that doesn't appear to work properly and it seems unable to retrieve the store. So from what I'm observing, to me it seems like repatching from both sides is crucial to enable the static generation store to be retrieved and consumed properly inside each route's edge function.

Anyway, I'm not sure I understand your point about "would effectively result in Next never hitting cache". Can you elaborate on that? In edge runtime functions, Next.js uses the Suspense Cache which involves making fetch calls that interact with remote suspense cache api endpoints. We handle these api calls in our patched fetch and direct them towards our own cache interface instead that interacts with the cache api. re-patching won't stop that from working because it's not an in-memory cache.

PR #540, opened 2023-11-12
Add support for extension .mjs in no-unsupported-configs rule

### LINKED ISSUE #532: [🐛 Bug]: ESLint rule no-unsupported-configs does not support config files with extension .mjs

System:
	Platform: darwin
	Arch: arm64
	Version: Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:43 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T6000
	CPU: (10) arm64 Apple M1 Pro
	Memory: 16 GB
	Shell: /bin/zsh
Package Manager Used: npm (10.1.0)

Relevant Packages:
	@cloudflare/next-on-pages: 1.7.2
	vercel: N/A
	next: N/A

### Description

ESLint rule `no-unsupported-configs` does not support config files with extension `.mjs`. Looking at the implementation it's going to have the same issue with the `.cjs` extension.

### Reproduction

1. Use ES module syntax in your Next.js config file and name it `next.config.mjs`.
2. Use a config value that is not allowed, e.g. `compress: true`.
3. Run linting.

It should report the error but it doesn't.

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

mattiasw:
Fixes #532

### DIFF
--- a/packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
+++ b/packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
@@ -93,7 +93,7 @@ const ruleSchema = {
 const rule: Rule.RuleModule = {
 	create: context => {
 		const code = context.sourceCode;
-		const exportedConfigName = context.filename.match(/next\.config\.js$/)
+		const exportedConfigName = context.filename.match(/next\.config\.m?js$/)
 			? getConfigVariableName(code)
 			: null;
 
@@ -217,6 +217,10 @@ function getConfigVariableName(code: SourceCode): string | null {
 		if (exportedValue?.type === 'Identifier') {
 			return exportedValue.name;
 		}
+		const esmExportedValue = extractESMExportValue(node);
+		if (esmExportedValue?.type === 'Identifier') {
+			return esmExportedValue.name;
+		}
 	}
 
 	const nodeAfterNextConfigComment = getNodeAfterNextConfigTypeComment(code);
@@ -269,6 +273,18 @@ function extractModuleExportValue(node: Node): Node | null {
 		node.expression.left.property.name === 'exports'
 	) {
 		return node.expression.right;
+	}
+
+	return null;
+}
+
+/**
+ * Gets the value of a node potentially representing: `export default ...`
+ * Returns the node of the value, or null if the input node doesn't represent the code
+ */
+function extractESMExportValue(node: Node): Node | null {
+	if (node.type === 'ExportDefaultDeclaration') {
+		return node.declaration;
 	}
 
 	return null;
--- a/packages/eslint-plugin-next-on-pages/tests/rules/no-unsupported-configs.test.ts
+++ b/packages/eslint-plugin-next-on-pages/tests/rules/no-unsupported-configs.test.ts
@@ -136,12 +136,12 @@ describe('no-unsupported-configs', () => {
 
 						/**
 						 *
-						 * @type {import(    "next"   )   .     
+						 * @type {import(    "next"   )   .
 						 *  NextConfig}
-						 * 
+						 *
 						 * The following is my config object
 						 * */
-						
+
 						/* this is another comment, surprisingly it doesn't break the type import! */
 						const nextConfig = {
 							experimental: {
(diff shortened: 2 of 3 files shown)


packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -93,7 +93,7 @@ const ruleSchema = {
 const rule: Rule.RuleModule = {
 	create: context => {
 		const code = context.sourceCode;
-		const exportedConfigName = context.filename.match(/next\.config\.js$/)
+		const exportedConfigName = context.filename.match(/next\.config\.[mc]?js$/)
```
dario-piotrowicz, line 96: I actually don't think that Next picks up `next.config.cjs` files, I can't see that in their documentation and when I tried to use it locally it wasn't indeed getting picked up, I think they only support `next.config.js` and `next.config.mjs`

so could remove the `c` here and remove the `cjs` extension mentioned in the rest of this PR? :pray: 

Unless I'm mistaken, please let me know if I am :slightly_smiling_face:

packages/eslint-plugin-next-on-pages/tests/rules/no-unsupported-configs.test.ts
```diff
@@ -463,4 +463,56 @@ describe('no-unsupported-configs', () => {
...
+
+			module.exports = nextConfig
+		`;
+		const invalidMessage =
+			'The "compress" configuration is not supported by next-on-pages (and is unlikely to be supported in the future).';
+		tester.run('', rule, {
+			valid: [
+				{
+					filename: 'next.config.mjs',
+					code: validCode,
```
dario-piotrowicz, line 489: here you are running tests for the `next.config.mjs` file which contains the standard cjs code, could you please change the code to use esm? :pray:

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -93,7 +93,7 @@ const ruleSchema = {
...
 			? getConfigVariableName(code)
```
dario-piotrowicz, line 97: `getConfigVariableName` only works with `module.exports = ...` could you also expand it to take into consideration `export default ...`? (for esm) :pray: 

(PS: we also have a fallback based on the next config type comment but I would prefer it if we could make this more robust and handle the export default as well)

dario-piotrowicz (commented):
Looks great! thanks a lot @mattiasw :smiley: 

I left a couple of comments thought, could you take a look? :slightly_smiling_face: :pray:

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
mattiasw: You are right, I only assumed, never looked it up. But it doesn't really make sense to have a .cjs config file when ESM is supported.

packages/eslint-plugin-next-on-pages/tests/rules/no-unsupported-configs.test.ts
mattiasw: Done!

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
mattiasw: Aha, so here is the tricky part. 😅 I will give it a try.

mattiasw: I have made a suggestion that seems to work.

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
```diff
@@ -273,3 +277,11 @@ function extractModuleExportValue(node: Node): Node | null {
 
 	return null;
 }
+
+function extractESMExportValue(node: Node): Node | null {
```
dario-piotrowicz, line 281: just for consistency with `extractModuleExportValue` could you adda tsdoc comment on top of the function? 🙂🙏

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
dario-piotrowicz: yeah looks good to me! thanks! 😄

dario-piotrowicz (approved):
Thanks so much @mattiasw 😃 

Just one minor comment, it's all good to go otherwise 😄

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
mattiasw: Yes, of course, it's done.

dario-piotrowicz:
Awesome! thanks a lot again @mattiasw 😄

mattiasw:
@dario-piotrowicz Thanks to you too for having such a welcoming attitude towards a first-time contributor.

packages/eslint-plugin-next-on-pages/src/rules/no-unsupported-configs.ts
mattiasw: Probably.

Issue #529, opened 2023-11-08
[⚡ Feature]: Add KV (or custom) support for revalidateTag (next/cache)


juanferreras:
Hi all!

I've seen several comments from maintainers ([[1]](https://github.com/cloudflare/next-on-pages/pull/419#issuecomment-1686123174), [[2]](https://github.com/cloudflare/next-on-pages/issues/444#issuecomment-1694771174)) following up the great work of [feat: fetch (suspense) cache handling, and next/cache support #419](https://github.com/cloudflare/next-on-pages/pull/419) by @james-elicx that given limitations on Cache API (e.g. per-colo cache and requires a custom domain to be configured), we'd mostly want to use alternative storage mechanisms in production (most likely KV, but perhaps other users might prefer D1/R2/DO based on specific use cases or even something else).

I understand that the ultimate intention is to let users define in their own `next.config.js` their own adaptor for caching – but there might be a pre-made snippet for KV, or an exported util they can reference.

I've noted that there isn't any issue tracking that – so wanted to drop this in for that purpose, and in case maintainers want to explain more what they have in mind (or if there are any blockers they're tracking). I'm happy to take a look at contributing the changes too.

Thanks!

james-elicx:
I actually originally wanted to let people opt in to other storage mechanisms by the presence of the binding being detected, e.g. `process.env.KV_SUSPENCE_CACHE` would use the KV cache, but I think others weren't a fan of doing that which is why it was scrapped from the initial PR.

The general idea that was settled upon would be to require people use a separate cache interface using the `incrementalCacheHandlerPath` option in `next.config.js`. The Next.js docs recommend the following usage: `incrementalCacheHandlerPath: require.resolve('./cache-handler.js')`. The problem with this is that you then get the import for the CJS version when trying to read the path from next.config.js, instead of the ESM one, because Next.js want you to use require for it, and that becomes a real pain since Workers want ESM.

Personally, I would actually quite like to add the binding-presence based approach for things like KV/R2/D1, and can quite easily revive that, but I'm not sure whether @dario-piotrowicz would be okay with that approach.

I would also note that Vercel's data cache is regional, so using the cache api is a rather similar experience.

ibobo:
Hi everyone!

[...]

PR #548, opened 2023-11-17
feat: kv binding support for suspense cache

james-elicx:
This PR does the following:
- Builds the cache templates to the nop dist dir (and fixes the import extension in them since we're not bundling the adaptor code).
- Changes to lazy import internal cache adaptors.
- Adds KV cache adaptor.
- Automatically uses KV cache adaptor when the `KV_SUSPENSE_CACHE` binding is present.
- Updates docs.

Cache API demo: https://next-cache-demo.pages.dev/
KV binding demo: https://kv.next-cache-demo.pages.dev/

partially solves https://github.com/cloudflare/next-on-pages/issues/529

### DIFF
--- a/.changeset/twelve-cobras-double.md
+++ b/.changeset/twelve-cobras-double.md
@@ -2,4 +2,9 @@
 '@cloudflare/next-on-pages': minor
 ---
 
-Support for using KV bindings with the Suspense Cache via `process.env.__NEXT_ON_PAGES__KV_SUSPENSE_CACHE`.
+Add support for using a KV to implement the Suspense Cache via naming convention
+
+With this change users can have their suspense cache implemented via a KV binding, in order to
+opt-in to such implementation they simply need to make sure that their application has a KV binding
+named `__NEXT_ON_PAGES__KV_SUSPENSE_CACHE` (the next-on-pages worker will pick up such
+binding and use it to implement the suspense cache instead of using the default workers cache API).
--- a/.changeset/twelve-cobras-double.md
+++ b/.changeset/twelve-cobras-double.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Support for using KV bindings with the Suspense Cache via `process.env.__NEXT_ON_PAGES__KV_SUSPENSE_CACHE`.
--- a/packages/next-on-pages/env.d.ts
+++ b/packages/next-on-pages/env.d.ts
@@ -5,6 +5,7 @@ declare global {
 			npm_config_user_agent?: string;
 			CF_PAGES?: string;
 			SHELL?: string;
+			KV_SUSPENSE_CACHE?: KVNamespace;
 			[key: string]: string | Fetcher;
 		}
 	}
--- a/packages/next-on-pages/env.d.ts
+++ b/packages/next-on-pages/env.d.ts
@@ -5,7 +5,7 @@ declare global {
 			npm_config_user_agent?: string;
 			CF_PAGES?: string;
 			SHELL?: string;
-			KV_SUSPENSE_CACHE?: KVNamespace;
+			__NEXT_ON_PAGES__KV_SUSPENSE_CACHE?: KVNamespace;
 			[key: string]: string | Fetcher;
 		}
 	}
--- a/packages/next-on-pages/src/buildApplication/buildWorkerFile.ts
+++ b/packages/next-on-pages/src/buildApplication/buildWorkerFile.ts
@@ -1,4 +1,4 @@
-import { readFile, writeFile } from 'fs/promises';
+import { writeFile } from 'fs/promises';
 import { join, relative } from 'path';
 import { build } from 'esbuild';
 import { tmpdir } from 'os';
@@ -88,21 +88,6 @@ export async function buildWorkerFile(
 		platform: 'neutral',
 		outdir: join(nopDistDir, 'cache'),
 		minify,
-		plugins: [
-			{
-				name: 'fix-import-extension-for-unbundled-adaptors',
-				setup(build) {
-					build.onLoad({ filter: /.*/ }, async args => {
-						const contents = await readFile(args.path, 'utf8');
-						const newContents = contents.replace(
-							"from './adaptor';",
-							"from './adaptor.js';",
-						);
-						return { contents: newContents, loader: 'default' };
-					});
-				},
-			},
-		],
 	});
 
 	return relative('.', outputFile);
--- a/packages/next-on-pages/src/buildApplication/buildWorkerFile.ts
+++ b/packages/next-on-pages/src/buildApplication/buildWorkerFile.ts
@@ -1,4 +1,4 @@
-import { writeFile } from 'fs/promises';
+import { readFile, writeFile } from 'fs/promises';
 import { join, relative } from 'path';
 import { build } from 'esbuild';
 import { tmpdir } from 'os';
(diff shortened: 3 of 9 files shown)


packages/next-on-pages/env.d.ts
```diff
@@ -5,6 +5,7 @@ declare global {
 			npm_config_user_agent?: string;
 			CF_PAGES?: string;
 			SHELL?: string;
+			KV_SUSPENSE_CACHE?: KVNamespace;
```
dario-piotrowicz, line 8:
```suggestion
			__NEXT_ON_PAGES__KV_SUSPENSE_CACHE?: KVNamespace;
```
(
or
```suggestion
			NEXT_ON_PAGES_KV_SUSPENSE_CACHE?: KVNamespace;
```
if the suggestion above is too much
)

I think it would be nice/beneficial to make it very obvious that this is a next-on-pages thing, so that people don't end up risking to confuse these with bindings they actually directly use, thinking that they can be Cloudflare internal ones (like the sites ones) etc...

dario-piotrowicz (commented):
This is missing a changeset :slightly_smiling_face:

packages/next-on-pages/env.d.ts
james-elicx: what about `NOP_KV_SUSPENSE_CACHE`?

dario-piotrowicz: 😢 

since it's just a binding that you can set and forget I don't see an issue in a longer (clearer/more recognizable) name

I guess you find `__NEXT_ON_PAGES__KV_SUSPENSE_CACHE` too too long? 😢 

(PS: if we do add a next-on-pages prefix, then we could actually simplify the latter part of the name, for example by dropping `SUSPENSE`? 🤔)

james-elicx: i guess `__NEXT_ON_PAGES__KV_CACHE` works, just kinda long 😓

james-elicx: or maybe `NEXT_ON_PAGES_KV_CACHE` without the extra underscores, would that be okay?

dario-piotrowicz: I was thinking of the extra underscores as a separation between `NEXT_ON_PAGES` and the rest so that it could be a nice prefix indicator for any potential other bindings we might want to add
(e.g. `__NEXT_ON_PAGES__KV_CACHE`, `__NEXT_ON_PAGES__xxx`, ...)

I think it'd be nice, but if you really dislike it I can concede and not having them 🙂

packages/next-on-pages/src/buildApplication/buildWorkerFile.ts
```diff
@@ -79,5 +79,38 @@ export async function buildWorkerFile(
...
+		outdir: join(nopDistDir, 'cache'),
+		minify,
+		plugins: [
+			{
+				name: 'fix-import-extension-for-unbundled-adaptors',
+				setup(build) {
+					build.onLoad({ filter: /.*/ }, async args => {
+						const contents = await readFile(args.path, 'utf8');
+						const newContents = contents.replace(
+							"from './adaptor';",
```
dario-piotrowicz, line 98: could you add `.js` in the ts template code instead?

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -71,6 +72,29 @@ export async function handleSuspenseCacheRequest(request: Request) {
...
-	return new CacheApiAdaptor();
+	if (process.env.KV_SUSPENSE_CACHE) {
+		return getInternalCacheAdaptor('kv');
+	}
+
+	return getInternalCacheAdaptor('cache-api');
+}
+
+type InternalAdaptors = 'kv' | 'cache-api';
+type DeriveAdaptorType<T extends InternalAdaptors> = T extends 'kv'
```
dario-piotrowicz, line 83: Is there a benefit in having different types of adaptors? couldn't/shouldn't all the adaptors just follow a single Adaptor type contract/interface?

packages/next-on-pages/env.d.ts
james-elicx: but people might make a typo is what i was thinking

dario-piotrowicz: I would assume most people not typing it but copy-pasting it from the docs anyways 🤔 🤷 

But yeah up to you, I'm fine with not having the double slashes if you're really concerned about that 🙂

smebberson: I like `__NEXT_ON_PAGES__KV_SUSPENSE_CACHE` FWIW.

packages/next-on-pages/templates/_worker.js/utils/cache.ts
james-elicx: my bad, these both extend CacheAdaptor so i can just use that 🤦🏻‍♂️

packages/next-on-pages/src/buildApplication/buildWorkerFile.ts
james-elicx: looks like it 👍🏻

.changeset/twelve-cobras-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Support for using KV bindings with the Suspense Cache via `process.env.__NEXT_ON_PAGES__KV_SUSPENSE_CACHE`.
```
dario-piotrowicz, line 5:
```suggestion
Add support for using a KV to implement the Suspense Cache via naming convention

With this change users can have their suspense cache implemented via a KV binding, in order to
opt-in to such implementation they simply need to make sure that their application has a KV binding
named `__NEXT_ON_PAGES__KV_SUSPENSE_CACHE` (the next-on-pages worker will pick up such
binding and use it to implement the suspense cache instead of using the default workers cache API).
```

dario-piotrowicz: (or something along those lines)

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -71,6 +70,22 @@ export async function handleSuspenseCacheRequest(request: Request) {
  * @returns Adaptor for the suspense cache.
  */
 export async function getSuspenseCacheAdaptor(): Promise<CacheAdaptor> {
-	// TODO: Try to lazy import the custom cache adaptor.
-	return new CacheApiAdaptor();
+	if (process.env.__NEXT_ON_PAGES__KV_SUSPENSE_CACHE) {
```
dario-piotrowicz, line 73: I think we haven't considered the case in which the user provides a `__NEXT_ON_PAGES__KV_SUSPENSE_CACHE` binding but that is not a KV

(although it seems quite unlikely to me 🤷)

Should we show a warning if this binding is not a KV and still use the default cache API adaptor? (if it doesn't contain `get`, `set`)?
(That developers could see locally or if they tail their worker)

dario-piotrowicz (approved):

packages/next-on-pages/templates/_worker.js/utils/cache.ts
james-elicx: if someone wants to set the wrong binding for that key, that's a user error and will probably result in an internal server error. i don't see the problem here

dario-piotrowicz: > if someone wants to set the wrong binding for that key, that's a user error and will probably result in an internal server error. i don't see the problem here

I don't think people "want" to make mistakes 😅

I'm just saying that if the user makes a mistake we can either make the whole thing blow up (like it currently would) or try to gracefully handle the situation.

I would prefer to gracefully handle it, but you could argue that that would make the error less evident/discoverable, so I am fine with either.

PR #557, opened 2023-11-22
fix: Falling back to header if cache tags are not given from fetched object

### LINKED ISSUE #556: [🐛 Bug]: `next/cache`: Handle changed type signature of Next.js 13.5+
> Hello! I have been debugging `revalidateTag` inconsistency and found out some breaking changes of upstream.
> 
> Internal cache handling behavior of Next.js seems to be changed starting from version 13.5 and above. `IncrementalCacheValue` no longer provides `tags` attribute, results in inability of flushing cache with tags. (despite [revalidate code](https://github.com/cloudflare/next-on-pages/blob/2d5dde95176bca46cdd63c8b204699127017c219/packages/next-on-pages/templates/_worker.js/utils/cache.ts#L19C44-L19C54) works fine)

looks like the type signature changed in next.js :(

https://github.com/vercel/next.js/blob/canary/packages/next/src/server/response-cache/types.ts

would you be able to open a new issue for this @susemeee? i recall on your other issue you ticked the box that you'd be interested in helping fix the bug. would you like to give this issue a go? if not, just lmk and i'll be happy to look into a fix

_Originally posted by @james-elicx in https://github.com/cloudflare/next-on-pages/issues/451#issuecomment-1822392271_

susemeee:
- Updated `CachedFetchValue` type introduced in Next.js 13.5.
- When retrieving cache tags, falling back to `x-vercel-cache-tags` header if cache tags are not given from fetched `CachedFetchValue` object.

fixes #556

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/utils/cache.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache.ts
@@ -2,6 +2,9 @@ import type { CacheAdaptor, IncrementalCacheValue } from '../../cache';
 import { SUSPENSE_CACHE_URL } from '../../cache';
 import { CacheApiAdaptor } from '../../cache/cache-api';
 
+// https://github.com/vercel/next.js/blob/48a566bc4fcfc48d8489f096a87e792912c4989f/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L19
+const CACHE_TAGS_HEADER = 'x-vercel-cache-tags';
+
 /**
  * Handles an internal request to the suspense cache.
  *
@@ -51,6 +54,9 @@ export async function handleSuspenseCacheRequest(request: Request) {
 			case 'POST': {
 				// Update the value in the cache.
 				const body = await request.json<IncrementalCacheValue>();
+				// Falling back to the cache tags header for Next.js 13.5+
+				body.tags = body.tags ?? body.data.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',') ?? [];
+
 				await cache.set(cacheKey, body);
 
 				return new Response(null, { status: 200 });
--- a/packages/next-on-pages/templates/_worker.js/utils/cache.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache.ts
@@ -56,7 +56,7 @@ export async function handleSuspenseCacheRequest(request: Request) {
 				const body = await request.json<IncrementalCacheValue>();
 				// Falling back to the cache tags header for Next.js 13.5+
 				if (body.data.tags === undefined) {
-					body.tags = body.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
+					body.tags ??= request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
 				} else {
 					body.tags = body.data.tags;
 					delete body.data.tags;
--- a/packages/next-on-pages/templates/_worker.js/utils/cache.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache.ts
@@ -2,7 +2,7 @@ import type { CacheAdaptor, IncrementalCacheValue } from '../../cache';
 import { SUSPENSE_CACHE_URL } from '../../cache';
 import { CacheApiAdaptor } from '../../cache/cache-api';
 
-// https://github.com/vercel/next.js/blob/48a566bc4fcfc48d8489f096a87e792912c4989f/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L19
+// https://github.com/vercel/next.js/blob/48a566bc/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L19
 const CACHE_TAGS_HEADER = 'x-vercel-cache-tags';
 
 /**
--- a/packages/next-on-pages/templates/_worker.js/utils/cache.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache.ts
@@ -55,7 +55,12 @@ export async function handleSuspenseCacheRequest(request: Request) {
 				// Update the value in the cache.
 				const body = await request.json<IncrementalCacheValue>();
 				// Falling back to the cache tags header for Next.js 13.5+
-				body.tags = body.tags ?? body.data.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',') ?? [];
+				if (body.data.tags === undefined) {
+					body.tags = body.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
+				} else {
+					body.tags = body.data.tags;
+					delete body.data.tags;
+				}
 
 				await cache.set(cacheKey, body);
 
--- a/packages/next-on-pages/templates/_worker.js/utils/cache.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache.ts
@@ -56,7 +56,11 @@ export async function handleSuspenseCacheRequest(request: Request) {
 				const body = await request.json<IncrementalCacheValue>();
 				// Falling back to the cache tags header for Next.js 13.5+
 				if (body.data.tags === undefined) {
-					body.tags ??= request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
+					body.tags ??=
+						request.headers
+							.get(CACHE_TAGS_HEADER)
+							?.split(',')
+							?.filter(Boolean) ?? [];
 				} else {
 					body.tags = body.data.tags;
 					delete body.data.tags;
--- a/packages/next-on-pages/templates/cache/adaptor.ts
+++ b/packages/next-on-pages/templates/cache/adaptor.ts
@@ -88,7 +88,7 @@ export class CacheAdaptor {
 
 				// Check if the cache entry is stale or fresh based on the tags.
 				const tags = getDerivedTags(
-					data.value.tags ?? data.value.data.tags ?? []
+					data.value.tags ?? data.value.data.tags ?? [],
 				);
 				const isStale = tags.some(tag => {
 					// If a revalidation has been triggered, the current entry is stale.
--- a/packages/next-on-pages/templates/cache/adaptor.ts
+++ b/packages/next-on-pages/templates/cache/adaptor.ts
@@ -196,7 +196,7 @@ export type CachedFetchValue = {
 		body: string;
 		url: string;
 		status?: number;
-		// old version; https://github.com/vercel/next.js/blob/fda1ecc/packages/next/src/server/response-cache/types.ts#L16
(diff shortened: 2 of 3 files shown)


packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -87,7 +87,7 @@ export class CacheAdaptor {
 				await this.loadTagsManifest();
 
 				// Check if the cache entry is stale or fresh based on the tags.
-				const tags = getDerivedTags(data.value.data.tags ?? []);
+				const tags = getDerivedTags(data.value.tags ?? []);
```
james-elicx, line 90:
```suggestion
				const tags = getDerivedTags(
					data.value.tags ?? data.value.data.tags ?? []
				);
```

This should also check for tags in `data.value.data.tags` otherwise it won't work with existing cache entries.

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -51,6 +54,9 @@ export async function handleSuspenseCacheRequest(request: Request) {
 			case 'POST': {
 				// Update the value in the cache.
 				const body = await request.json<IncrementalCacheValue>();
+				// Falling back to the cache tags header for Next.js 13.5+
+				body.tags = body.tags ?? body.data.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',') ?? [];
```
james-elicx, line 58: I think we should also delete any `body.data.tags` property that exists so that we aren't storing the tags twice (on both `body.tags` and `body.data.tags`) for older Next.js versions.

james-elicx, line 58:
```suggestion
				body.tags = body.tags ?? body.data.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
```
Since the header defaults to `''` (`[CACHE_TAGS_HEADER]: tags?.join(',') || ''`), I think that might mean we could end up with `['']` since we're splitting on `,`, so we should filter that out.

james-elicx (commented):
thank you for your work on this, I just have a few comments/suggestions :)

deployed demo for these changes with latest version of next.js: https://848091be.next-cache-demo.pages.dev/

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -55,7 +55,12 @@ export async function handleSuspenseCacheRequest(request: Request) {
 				// Update the value in the cache.
 				const body = await request.json<IncrementalCacheValue>();
 				// Falling back to the cache tags header for Next.js 13.5+
-				body.tags = body.tags ?? body.data.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',') ?? [];
+				if (body.data.tags === undefined) {
+					body.tags = body.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
+				} else {
+					body.tags = body.data.tags;
+					delete body.data.tags;
```
susemeee, line 62: Not sure whether to have this or not - because we're also checking for `body.data.tags` in retrieval phase

susemeee (commented):
Thank you for detailed comment! Understood and accepted most of suggestions.

Also I proposed a change for the fallback branch. May I ask for another review? @james-elicx

packages/next-on-pages/templates/_worker.js/utils/cache.ts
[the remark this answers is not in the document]
susemeee: I wonder where should I put it if I move this thing out and make it in a function. In following commit I've changed the code just to make it a little clearer, will that be enough or still think that separate function would be better?

packages/next-on-pages/templates/_worker.js/utils/cache.ts
james-elicx: Getting from the cache checks `body.data.tags` for backwards compatibility. Deleting here is not going to impact that since this it also checks `body.tags` which is what this will use instead.

packages/next-on-pages/templates/_worker.js/utils/cache.ts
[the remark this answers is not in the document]
james-elicx: it's fine no worries, was just being extra picky

james-elicx (approved):
Tested using Next.js 13.4.1 cache entries from current next-on-pages, then using the changes in this branch, then switching to 14.0.3. Everything seems good to me - thanks :)

packages/next-on-pages/templates/_worker.js/utils/cache.ts
dario-piotrowicz: Is deleting the `tags` field here necessary/beneficial in any way? if not I would suggest not to do it

packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -54,7 +54,7 @@ export class CacheAdaptor {
 		switch (newEntry.value?.kind) {
 			case 'FETCH': {
 				// Update the tags with the cache key.
-				const tags = newEntry.value.data.tags ?? [];
+				const tags = newEntry.value.tags ?? [];
```
dario-piotrowicz, line 57: is this correct? should it not be `newEntry.value.tags ?? newEntry.value.data.tags ?? []`?

packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -186,16 +188,20 @@ export type TagsManifest = {
...
-// https://github.com/vercel/next.js/blob/fda1ecc/packages/next/src/server/response-cache/types.ts#L16
+// https://github.com/vercel/next.js/blob/df4c2aa8/packages/next/src/server/response-cache/types.ts#L24
 export type CachedFetchValue = {
 	kind: 'FETCH';
 	data: {
 		headers: { [k: string]: string };
 		body: string;
 		url: string;
 		status?: number;
+		// old version; https://github.com/vercel/next.js/blob/fda1ecc/packages/next/src/server/response-cache/types.ts#L16
```
dario-piotrowicz, line 199:
```suggestion
		// field used by older versions of Next.js (see: https://github.com/vercel/next.js/blob/fda1ecc/packages/next/src/server/response-cache/types.ts#L23)
```

packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -51,6 +54,14 @@ export async function handleSuspenseCacheRequest(request: Request) {
 			case 'POST': {
 				// Update the value in the cache.
 				const body = await request.json<IncrementalCacheValue>();
+				// Falling back to the cache tags header for Next.js 13.5+
+				if (body.data.tags === undefined) {
+					body.tags = body.tags ?? request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
```
dario-piotrowicz, line 59:
```suggestion
					body.tags ??= request.headers.get(CACHE_TAGS_HEADER)?.split(',')?.filter(Boolean) ?? [];
```

packages/next-on-pages/templates/cache/adaptor.ts
james-elicx: the tags are changed to be on value.tags instead of value.data.tags in the other file before calling .set.

packages/next-on-pages/templates/_worker.js/utils/cache.ts
james-elicx: > Is deleting the `tags` field here necessary/beneficial in any way? if not I would suggest not to do it

storing less data in the cache if the user is on an older version of next.js. no need to store the tags twice. while it would be a small amount of extra data, things like kv still bill for data stored.

dario-piotrowicz: yeah, sorry, my bad, you're right :+1:

packages/next-on-pages/templates/cache/adaptor.ts
dario-piotrowicz: yes, you're totally right! sorry my bad again :sweat: 

Although... I wonder if we could improve the types to make clear which contains which tags :thinking: 
(PS: even if we want to do that it does not need to be done in this PR, it's just a thought)

dario-piotrowicz (approved):

dario-piotrowicz:
@susemeee could you format the code? 🙂 (by just running `npm run prettier:fix` at the repo's root)

Besides that I'm happy to merge the PR 😄 (thanks a lot for the PR btw! ❤️ ), unless you'd like to improve the typing before the merge (https://github.com/cloudflare/next-on-pages/pull/557#discussion_r1405234521), please let me know

dario-piotrowicz:
@susemeee I formatted the code for you so that we could merge the PR and include it in the release we're publishing, I hope you don't mind 🙏 

Regarding the types improvement please feel free to open a new PR for it if you want 🙂 (if you don't want to make that improvement please let me know and I will open a PR for it)

susemeee:
> @susemeee could you format the code? 🙂 (by just running `npm run prettier:fix` at the repo's root)
> 
> Besides that I'm happy to merge the PR 😄 (thanks a lot for the PR btw! ❤️ ), unless you'd like to improve the typing before the merge ([#557 (comment)](https://github.com/cloudflare/next-on-pages/pull/557#discussion_r1405234521)), please let me know

Thank you! should have been known about that.

> Regarding the types improvement please feel free to open a new PR for it if you want 🙂 (if you don't want to make that improvement please let me know and I will open a PR for it)

What will be the scope if the types improvement will take place?

dario-piotrowicz:
> Thank you! should have been known about that.

No worries 😄

> What will be the scope if the types improvement will take place?

The scope would be pretty limited, roughly I was just thinking of this type: https://github.com/cloudflare/next-on-pages/blob/733fed9a9dbb0af27d6e1c2fb8d237a8669beb3c/packages/next-on-pages/templates/cache/adaptor.ts#L202-L216 being split into two, something like:
```ts
// https://github.com/vercel/next.js/blob/df4c2aa8/packages/next/src/server/response-cache/types.ts#L24
type CachedFetchValue = {
	kind: 'FETCH';
	data: {
		headers: { [k: string]: string };
		body: string;
		url: string;
		status?: number;
	};
	tags?: string[];
	revalidate: number;
};

// comment here indicating that this matches the type we get from Next.js
type CachedFetchValueRawFromNextjs = CachedFetchValue & {
	data: CachedFetchValue['data'] & {
		// field used by older versions of Next.js (see: https://github.com/vercel/next.js/blob/fda1ecc/packages/next/src/server/response-cache/types.ts#L23)
		tags?: string[];
	};
}
```
and then in the code appropriately use `CacheFetchValueRawFromNextjs` where the value comes from Next.js (and can contain the data.tags array) and `CacheFetchValue` otherwise (so that it is clear when the data.tags field can be present or not)

Or something like this 🙂 (I'd need to look at the code better but this is roughly what I had in mind)

If you'd like to do it please go ahead otherwise I can totally do it 🙂👍

dario-piotrowicz:
@susemeee sure we can wait and see, there's no rush 🙂, it was also just a minor idea just to clarify where the tags can be present in the program flow more accurately

PR #559, opened 2023-11-25
fix: set `x-vercel-sc-host` for older versions of Next.js

james-elicx:
Older versions of Next.js only use the `x-vercel-sc-host` request header and not the `SUSPENSE_CACHE_URL` env variable, therefore set the request header.


dario-piotrowicz (approved):
LGTM :slightly_smiling_face::+1:

Issue #558, opened 2023-11-22
[🐛 Bug]: Defects when combining dynamic, parallel, and intercept routes


Karibash:
### next-on-pages environment related information

```
System:
	Platform: darwin
	Arch: x64
	Version: Darwin Kernel Version 22.6.0: Fri Sep 15 13:39:52 PDT 2023; root:xnu-8796.141.3.700.8~1/RELEASE_X86_64
	CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
	Memory: 32 GB
	Shell: /usr/local/bin/fish
Package Manager Used: npm (10.1.0)

Relevant Packages:
	@cloudflare/next-on-pages: 1.7.4
	vercel: N/A
	next: 14.0.3
```

### Description

When combining dynamic routes, parallel routes, and interceptor routes, the following issues occur:
- When accessing a page through an intra-page transition to a parallel route, the page itself can be displayed, but the path parameters are not being successfully retrieved.
- Reloading the page in a parallel route results in a 404 error, rendering the page inaccessible.

Perhaps there is a glitch in the [routes matcher for next-on-pages](https://github.com/cloudflare/next-on-pages/blob/main/packages/next-on-pages/templates/_worker.js/routes-matcher.ts), but it is so complex that I could not pinpoint what the problem is.

### Reproduction

https://github.com/Karibash/nextjs-dynamic-interception-parallel-routes

### Pages Deployment Method

Direct Upload (`wrangler pages publish` or the @cloudflare/pages-action GitHub Action)

### Would you like to help?

- [ ] Would you like to help fixing this bug?

james-elicx:
> I have updated the version of @cloudflare/next-on-pages to 1.8.0, but the problem is still not resolved.

will look into it tonight

james-elicx:
Hey there @Karibash, so sorry for the delay in getting to this and thank you very much for the reproduction :)

I have opened a PR which should fix these issues, and I would welcome you to try out the prerelease in https://github.com/cloudflare/next-on-pages/pull/564#issuecomment-1828683330

Karibash:
Hi, @james-elicx 
Thanks for the fix!
I have tried it in the pre-release and can confirm that it works correctly.
I look forward to the release of this fix🥳

PR #564, opened 2023-11-27
fix: route intercepts causing 404s + missing params

james-elicx:
This PR does the following:
- Fixes an issue where route intercepts would cause 404s for the non-intercepted route.
- Fixes an issue where route intercepts would not have all the correct parameters for the route.

See the comments in the code for more context.

Reproduction that this fixes is available at https://github.com/Karibash/nextjs-dynamic-interception-parallel-routes

fixes https://github.com/cloudflare/next-on-pages/issues/558

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
@@ -90,7 +90,10 @@ export class RoutesMatcher {
 	 */
 	private checkRouteMatch(
 		route: VercelSource,
-		checkStatus?: boolean,
+		{
+			checkStatus,
+			checkIntercept,
+		}: { checkStatus: boolean; checkIntercept: boolean },
 	): { routeMatch: MatchPCREResult; routeDest?: string } | undefined {
 		const srcMatch = matchPCRE(route.src, this.path, route.caseSensitive);
 		if (!srcMatch.match) return;
@@ -133,6 +136,19 @@ export class RoutesMatcher {
 
 		// Required status code must match (i.e. for error routes) - skip if not met.
 		if (checkStatus && route.status !== this.status) {
+			return;
+		}
+
+		// If the new destination is an intercept route, only allow it if the current path is also an
+		// intercept route. The build output config correctly maps relevant request paths to be
+		// intercepts in the `none` phase, while the `rewrite` phase can contain entries that rewrite
+		// to an intercept that matches requests that are not actually intercepts, causing a 404.
+		if (
+			checkIntercept &&
+			route.dest &&
+			/\/(\(\.+\))+/.test(route.dest) &&
+			!/\/(\(\.+\))+/.test(this.path)
+		) {
 			return;
 		}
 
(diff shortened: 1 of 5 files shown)


packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -136,6 +139,19 @@ export class RoutesMatcher {
 			return;
 		}
 
+		// If the new destination is an intercept route, only allow it if the current path is also an
+		// intercept route. The build output config correctly maps relevant request paths to be
+		// intercepts in the `none` phase, while the `rewrite` phase can contain entries that rewrite
+		// to an intercept that matches requests that are not actually intercepts, causing a 404.
```
dario-piotrowicz, line 145: Here to me mentioning the phases is a bit confusing as they don't seem to be relevant, I would suggest to reduce this comment
```suggestion
		// If the new destination is an intercept route, only allow it if the current path is also an
		// intercept route.
```
and add the rest below above where you set the `checkIntercept`:
```diff
+       // The build output config correctly maps relevant request paths to be
+		// intercepts in the `none` phase, while the `rewrite` phase can contain entries that rewrite
+		// to an intercept that matches requests that are not actually intercepts, causing a 404.
        checkIntercept: phase === 'rewrite',
```

What do you think? 🙂

packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -136,6 +139,19 @@ export class RoutesMatcher {
...
+		if (
+			checkIntercept &&
+			route.dest &&
+			/\/(\(\.+\))+/.test(route.dest) &&
+			!/\/(\(\.+\))+/.test(this.path)
```
dario-piotrowicz, line 150: I don't thing I understand the comment above in regards of these, the comment says
"if the new destination is an intercept route, only allow it if the current path is also an intercept route"
but here is looks to me like the dest is an intercept route and the current path is not, am I missing something?

dario-piotrowicz (approved):
Thanks a lot @james-elicx 😄 (sorry for the late review 🙇)

Look great to me, I just left a couple of minor comments

packages/next-on-pages/templates/_worker.js/routes-matcher.ts
james-elicx: yeah so what im doing here is not letting the path be rewritten to the intercept dest if the path isn't already an intercept

so i guess the comment is kind of the inverse?

james-elicx: happy to change it something clearer

dario-piotrowicz: ah ok I see! sorry my bad I got confused by it...

So I think that the comment is ok indeed, but what about maybe changing this to something like this:
Screenshot 2023-11-29 at 09 25 01

In this way the `if` check becomes very clear and the comment also follows the same structure all the other ones in this function do

what do you think?

james-elicx: looks like a good plan to me

PR #568, opened 2023-12-01
Properly handle external middleware rewrites

### LINKED ISSUE #551: [🐛 Bug]: Allow `NextResponse.rewrite()` to proxy a different host in middleware.
### Description

I am not sure this is even within the scope of next-on-pages or a fundamental Workers security choice. If it is in scope, I'm not sure if it is a feature request or a bug in how Pages interprets Next.js middleware.

For context, we have a marketing site, docs site, and web app. Each are Next.js projects, but it is required for multiple reasons that they be separate sites. 

On Vercel we hosted the app and docs on subdomains. The marketing site, on the root domain, used Next.js middleware to rewrite requests. For example, `/dashboard` to `app.site.com/app/dashboard` and `/docs/welcome` to `docs.site.com/docs/welcome`. By setting `app` and `docs` as the `basePath` for those sites, this worked well and resulted in no routing conflicts. 

On Cloudflare Pages, redirects and same-host rewrites seem to work the same, but rewrites that change the host (to another Pages project on a subdomain) fail silently at runtime. The middleware runs, and `console.log()` shows that the logic is functioning fine, but when the middleware returns the result of `NextResponse.rewrite()`, Pages returns a 404 rather than rewriting. 

Because Vercel uses Workers as their 'edge runtime', I had assumed the behavior would match between Vercel and CF Pages.

### Additional Information

I first submitted [this bug](https://github.com/cloudflare/next-on-pages/issues/546) but realized that the actual difference in behavior was more narrow and possibly not a bug at all. 

### Would you like to help?

dario-piotrowicz:
resolves #551

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
@@ -197,16 +197,7 @@ export class RoutesMatcher {
 			const rewriteHost = newUrl.hostname;
 			const rewriteIsExternal = thisHost !== rewriteHost;
 
-			if (rewriteIsExternal) {
-				const externalResp = await fetch(newUrl, this.reqCtx.request);
-				// The rewrite is external so we do not want to continue and we just
-				// set the final response body, status and headers here.
-				this.body = externalResp.body;
-				this.status = externalResp.status;
-				applyHeaders(this.headers.normal, externalResp.headers);
-			}
-
-			this.path = newUrl.pathname;
+			this.path = rewriteIsExternal ? newUrl.toString() : newUrl.pathname;
 			applySearchParams(this.searchParams, newUrl.searchParams);
 
 			resp.headers.delete(rewriteKey);
--- a/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
+++ b/packages/next-on-pages/templates/_worker.js/routes-matcher.ts
@@ -161,7 +161,7 @@ export class RoutesMatcher {
 	 *
 	 * @param resp Middleware response object.
 	 */
-	private async processMiddlewareResp(resp: Response): Promise<void> {
+	private processMiddlewareResp(resp: Response): void {
 		const overrideKey = 'x-middleware-override-headers';
 		const overrideHeader = resp.headers.get(overrideKey);
 		if (overrideHeader) {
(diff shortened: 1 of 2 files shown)


packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -189,8 +189,24 @@ export class RoutesMatcher {
...
+			const rewriteIsExternal = thisHost !== rewriteHost;
+
+			if (rewriteIsExternal) {
+				const externalResp = await fetch(newUrl, this.reqCtx.request);
+				// The rewrite is external so we do not want to continue and we just
+				// set the final response body, status and headers here.
+				this.body = externalResp.body;
+				this.status = externalResp.status;
+				applyHeaders(this.headers.normal, externalResp.headers);
+				return;
```
james-elicx, line 207: I don't think returning here is the right thing to do. In every other case, we are apply the middleware response's headers to headers.normal, but for this we are not. When it generates a response to return later on, it would apply the headers.normal to the final response, but since this is skipping updating headers.normal, it will never get other headers that were set in the middleware.

dario-piotrowicz: good point, return removed, thanks 🙂

packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -189,8 +189,23 @@ export class RoutesMatcher {
...
+			const rewriteHost = newUrl.hostname;
+			const rewriteIsExternal = thisHost !== rewriteHost;
+
+			if (rewriteIsExternal) {
+				const externalResp = await fetch(newUrl, this.reqCtx.request);
+				// The rewrite is external so we do not want to continue and we just
+				// set the final response body, status and headers here.
+				this.body = externalResp.body;
+				this.status = externalResp.status;
+				applyHeaders(this.headers.normal, externalResp.headers);
```
james-elicx, line 206: I've just been looking at the part where we generate a response and I wonder if we should just update the path to be the full URL here instead of doing the fetch etc., and let that be handled in the generate response function instead.

https://github.com/cloudflare/next-on-pages/blob/1e73555dcb2ead5a83436ed619d5eec660b6c2f3/packages/next-on-pages/templates/_worker.js/handleRequest.ts#L99-L103

What do you think?

dario-piotrowicz: Yes, of course! I totally missed that :facepalm:

jculvey (approved):

PR #569, opened 2023-12-02
some support for suspense cache soft/implicit tags

james-elicx:
This PR does the following:
- Fixes old Next.js versions not updating cache entry properly because they didn't receive the tags how they expected. Sorry, this is my fault since I suggested to delete that property.
- Adds some support for soft/implicit tags that were introduced by https://github.com/vercel/next.js/pull/53321.

Fwiw, I think there might still be some more work potentially to improve support for newer versions of the suspense cache, but this PR should bring us to a good spot.

I did actually notice that Next.js deciding when it needs to update a stale entry was more reliable on older Next.js versions before they made these changes over the past few months, but that would be something that might need looking into separately, but I don't think there's anything we can really do in that department.

🙃

### DIFF
--- a/.changeset/good-squids-join.md
+++ b/.changeset/good-squids-join.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Improved support for newer version of Next.js with the suspense cache through trying to handle soft/implicit tags properly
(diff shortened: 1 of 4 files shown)


dario-piotrowicz (approved):
LGTM, thanks :slightly_smiling_face:

PR #573, opened 2023-12-04
Improving e2es and removing e2e-like unit tests (second part)

dario-piotrowicz:
Continuation of #426

____

This PR:
 - adds e2es for next 14.0.0 (and adds 14 to our supported.md doc)
 - removes the `handleRequest` unit tests and replaces them with e2es


pages-e2e/features/_utils/getAssertVisible.ts
```diff
@@ -19,14 +19,21 @@ async function assertVisible(
...
+		try {
+			await locator.waitFor({
+				timeout: 200,
+			});
+		} catch {}
+		isVisible = await locator.isVisible();
+		if(isVisible) {
+			break;
+		} else {
+			await new Promise(resolve => setTimeout(resolve, 200));
```
petebacondarwin, line 34: Do you need this extra timeout of 200ms, given that waitFor will wait for 200ms itself?

petebacondarwin: Actually I am confused. The waitFor on the locator can return even if the selected thing is not visible?

packages/next-on-pages/tests/templates/handleRequest.test.ts
```diff
@@ -1,7 +1,6 @@
 import { afterAll, expect, suite, test, vi } from 'vitest';
 import {
 	i18nTestSet,
-	infiniteLoopTestSet,
```
petebacondarwin, line 6: Why is it not possible to create an e2e test for this?

pages-e2e/features/appConfigsTrailingSlashFalse/setup.ts
```diff
@@ -0,0 +1,6 @@
+import { copyWorkspaceAssets } from '../_utils/copyWorkspaceAssets';
+await copyWorkspaceAssets();
+
+// Note:
+//  There's no need to update the next.config.js file here as the
+//  trailingSlash property if false by default
```
petebacondarwin:
```suggestion
//  trailingSlash property is false by default
```

pages-e2e/features/appConfigsTrailingSlashFalse/trailingSlash.test.ts
```diff
@@ -10,6 +10,6 @@ describe('default trailing slashes redirection', () => {
 
 		expect(response.redirected).toBe(true);
 		expect(response.status).toBe(200);
-		expect(response.url, `${DEPLOYMENT_URL}/api/routing-trailing-slash-test`);
+		expect(response.url).toEqual(`${DEPLOYMENT_URL}/api/routing-trailing-slash-test`);
```
petebacondarwin: 👍

pages-e2e/features/appConfigsTrailingSlashTrue/trailingSlash.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, test } from 'vitest';
+
+describe('default trailing slashes redirection', () => {
+	test(`fetching without an added trailing slashed results in a redirected response`, async ({
```
petebacondarwin, line 4:
```suggestion
	test(`fetching without an added trailing slash results in a redirected response`, async ({
```

pages-e2e/features/appConfigsTrailingSlashFalse/setup.ts
```diff
@@ -3,4 +3,4 @@ await copyWorkspaceAssets();
 
 // Note:
 //  There's no need to update the next.config.js file here as the
-//  trailingSlash property if false by default
\ No newline at end of file
+//  trailingSlash property if false by default
```
petebacondarwin, line 6:
```suggestion
//  trailingSlash property is false by default
```

.github/workflows/pull-request.yml
```diff
@@ -58,8 +58,9 @@ jobs:
         uses: nick-fields/retry@v2
         with:
           max_attempts: 3
-          timeout_minutes: 20
-          command: npm run test:e2e -- -- --log-timestamp --install-wrangler=3.8.0
+          timeout_minutes: 10
+          # Note: the latest wrangler version compatible with pages-e2e is 3.17.0
+          command: npm run test:e2e -- -- --log-timestamp --install-wrangler=3.17.0
```
petebacondarwin, line 63: Are newer versions of wrangler not compatible? How?

petebacondarwin (approved):
A bunch of minor nits. Thanks for improving the confidence and reliability of the tests.

pages-e2e/features/_utils/getAssertVisible.ts
dario-piotrowicz: Yeah sorry I am not really sure, I might have left it there by mistake as I was trying things out 😓 

> The waitFor on the locator can return even if the selected thing is not visible?
yes, after the timeout it would throw if the element is not visible (/present)

dario-piotrowicz: I've removed the `setTimeout`, I don't think it was right, thanks a lot for noticing it! ❤️

packages/next-on-pages/tests/templates/handleRequest.test.ts
dario-piotrowicz: I think we simply haven't had a clear minimal reproduction for this (and I feel that trying to find/create one could be quite time consuming) 🤔 

I'll double check the i18n app as mentioned in #222 maybe I could use that to test the infinite loop 👍

dario-piotrowicz: it doesn't look like the app is generating the issue (if the app mentioned in the issue is the one I think it is), so again I'm not sure if we can reproduce this, we could explore this but again I feel like the extra security around this rare edge case wouldn't be worth all the effort required to do that

@james-elicx, since you fixed the issue back when, do you have any idea of how it could be reproduced?

.github/workflows/pull-request.yml
dario-piotrowicz: It seems like the issue is caused by the pages-e2e testing framework using `--port=0`: https://github.com/cloudflare/workers-sdk/issues/4563

packages/next-on-pages/tests/templates/handleRequest.test.ts
james-elicx: i've been trying to reproduce this evening but had no luck 😭

dario-piotrowicz: no problem, thanks a lot for giving it a try! 🙏

I feels like not testing it is not going to be a huge risk (especially since it's not even something that we can reproduce easily), hopefully you're ok with it?

PR #574, opened 2023-12-05
add option to build .dev.vars file

### LINKED ISSUE #441: [🐛 Bug]: Cannot access to non-public env variables in .env.local from RSC

_No response_

### Description

I have 3 .env files, 

.env.local
.env.production
.env.development

I have some sensitive variables defined in .env.local however after building my app they become undefined in RSC but on dev mode I have access to them, So I assume that it might be related to next-on-pages,

### Reproduction

add something like `RECAPTCHA_SECRET_KEY` to .env.local and build app using next-on-pages and run it and log it in a RSC. it's undefined.

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

dario-piotrowicz:
resolves #441

### DIFF
--- a/packages/next-on-pages/src/additional/dev.vars.ts
+++ b/packages/next-on-pages/src/additional/dev.vars.ts
@@ -1,0 +1,29 @@
+import { validateFile } from '../utils';
+import { resolve } from 'node:path';
+import { readFile, writeFile } from 'node:fs/promises';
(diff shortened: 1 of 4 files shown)


dario-piotrowicz:
Not something we're too likely to proceed.... so I'm closing the PR for the time being instead of leaving it open and stale

PR #582, opened 2023-12-08
Fix pages i18n catch-all issue

### LINKED ISSUE #578: [🐛 Bug]: API routes not taking effect in Pages apps with a root catch-all route and i18n
### Description

In applications using the Pages router with a root optional catch-all route which also have the `i18n` config set, requests to API routes are not handled correctly and are instead redirected to the catch-all route.

For more details: [minimal reproduction](https://github.com/dario-piotrowicz/next-on-pages-pages-i18n-root-catchall-issue-reproduction)

dario-piotrowicz:
resolves #578


packages/next-on-pages/templates/_worker.js/routes-matcher.ts
```diff
@@ -79,6 +79,8 @@ export class RoutesMatcher {
 		this.wildcardMatch = wildcardConfig?.find(
 			w => w.domain === this.url.hostname,
 		);
+
+		this.locales = collectLocales(routes);
```
james-elicx, line 83: btw, I think we should probably move this to build-time and inject them instead of doing it every time we create a new router. I lowkey have been thinking we should do so since I added the locale logic a few months back lol...

dario-piotrowicz: yes that's a good point, we had something else like this and agreed on moving it in the build process, this could be moved there as well, I'll have a look 👍

dario-piotrowicz: I've moved this to build time and introduced a new variable called `__BUILD_METADATA__` to pass this sort of data along, please have a look and let me know what you think 🙂👍

james-elicx: perfect, thank you!

PR #585, opened 2023-12-11
fix app router invalid _error functions

### LINKED ISSUE #413: [🐛 Bug]: Not-found page doesn't work with last vercel version v31.2.0 in edge runtime

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 22.4.0: Mon Mar  6 21:00:41 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8103
        CPU: (8) arm64 Apple M1
        Memory: 16 GB
        Shell: /bin/zsh
Binaries:
        Node: 18.15.0
        Yarn: 1.22.17
        npm: 9.5.0
        pnpm: 8.6.5
Relevant Packages:
        @cloudflare/next-on-pages: 1.4.0
        vercel: N/A
        next: N/A

### Description

Hi guys.
I tried to run `npx @cloudflare/next-on-pages` on Friday and got this error. (Everything was fine a day before).

Screenshot 2023-07-30 at 19 50 32

Vercel version had been up to 31.2.0 this day so maybe that was the problem.

When I tried to add `export const runtime = "edge"; ` to `not-found.tsx` page I've got this error:

Screenshot 2023-07-30 at 19 57 47

not-found.tsx

```javascript
export const runtime = "edge";

export default async function NotFoundPage() {
  return (
    <html lang="en">
      <body>
        <div>test</div>
      </body>
    </html>
  );
}

### Reproduction

_No response_

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

### PR DESCRIPTION
**Pull request by dario-piotrowicz**
in the App router, invalid _error functions can get generated causing the next-on-pages build process to fail, these functions don't seem to be necessary for applications to run as expected so they can be ignored

___

Example of application of this: https://github.com/dario-piotrowicz/next-on-pages-ignore-error-layout-example

___

resolves #413

### CODE REVIEW

**Review by ndodson — APPROVED** [review]

File: packages/next-on-pages/src/buildApplication/getVercelConfig.ts [hunk 1]
```diff
@@ -134,41 +134,26 @@ export function isVercelHandler(route: VercelRoute): route is VercelHandler {
...
-	return noneRoutes.some((route, i) => {
-		const nextRoute = noneRoutes[i + 1];
-
-		if (!isRscRoute(route) || !isRscRoute(nextRoute)) return false;
-
-		if (!route.src.endsWith('/')) return false;
-		if (!route.dest?.endsWith('/index.rsc')) return false;
-		if (!nextRoute.src.endsWith('/((?!.+\\.rsc).+?)(?:/)?$')) return false;
-		if (!nextRoute.dest?.endsWith('/$1.rsc')) return false;
+	return !!vercelConfig.overrides?.['_app.rsc.json'];
```
petebacondarwin, line 144: OK nice! Better than relying upon the special routes.

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -110,6 +114,35 @@ async function tryToFixAppRouterNotFoundRoute({
 	}
 }
 
+/**
+ * In the App router, error boundaries are implemented as client components
+ * (see: https://nextjs.org/docs/app/api-reference/file-conventions/error), meaning that they
+ * should not produce server side logic.
+ *
+ * The Vercel build process can however generate _error.func lambdas (as they are useful in the
+ * Vercel network I'd assume), through experimentation we've seen that those does not seem to be
```
petebacondarwin, line 123:
```suggestion
 * Vercel network I'd assume), through experimentation we've seen that those do not seem to be
```

packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -110,6 +114,35 @@ async function tryToFixAppRouterNotFoundRoute({
...
+ * necessary when building the application with next-on-pages so they should be safe to ignore.
+ *
+ * This function make such invalid _error.func lambdas (if present) ignored (as they would otherwise
```
petebacondarwin, line 126:
```suggestion
 * This function makes such invalid _error.func lambdas (if present) ignored (as they would otherwise
```

pages-e2e/features/_utils/getAssertVisible.ts
```diff
@@ -38,4 +39,5 @@ async function assertVisible(
 			: ''
 	}`;
 	assert(isVisible, `expected ${elementStr} to be visible but it isn't`);
+	return locator;
```
petebacondarwin, line 42: Why not simply `return locator;` at the point where you `break` out of the loop? Then you don't need to define the `locator` outside the loop, and also you can just immediately throw an error if you exit the loop rather than testing again for visibility.

petebacondarwin (approved):

pages-e2e/features/_utils/getAssertVisible.ts
dario-piotrowicz: ah yeah good point! 🙂👍

packages/next-on-pages/src/buildApplication/getVercelConfig.ts
dario-piotrowicz: yeah it's quite less brittle now isn't it? 😃

the only downside (that applies for the Pages router check too) is that it is hard to tell/find Vercel source code that guarantees that this override always exists, but by testing this on application that does seem to be the case, I guess we can see how it works in the wild and change course later if need be

pages-e2e/features/_utils/getAssertVisible.ts
dario-piotrowicz: code updated 🙂

PR #589, opened 2023-12-13
Fix localhost issue for server actions

### LINKED ISSUE #588: [🐛 Bug]: Server actions don't work with `wrangler pages dev`
### Description

Server actions do work fine when using next dev (with or without `next-dev`) and also in production, they do however not work locally when previewing the application with `wrangler pages dev` since there seem to be a discrepancy between the request's origin header which uses `localhost` and the `x-forwarded-host` header which uses `127.0.0.1`.

> [!tip]
> People can still use server actions in local mode if they go to the specific address the worker is running on
> eg. instead of going to `http://localhost:8788/server-actions/simple-kv-forms` they can go to `http://127.0.0.1:xxxxx/server-actions/simple-kv-forms` and there actions should works as expected

### Reproduction

https://github.com/dario-piotrowicz/next-on-pages-server-actions-test/tree/main/next-app

To reproduce the issue:
 - run `npm i`
 - run `npm run pages:build`
 - run `npm run pages:dev`
 - navigate to `http://localhost:8788/server-actions/simple-kv-forms`
 - try to set a new value using the button
 
 This should be the result:
Screenshot 2023-12-12 at 22 50 40

### Additional Information

One solution could be to set, during local development a proper value to [`allowedOrigins`](https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#allowedorigins), the issue is that the config doesn't accept wildcards, and the port which is part of `x-forwarded-host` changes each time, so I think this is not really a feasible solution

Another idea would be to apply some logic in the build process itself so that for server actions we try to convert all potentially localhost urls to a well defined format (as for example `localhost` without a port), I am not a big fan of dev changes that effect the production build output, but this would likely be such a small one that it might be acceptable

dario-piotrowicz:
resolves #588

### DIFF
--- a/.changeset/happy-zoos-shout.md
+++ b/.changeset/happy-zoos-shout.md
@@ -1,0 +1,9 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+make sure that server actions can work in `wrangler pages dev`
+
+fix `wrangler pages dev` breaking server actions because of a misalignment between headers
+where some use `localhost` and others `127.0.0.1` (causing server actions not to work because
+they don't recognize the host origin)
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -497,6 +497,23 @@ function fixFunctionContents(contents: string): string {
 	return contents;
 }
 
+/**
+ * Server actions check the request headers.origin url host and try to make sure that that is
+ * the same as that of other Next.js-specific headers (this is a measure against CSRF attacks)
+ *
+ * This resulted problematic when trying to run the application with `wrangler pages dev` as the
+ * two hosts can get misaligned (one being `localhost:<port>` while the other `127.0.0.1:<port>`)
+ *
+ * This function updates contents of a function file by tweaking the line creating the host variable
+ * that is used for checking with `null` in the case the header.origin's host is `localhost` thus solving
+ * this issue (while preserving the same functionality in production)
+ *
+ * Additionally the function also removes a console.warn that appears when the above value is falsy
+ * (but again the production behavior should remain unaltered)
+ *
+ * @param contents the function file contents to update/fix
+ * @returns the updated function file contents
+ */
 function fixFunctionContentsForLocalhostServerActions(contents: string) {
 	const headerOriginRegex =
 		/"string"==typeof (\w+)\.headers\.origin\?(new URL\(\1\.headers\.origin\)\.host):void 0,/;
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -460,6 +460,8 @@ export type CollectedFunctionIdentifiers = {
  * @returns the updated/fixed contents
  */
 function fixFunctionContents(contents: string): string {
+	contents = fixFunctionContentsForLocalhostServerActions(contents);
+
 	contents = contents.replace(
 		// TODO: This hack is not good. We should replace this with something less brittle ASAP
 		// https://github.com/vercel/next.js/blob/2e7dfca362931be99e34eccec36074ab4a46ffba/packages/next/src/server/web/adapter.ts#L276-L282
@@ -492,6 +494,24 @@ function fixFunctionContents(contents: string): string {
 		';let{originalRequest:$1=$2}=$2;',
 	);
 
+	return contents;
+}
+
+function fixFunctionContentsForLocalhostServerActions(contents: string) {
+	const headerOriginRegex =
+		/"string"==typeof (\w+)\.headers\.origin\?(new URL\(\1\.headers\.origin\)\.host):void 0,/;
+	if (contents.match(headerOriginRegex)) {
+		contents = contents.replace(headerOriginRegex, (fullMatch, _i, origin) => {
+			const originWithCheck = `(/^localhost:\\d{4}$/.test(${origin})?null:${origin})`;
+			const updatedFullMatch = fullMatch.replace(origin, originWithCheck);
+			return `${updatedFullMatch} wranglerPagesDevLocalhostFixApplied = true,`;
+		});
+
+		contents = contents.replace(
+			'console.warn("Missing `origin` header from a forwarded Server Actions request.");',
+			'try { !wranglerPagesDevLocalhostFixApplied && console.warn("Missing `origin` header from a forwarded Server Actions request."); } catch {}',
+		);
+	}
 	return contents;
 }
 
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -526,6 +526,10 @@ function fixFunctionContentsForLocalhostServerActions(contents: string) {
 
 		contents = contents.replace(
 			'console.warn("Missing `origin` header from a forwarded Server Actions request.");',
+			// Note: this is a pretty brittle change and we even introduce a new variable named `wranglerPagesDevLocalhostFixApplied`,
+			//       just in case the source code changes making the variable get out of scope, let's add a try catch here so that in
+			//       the worst case scenario if the variable is not recognized the user code still runs fine (and the only broken thing
+			//       would be that the console.warn would not get called)
 			'try { !wranglerPagesDevLocalhostFixApplied && console.warn("Missing `origin` header from a forwarded Server Actions request."); } catch {}',
 		);
 	}


.changeset/happy-zoos-shout.md
```diff
@@ -0,0 +1,9 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+make sure that server actions can work in `wrangler pages dev`
+
+fix `wrangler pages dev` breaking server actions because of a misalignment between headers
```
james-elicx, line 7: Question: If this is only intended to be for wrangler pages dev, would it be problematic or introduce a security concern if someone deploys the build?

dario-piotrowicz: It shouldn't introduce any problem/security concern, if it would then I don't think the change would be worth it and we could instead ask people to set an appropriate `allowedOrigins` for during local development (it could rely on an `.env.local` file making it, although still inconvenient, not terrible).

It would be a bit cumbersome so I would prefer to try to avoid that if we could 😓

Regarding why I don't think it's introducing a risk, the fact is that by the code below you can see that I am updating a pretty specific line and I add a regex check that strictly changes logic only on hosts like: `localhost:xxxx` so I feel like it's not introducing any risks.

(then we also update a `console.warn` but even if the changes break that I don't think that's not the end of the world)

More than introducing security concerns I am a bit afraid that this is a bit brittle and can easily break... hopefully not but I guess we can only be sure by actually seeing how it performs in the wild 😅

dario-piotrowicz: But if you see a security concern or something that looks problematic to you please let me know 🙏 
(I think this could be quite important for local development but for sure, better to break local dev for users instead of breaking their production applications 😅)

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -495,6 +497,45 @@ function fixFunctionContents(contents: string): string {
...
+ *
+ * @param contents the function file contents to update/fix
+ * @returns the updated function file contents
+ */
+function fixFunctionContentsForLocalhostServerActions(contents: string) {
+	const headerOriginRegex =
+		/"string"==typeof (\w+)\.headers\.origin\?(new URL\(\1\.headers\.origin\)\.host):void 0,/;
+	if (contents.match(headerOriginRegex)) {
+		contents = contents.replace(headerOriginRegex, (fullMatch, _i, origin) => {
+			const originWithCheck = `(/^localhost:\\d{4}$/.test(${origin})?null:${origin})`;
```
petebacondarwin, line 522: Out of interest, why use `null` here when the original code is using `void 0`?
Better to be consistent?

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -495,6 +497,45 @@ function fixFunctionContents(contents: string): string {
 	return contents;
 }
 
+/**
+ * Server actions check the request headers.origin url host and try to make sure that that is
+ * the same as that of other Next.js-specific headers (this is a measure against CSRF attacks)
+ *
+ * This resulted problematic when trying to run the application with `wrangler pages dev` as the
+ * two hosts can get misaligned (one being `localhost:<port>` while the other `127.0.0.1:<port>`)
```
petebacondarwin, line 505: I kind of feel like this is a problem with `wrangler pages dev` no?
Why is there this misalignment? Could you expand on that?

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: no reason actually, it's just what felt more natural to me 😅, since this is just changing the already built and minified code I didn't think it'd make much difference if it were consistent or not

.changeset/happy-zoos-shout.md
james-elicx: Hmm, I'm really not sure I like the idea of doing a modification to something that has a high chance of being deployed to production, when the modification is only supposed to be presented in one single specific dev environment...

I really think that we need a way to determine if the user is using wrangler pages dev if we want to do this kind of hack. These kind of changes/hacks we've done in the past have always been intended to be shipped to production, whereas this change is only intended to be used in wrangler pages dev.

And since the Next.js PR for wildcards got merged, then surely it would be better to leave this to the user if they want to do this in their next.config.js instead. It might take a week for a new Next.js release, but to me that is a better choice than making modifications to their CSRF protections that would be shipped to prod when it shouldn't be.

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: That's a totally valid point, I am not sure if anything can be changes in wrangler but I can check with the team

The problem is that this issue is caused by a fundamental design aspect of the `startDevWorker` changes

Basically `wrangler pages dev` gives you the proxy worker's address (e.g. `http://localhost:8788`) but the actual worker is running on a different port (and not on `localhost`) (e.g. `http://127.0.0.1:51405/`) so, when, with the browser you to go the address `wrangler pages dev` provides you and try to make a request, the host from the browser (which is the proxy worker's address) is different from the actual worker's address (which is not really visible by the user) and that's what causing the misalignment.

So if in the browser I were to go to the actual worker's address (e.g. `http://127.0.0.1:51405/server-actions/simple-kv-forms`) things would just work as expected, the issues there are that:
 - the address is not provided to you by wrangler (the way I find it is by going to the standard proxy address, trying to use action and checking the terminal error and picking up the address from that)
 - the address is not stable and (as far as I know) changes whenever the worker code changes

.changeset/happy-zoos-shout.md
dario-piotrowicz: As I mentioned `allowedOrigins` even when the wildcard change is released, is going to be pretty tedious and error prone in my opinion (maybe having a local address in the as an allowed origin in production is not going to be an issue, so people could just add it there and leave it... I am not really sure). Ideally I would really not want developers to have to deal with it themselves.

Regarding detecting if the build output is going to be used for `wrangler pages dev` I think that's not really a great idea based on how the adapter works. Since it just builds the worker for you, I don't think it should have any knowledge on how you are going to use that worker. Actually, I can imagine people building their worker with next-on-pages, previewing it with `wrangler pages dev` and then deploying it with `wrangler pages deploy`, so the same exact worker should work in both dev and production (and it should be the same worker since people are trying to validate it before the deployment, if we produced different workers then people could not actually accurately "preview" the actual production worker before deploying it).

I think it is very clear that modifying production code for the sake of local development is not ideal in the slightest, I just think that the changes here are so negligible (both in terms of performance and security) that it's worth it in this case given the DX gain.

james-elicx: > I just think that the changes here are so negligible (both in terms of performance and security)

But this is a limitation that Next.js specifically added for production builds to help prevent CSRF attacks, so I would really have to disagree that this is negligible. They wouldn't go through the effort to add this additional protection if there wasn't any value in it.

If we are modifying it to allow localhost:{num} to be bypass this protection in any deployment, then IMO that is a very slippery slope and could end up being consequential for some of our users.

> Regarding detecting if the build output is going to be used for wrangler pages dev I think that's not really a great idea based on how the adapter works

I meant through something like an env variable existing from wrangler to dictate that this was in pages dev as a *gate* for when this hack works in their code. something like `process.env.WRANGLER_DEV && /.../.test(...) ? null : origin`, but i dont know if that is currently possible with wrangler. this would then give the desired behaviour in local dev, and the correct behaviour/protections in deployments.

dario-piotrowicz: > I meant through something like an env variable existing from wrangler to dictate that this was in pages dev as a _gate_ for when this hack works in their code. something like `process.env.WRANGLER_DEV && /.../.test(...) ? null : origin`, but i dont know if that is currently possible with wrangler. this would then give the desired behaviour in local dev, and the correct behaviour/protections in deployments.

ah ok, yeah that sounds like an interesting idea, I misunderstood what you were suggesting 🙂

dario-piotrowicz: anyways thanks a lot for the insights 🙂

but as we were discussing this with @petebacondarwin, it sounds like maybe things can be updated in wrangler so to have it set a proper `x-forwarded-host` when passing requests from the proxy worker to the actual worker without any hacky changes needed in next-on-pages

so I'm going to close the PR for now and see if things can be changes in wrangler instead, if not we can revisit things here later 🙂

dario-piotrowicz:
Closing this PR since, as discussed with @petebacondarwin, it sounds like maybe the proper fix for this issue needs to be applied in wrangler itself (having it set a proper `x-forwarded-host` on requests send from the proxy worker to the actual worker)

.changeset/happy-zoos-shout.md
james-elicx: ah nice, that actually sounds like a pretty good middleground!

PR #597, opened 2023-12-17
enable ai binding in next-dev module

dario-piotrowicz:


dario-piotrowicz:
Closing this PR as we should probably just add this to wrangler's `getBindingsProxy` (https://github.com/cloudflare/workers-sdk/pull/4523) and use that here instead.

PR #599, opened 2023-12-18
Add gotchas documentation

### LINKED ISSUE #598: [🐛 Bug]: Using `headers` or `cookies` in the root layout fail the build
### Description

If I create a new App router application and use the edge runtime in the root layout file, everything works as intended (thanks to #585 which causes the deletion of the extra `_error` lambda).

If however the layout uses `headers` or `cookies` from `next/headers` then the next-on-pages build process fails because an extra `_not-found` lambda gets generated (an `_error` lambda is also generated but again that's not an issue thanks to #585).

### Reproduction

https://github.com/dario-piotrowicz/next-on-pages-issue-598-reproduction

dario-piotrowicz:
resolves #598


jculvey (approved):

sanketnaik99:
@dario-piotrowicz Hey! I've been trying to implement dynamic routes for my Next.js pages project using the app router but my builds keep failing on Cloudflare. 

When I try to use the edge runtime I get a segmentation fault. 

Should it be possible to use dynamic routes if I use the edge runtime or is that not supported on Cloudflare at all?

dario-piotrowicz:
Hi @sanketnaik99, yeah dynamic routes should be supported with no issues

I am not sure what problem you're facing, please open a [new issue](https://github.com/cloudflare/next-on-pages/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml&title=%5B%F0%9F%90%9B+Bug%5D%3A+) regarding that and we can discuss further there 🙂
(please try to provide a minimal reproduction as that does help massively 🙏)

___

PS: some people have had issues building their next-on-pages applications using the Pages git integration and they have had segfaults caused by the `next build` process (we haven't been able to look into it as we lack code to reproduce the issue), if that's your case please also try to deploy the application manually using `wrangler pages deploy .vercel/output/static`.

sanketnaik99:
@dario-piotrowicz Thanks! Yeah, it has something to do with my application because it's a huge codebase and I'm migrating it from pages to app router. I'll keep digging into it, but I got dynamic routes to work in a smaller application.

Once again, thanks for your quick reply!

dario-piotrowicz:
@sanketnaik99 not a problem 🙂 

yes, please keep digging into it and feel free to open an issue whenever you want if you get stuck as we are always keen to help 😄 

Although if you have a big app and cannot share its code there might be little we can do to help 😓

PR #612, opened 2024-01-04
Stabilize the next-dev API

### LINKED ISSUE #271: [⚡ Feature]: Mocking `process.env.<BINDING>` when running `next dev`

### LINKED ISSUE #527: Finalize `next-dev`'s `DevBindingsOptions` API (and remove `__experimental__` prefix)
In https://github.com/cloudflare/next-on-pages/pull/486 we've added the `next-dev` submodule for adding local bindings to next dev.

The submodule exports the [`setupDevBindings` function](https://github.com/cloudflare/next-on-pages/blob/457287359bca098b8ca3cb0b10655addc846578b/internal-packages/next-dev/src/index.ts#L14) which accepts a [`DevBindingsOptions`](https://github.com/cloudflare/next-on-pages/blob/457287359bca098b8ca3cb0b10655addc846578b/internal-packages/next-dev/src/index.ts#L27) object for the various options.

Most of the options there follow [Miniflare's API](https://github.com/cloudflare/workers-sdk/tree/main/packages/miniflare#api) with the exception of `textBindings` (which in Miniflare is just `bindings`), adding a new term there could be worse/more confusing then just sticking with the (potentially confusing) `bindings` term (see https://github.com/cloudflare/next-on-pages/pull/486#discussion_r1368980906).

So before our next release we need to decide how the `DevBindingsOptions` object needs to look like, the alternatives are:
 - leave it as is (as it is clear enough)
 - leave it as is but replace `textBindings` with `bindings` (to align it more with Miniflare)
 - come up with some new structure that could make things more clear/intuitive (for example passing the bindings in an object structured like)

dario-piotrowicz:
Example/Demo of this new API's usage: https://github.com/dario-piotrowicz/next-on-pages-stable-next-dev-test-demo

resolves #527 
resolves #271

### DIFF
--- a/.changeset/silver-kings-attend.md
+++ b/.changeset/silver-kings-attend.md
@@ -1,0 +1,6 @@
+---
+'@cloudflare/next-on-pages-next-dev': patch
+'@cloudflare/next-on-pages': patch
+---
+
+Update the `next-dev`'s `setupDevBindings` API and remove the module's `__experiemtnal__` prefix
--- a/.changeset/silver-kings-attend.md
+++ b/.changeset/silver-kings-attend.md
@@ -1,5 +1,4 @@
 ---
-'@cloudflare/next-on-pages-next-dev': patch
 '@cloudflare/next-on-pages': patch
 ---
 
--- a/.changeset/silver-kings-attend.md
+++ b/.changeset/silver-kings-attend.md
@@ -3,4 +3,4 @@
 '@cloudflare/next-on-pages': patch
 ---
 
-Update the `next-dev`'s `setupDevBindings` API and remove the module's `__experiemtnal__` prefix
+Update the `next-dev`'s `setupDevBindings` API and remove the module's `__experimental__` prefix
--- a/internal-packages/next-dev/README.md
+++ b/internal-packages/next-dev/README.md
@@ -2,7 +2,7 @@
 
 The `next-dev` submodule of the `@cloudflare/next-on-pages` package implements a utility that allows you to run the [standard Next.js development server](https://nextjs.org/docs/app/api-reference/next-cli#development) (with hot-code reloading, error reporting, HMR and everything it has to offer) with also adding local Cloudflare bindings simulations (implemented via [Miniflare](https://github.com/cloudflare/miniflare)).
 
-IMPORTANT: As mentioned above the module allows you to run the standard Next.js dev server as is and it only makes sure that Cloudflare bindings are accessible, it does not generate a worker nor faithfully represent the final application that will be deployed to Cloudflare Pages, so please use this only as a development tool and make sure to properly test your application with `wrangler pages dev` before actually deploying it to.
+IMPORTANT: As mentioned above the module allows you to run the standard Next.js dev server as is and it only makes sure that Cloudflare bindings are accessible, it does not generate a worker nor faithfully represent the final application that will be deployed to Cloudflare Pages, so please use this only as a development tool and make sure to properly test your application with `wrangler pages dev` before actually deploying it to Cloudflare.
 
 ## How to use the module
 
--- a/internal-packages/next-dev/README.md
+++ b/internal-packages/next-dev/README.md
@@ -1,8 +1,5 @@
 # Next-on-pages Next-Dev
 
-> **Warning**
-> The submodule API needs to be finalized, in the meantime the submodule is being exported as `__experimental__next-dev`, the `__experimental__` prefix will be removed in a future release
-
 The `next-dev` submodule of the `@cloudflare/next-on-pages` package implements a utility that allows you to run the [standard Next.js development server](https://nextjs.org/docs/app/api-reference/next-cli#development) (with hot-code reloading, error reporting, HMR and everything it has to offer) with also adding local Cloudflare bindings simulations (implemented via [Miniflare](https://github.com/cloudflare/miniflare)).
 
 IMPORTANT: As mentioned above the module allows you to run the standard Next.js dev server as is and it only makes sure that Cloudflare bindings are accessible, it does not generate a worker nor faithfully represent the final application that will be deployed to Cloudflare Pages, so please use this only as a development tool and make sure to properly test your application with `wrangler pages dev` before actually deploying it to.
--- a/internal-packages/next-dev/package.json
+++ b/internal-packages/next-dev/package.json
@@ -20,7 +20,7 @@
 		"devBindingsOptions.ts"
 	],
 	"dependencies": {
-		"miniflare": "^3.20231030.2",
+		"miniflare": "3.20231002.0",
 		"node-fetch": "^3.3.2"
 	},
 	"devDependencies": {
--- a/internal-packages/next-dev/src/durableObjects.ts
+++ b/internal-packages/next-dev/src/durableObjects.ts
@@ -116,7 +116,7 @@ function getRegisteredWorkersWithDOs(registeredWorkers: WorkerRegistry) {
 
 	for (const workerName of Object.keys(registeredWorkers ?? {})) {
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-		const worker = registeredWorkers![workerName]!;
+		const worker = registeredWorkers[workerName]!;
 		const containsDOs = !!worker.durableObjects.length;
 		if (containsDOs) {
 			registeredWorkersWithDOs.set(workerName, worker);
--- a/internal-packages/next-dev/src/durableObjects.ts
+++ b/internal-packages/next-dev/src/durableObjects.ts
@@ -47,7 +47,8 @@ export async function getDOBindingInfo(
 
 	if (!registeredWorkers) {
 		warnAboutLocalDurableObjectsNotFound(
-			new Set(requestedDurableObjectBindings.keys()));
+			new Set(requestedDurableObjectBindings.keys()),
+		);
 		return;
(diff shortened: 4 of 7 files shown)


internal-packages/next-dev/package.json
```diff
@@ -20,7 +20,7 @@
 		"devBindingsOptions.ts"
 	],
 	"dependencies": {
-		"miniflare": "^3.20231030.2",
+		"miniflare": "3.20231002.0",
```
dario-piotrowicz, line 23: - [ ] Create a minimal reproduction of the regression and open a new issue regarding this in workers-sdk

internal-packages/next-dev/src/durableObjects.ts
```diff
@@ -103,7 +105,8 @@ export async function getDOBindingInfo(
 				[externalDO.durableObjectName]: externalDO,
 			};
 		},
-		{} as DevBindingsOptions['durableObjects'],
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dario-piotrowicz, line 108:
```suggestion
```

internal-packages/next-dev/src/durableObjects.ts
```diff
@@ -9,18 +9,20 @@ import {
...
 
+export type DevBindingsDurableObjectOptions = Record<string, Extract<Binding, { type: 'durable-object' }>>;
+
 /**
  * Gets information regarding DurableObject bindings that can be passed to miniflare to access external (locally exposed in the local registry) Durable Object bindings.
  *
  * @param durableObjects
  * @returns the durableObjects and WorkersOptions objects to use or undefined if connecting to the registry and/or creating the options has failed
  */
 export async function getDOBindingInfo(
```
dario-piotrowicz, line 20: this function's logic is a bit off, only the binding name is used to match the DO of external workers and not the scriptName nor the class name is used, not sure how I made such a blunder 😓, but it needs fixing

dario-piotrowicz: amended

irvinebroque (commented):
Progress! Ship it!

.changeset/silver-kings-attend.md
```diff
@@ -0,0 +1,6 @@
+---
+'@cloudflare/next-on-pages-next-dev': patch
+'@cloudflare/next-on-pages': patch
+---
+
+Update the `next-dev`'s `setupDevBindings` API and remove the module's `__experiemtnal__` prefix
```
petebacondarwin, line 6:
```suggestion
Update the `next-dev`'s `setupDevBindings` API and remove the module's `__experimental__` prefix
```

internal-packages/next-dev/README.md
```diff
@@ -1,8 +1,5 @@
 # Next-on-pages Next-Dev
 
-> **Warning**
-> The submodule API needs to be finalized, in the meantime the submodule is being exported as `__experimental__next-dev`, the `__experimental__` prefix will be removed in a future release
-
 The `next-dev` submodule of the `@cloudflare/next-on-pages` package implements a utility that allows you to run the [standard Next.js development server](https://nextjs.org/docs/app/api-reference/next-cli#development) (with hot-code reloading, error reporting, HMR and everything it has to offer) with also adding local Cloudflare bindings simulations (implemented via [Miniflare](https://github.com/cloudflare/miniflare)).
 
 IMPORTANT: As mentioned above the module allows you to run the standard Next.js dev server as is and it only makes sure that Cloudflare bindings are accessible, it does not generate a worker nor faithfully represent the final application that will be deployed to Cloudflare Pages, so please use this only as a development tool and make sure to properly test your application with `wrangler pages dev` before actually deploying it to.
```
petebacondarwin, line 5: before actually deploying it to...?

```suggestion
IMPORTANT: As mentioned above the module allows you to run the standard Next.js dev server as is and it only makes sure that Cloudflare bindings are accessible, it does not generate a worker nor faithfully represent the final application that will be deployed to Cloudflare Pages, so please use this only as a development tool and make sure to properly test your application with `wrangler pages dev` before actually deploying it to Cloudflare.
```

internal-packages/next-dev/src/durableObjects.ts
```diff
@@ -112,20 +119,15 @@ export async function getDOBindingInfo(
...
-	registeredWorkers: WorkerRegistry,
-	durableObjectsNames: Set<string>,
-) {
+function getRegisteredWorkersWithDOs(registeredWorkers: WorkerRegistry) {
 	const registeredWorkersWithDOs: Map<string, WorkerRegistry[string]> =
 		new Map();
 
 	for (const workerName of Object.keys(registeredWorkers ?? {})) {
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 		const worker = registeredWorkers![workerName]!;
```
petebacondarwin, line 128:
```suggestion
		const worker = registeredWorkers[workerName]!;
```

petebacondarwin (changes requested):
Looking good. Let's fix up these things I note below and then confirm the config property names with Rahul before landing.

internal-packages/next-dev/package.json
dario-piotrowicz: PS: I did try to create a minimal Miniflare reproduction by using standard DOs there but I couldn't really reproduce this, so I _think_ that this is related to both Miniflare and the external DOs worker we generate here 🤔 😓

internal-packages/next-dev/src/durableObjects.ts
```diff
@@ -41,31 +46,33 @@ export async function getDOBindingInfo(
...
-				found = !!worker.durableObjects.find(
-					durableObject => durableObject.name === durableObjectName,
-				);
+			for (const [workerName, worker] of registeredWorkersWithDOs.entries()) {
+				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+				const durableObject = durableObjects[durableObjectBindingName]!;
+				if (workerName === durableObject.service.name) {
+					found = !!worker.durableObjects.find(
+						durableObject =>
+							durableObject.className === durableObject.className,
```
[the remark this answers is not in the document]
dario-piotrowicz: > This will always be true!

Yes, such a silly mistake! 😓 
(basically I reused the same variable name and didn't notice it 🤦)

___

> I think this whole block of code could be simpler.
Thanks, I've applied the suggested change, I agree that it looks better 😄 
(I was going a bit overboard with the `reduce` there 🙈)

internal-packages/next-dev/README.md
dario-piotrowicz: 🤦👍

petebacondarwin (approved):
Approved - subject to the API discussion, especially around `localStoragePersistence`.

internal-packages/next-dev/package.json
mrbbot: Interesting, thanks for the ping. @dario-piotrowicz do you have any details about the error/type of failure? What problems were you having?

dario-piotrowicz: @mrbbot this [demo/test app](https://github.com/dario-piotrowicz/next-on-pages-stable-next-dev-test-demo) renders the following home page:
Screenshot 2024-01-08 at 12 19 31

If I however use the latest miniflare version this is what I get instead:
Screenshot 2024-01-08 at 12 18 39

I guess that it might be coming from here:
https://github.com/cloudflare/next-on-pages/blob/f0df948a8ed3df15f20259a45b5d60ddb919dfa8/internal-packages/next-dev/src/wrangler.ts#L68

I'm not sure why this is anyways, but it does indeed only happen with the latest miniflare versions 😕 

(Also I just noticed that the var is not showing up! 😮)

internal-packages/next-dev/src/index.ts
```diff
@@ -207,3 +215,94 @@ function shouldSetupContinue(): boolean {
...
+	WorkerOptions,
+	| 'bindings'
+	| 'kvNamespaces'
+	| 'durableObjects'
+	| 'r2Buckets'
+	| 'd1Databases'
+	| 'serviceBindings'
+>;
+
+async function getMiniflareBindingOptions(
```
dario-piotrowicz, line 229: @RamIdeas this is the mapping function 🙂 

but anyways this function is just temporary, as soon as `getBidingsProxy` is released, most of the `next-dev` code (including this function) will just go away and replaced with a call to the utility 🙂

(there will not be any user facing change as the API should be the same 🙂)

internal-packages/next-dev/package.json
dario-piotrowicz: I've created a new issue in the next-on-pages repo to keep track of this: https://github.com/cloudflare/next-on-pages/issues/621

PR #630, opened 2024-01-10
Fix Durable Object bindings proxy with new Miniflare versions

### LINKED ISSUE #621: Chore: Explore why the latest Miniflare versions don't work with the `next-dev` sub-module
### Description

In #612 we've downgraded and pinned the Miniflare's version used in the `next-dev` sub-module

We need to properly investigate the issue and:
 - if the issue comes (or can be easily fixed) from the `next-dev` code we should change our code
 - if the issue comes from miniflare itself we should create an issue in the workers-sdk repo, have the package fixed and bump its dependency here

___

For more context and a reproduction of the issue see: https://github.com/cloudflare/next-on-pages/pull/612#discussion_r1441563854

mrbbot:
Hey! 👋 Tried to debug #621 this afternoon. I think PR this should fix things. I'm still a bit confused why this issue was only exposed with new Miniflare versions, but this seems to be working for me locally.

The issue was that `createClass()` was being called with a class name like `<workerName>_<className>` as opposed to just `<className>`. This meant in the target `wrangler dev` session containing the Durable Object, we were trying to `env["<workerName>_<className>"]` instead of `env["<className>"]` resulting in the `Cannot read properties of undefined` error.

___

resolves #621

### DIFF
--- a/.changeset/witty-kids-repair.md
+++ b/.changeset/witty-kids-repair.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+chore: bump to `miniflare@3.20231218.1`
--- a/internal-packages/next-dev/src/durableObjects.ts
+++ b/internal-packages/next-dev/src/durableObjects.ts
@@ -99,7 +99,11 @@ export async function getDOBindingInfo(
 		(all, externalDO) => {
 			return {
 				...all,
-				[externalDO.durableObjectName]: externalDO,
+				[externalDO.durableObjectName]: {
+					className: externalDO.identifier,
+					scriptName: externalDO.scriptName,
+					unsafeUniqueKey: externalDO.unsafeUniqueKey,
+				},
 			};
 		},
 		{} as WorkerOptions['durableObjects'],
@@ -150,7 +154,8 @@ function collectExternalDurableObjects(
 				return {
 					workerName,
 					durableObjectName: name,
-					className: getIdentifier(`${workerName}_${className}`),
+					className: className,
+					identifier: getIdentifier(`${workerName}_${className}`),
 					scriptName: EXTERNAL_DURABLE_OBJECTS_WORKER_NAME,
 					unsafeUniqueKey: `${workerName}-${className}`,
 				};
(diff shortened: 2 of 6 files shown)


dario-piotrowicz:
@mrbbot Thanks a lot for looking into this!!! ❤️ 

Given the fact that things work with a version of Miniflare but not others I didn't think the bug could be in next-on-pages, how odd 😕 🤷 

___

I've run your changes locally and they do seem to be doing the trick! However there is one thing that is inconsistent between the current `next-dev` behavior and the behavior I get from your changes, I can only think of Miniflare being the culprit 😕

Basically it seems like currently we keep newline characters in DO messages but with your changes we loose them.

I've seen this in this demo application: https://github.com/dario-piotrowicz/next-on-pages-stable-next-dev-test-demo

More specifically from this code here ([source](https://github.com/dario-piotrowicz/next-on-pages-stable-next-dev-test-demo/blob/cc8465ed92fd19ff758a7649b36e709d9f35a1d9/workers/do-worker/src/index.ts#L15)): 
```ts
   return new Response(`Response from DO: \n\n\n${txt}`);
```

This is what I get currently vs with your changes:
| current `next-dev` | `next-dev` with latest miniflare |
|--------|--------|
|Screenshot 2024-01-14 at 16 28 55|Screenshot 2024-01-14 at 16 27 20 | 

do you know the reason for this? I don't suspect the latest version (without the `\n`s) to be correct, or is it?

internal-packages/next-dev/src/durableObjects.ts
```diff
@@ -198,10 +198,17 @@ function generateDurableObjectProxyWorkerScript(
 type RegisteredWorkersWithDOs = Map<string, WorkerDefinition>;
 
 type ExternalDurableObject = {
+	/** Name of the proxied-to worker containing this Durable Object */
```
dario-piotrowicz, line 201: thanks so much for clarifying things here 🙂🙏

.changeset/witty-kids-repair.md
```diff
@@ -0,0 +1,6 @@
+---
+'@cloudflare/next-on-pages-next-dev': patch
```
dario-piotrowicz, line 2:
```suggestion
```

I don't think we should include change logs for `next-on-pages-next-dev` as it is not a package that the user sees anyways, that's my bad as I didn't exclude it from changesets, I'll do that now, could you remove this anyways? 🙂

mrbbot: Done! ✅

.changeset/witty-kids-repair.md
```diff
@@ -0,0 +1,6 @@
...
+'@cloudflare/next-on-pages': patch
+---
+
+chore: bump to `miniflare@3.20231218.1`
```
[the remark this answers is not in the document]
dario-piotrowicz: Awesome! thanks 😄

dario-piotrowicz (approved):
LGTM, Thanks a lot again for this 😄

Issue #623, opened 2024-01-08
Chore: use `getBindingsProxy` in the `next-dev` sub-module


dario-piotrowicz:
### Description

Currently the [next-dev sub-module](https://github.com/cloudflare/next-on-pages/tree/main/internal-packages/next-dev) implements its own logic for getting access to binding proxies from Miniflare, it should (for obvious reasons) instead use the `getBidingsProxy` utility (being implemented in https://github.com/cloudflare/workers-sdk/pull/4523) as soon as it is available

PR #633, opened 2024-01-12
docs: update next-dev README to include link to available binding types

admah:


IgorMinar (approved):
lgtm. the line numbers will likely change over time, but this type needs to be documented in wrangler docs as part of getBindingsProxy anyway we can keep it as is and update the link here then.

IgorMinar:
@admah can you please fix the prettier check failure?

internal-packages/next-dev/README.md
```diff
@@ -58,6 +58,9 @@ if (process.env.NODE_ENV === 'development') {
 }
 ```
 
+
+> View [this file](https://github.com/cloudflare/next-on-pages/blob/main/internal-packages/next-dev/src/index.ts#L31-L60) for a list of available binding types.
```
dario-piotrowicz, line 62:
```suggestion
> View [these types](https://github.com/cloudflare/next-on-pages/blob/main/internal-packages/next-dev/src/index.ts#L31-L60) for a list of available binding types.
```

or something along those lines as user should not care about the whole file no? 🤔

dario-piotrowicz:
> the line numbers will likely change over time

For that reason I always go with permalinks, so that the link itself will always remain valid, wouldn't then something like
`https://github.com/cloudflare/next-on-pages/blob/8e93067/internal-packages/next-dev/src/index.ts#L31-L60`
be better?

dario-piotrowicz:
I am actually not sure of this change... I think that presenting the bindings in the md file instead of pointing users to a ts file would be a very simple, quick and nicer alternative to this 🤔 (we could also add docs links to the binding types, etc...)

> [!NOTE]
> You can notice that the binding types lack ts-docs, I didn't add them since as Igor mentioned, the logic here [should be replaced with `getBindingsProxy`](https://github.com/cloudflare/next-on-pages/issues/623) soon

@admah if you wouldn't mind I could give that a go?

dario-piotrowicz:
I'm closing this PR since we're moving away from the `setupDevBindings` and the types references by the PR (see: https://github.com/cloudflare/next-on-pages/pull/648)

I'll make sure to update the next-on-pages docs with all the available bindings 🙂

Or more likely I'll just refer to the `getBindingsProxy` documentation 🤔

I hope that works for you @admah 🙂🙏

PR #640, opened 2024-01-15
add assertions for making sure that the required `setupDevBindings` options are provided

dario-piotrowicz:
I just noticed that providing an `options` object without a `bindings` field results in a terminal error:
Screenshot 2024-01-15 at 13 30 50

The error is not really very clear so instead of it here I am simply adding assertions and throwing more clear errors for users:
Screenshot 2024-01-15 at 17 43 13
Screenshot 2024-01-15 at 17 43 39

> [!NOTE]
> Throwing makes the error show up in the console but does not kill the nodejs process, I don't think I can kill the dev server process from within `setupDevBindings`

### DIFF
--- a/internal-packages/next-dev/src/index.ts
+++ b/internal-packages/next-dev/src/index.ts
@@ -18,6 +18,14 @@ export async function setupDevBindings(
 	const continueSetup = shouldSetupContinue();
 	if (!continueSetup) return;
 
+	if (!options) {
+		throwError('No options provided to setupDevBindings');
+	}
+
+	if (!options.bindings) {
+		throwError("The provided options object doesn't include a bindings field");
+	}
+
 	const mf = await instantiateMiniflare(options);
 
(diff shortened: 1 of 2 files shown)


internal-packages/next-dev/src/index.ts
```diff
@@ -73,7 +73,7 @@ async function instantiateMiniflare(
 	};
 
 	const devBindingsDurableObjectOptions = Object.fromEntries(
-		Object.entries(options.bindings).filter(
+		Object.entries(options.bindings ?? {}).filter(
```
petebacondarwin, line 76: If it is possible to have no bindings in the options, then we should update the type of `DevBindingsOptions`.

dario-piotrowicz: as we discussed I added assertions to make sure that the `options` structure is what we expect 🙂👍

petebacondarwin (approved):

PR #642, opened 2024-01-17
add `databaseId` to the `setupDevBindings` D1 binding type

dario-piotrowicz:
Result in case only the `databaseName` is used:
Screenshot 2024-01-17 at 16 10 08

### DIFF
--- a/internal-packages/next-dev/src/index.ts
+++ b/internal-packages/next-dev/src/index.ts
@@ -50,7 +50,7 @@ export type DevBindingsOptions = {
 export type Binding =
 	| { type: 'kv'; id: string }
 	| { type: 'r2'; bucketName: string }
-	| { type: 'd1'; databaseName: string }
+	| { type: 'd1'; databaseName: string; databaseId?: string }
 	| { type: 'durable-object'; className: string; service: ServiceDesignator }
 	| { type: 'service'; service: ServiceDesignator }
 	| { type: 'var'; value: string | Json };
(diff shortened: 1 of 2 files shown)


markdembo:
@dario-piotrowicz did a functional test.
- Just using databaseName shows the warning as intended
- Additionally specifying the databaseId removes that warning and allows for local access to the specified local D1 instace

LGTM 🙏

dario-piotrowicz:
@markdembo Awesome! thanks a lot for checking this! 🙏

internal-packages/next-dev/src/index.ts
```diff
@@ -50,7 +50,7 @@ export type DevBindingsOptions = {
 export type Binding =
 	| { type: 'kv'; id: string }
 	| { type: 'r2'; bucketName: string }
-	| { type: 'd1'; databaseName: string }
+	| { type: 'd1'; databaseName: string; databaseId?: string }
```
petebacondarwin, line 53: Should we make the `databaseName` field optional?

petebacondarwin (approved):

internal-packages/next-dev/src/index.ts
dario-piotrowicz: mh.... yeah good point! 😕 

Yes let's! 🙂

petebacondarwin: I wondered about that approach but was worried it would complicate the bindings type too much.
If you are happy with it then that's good for me.

dario-piotrowicz: To me it feels like a more sensible type less likely to confuse people 🤔 (hopefully), since otherwise it would be less clear if and when users should provide the db name 🤔

I'd go with this since you're ok with it 😄

PR #657, opened 2024-02-01
set package.json engines and nvmrc to 20.10.0 (to match the node version used in CI)

dario-piotrowicz:


IgorMinar (approved):
LGTM! Thanks @dario-piotrowicz

PR #658, opened 2024-02-03
feat: add the `cf` object to the request object

james-elicx:
This PR does the following:
- Adds the `cf` object to the request object that is accessible in API routes and middleware and so on.

Demo: https://purple-waterfall-0bcc.pages.dev/api/hello

Usage: https://github.com/james-elicx/nop-req-cf/blob/main/app/api/hello/route.ts
TypeScript Types: https://github.com/james-elicx/nop-req-cf/blob/main/env.d.ts

The value is added to the following two objects in the generated function's code, which results in it being on the request object passed to a route handler.

image

image

### DIFF
--- a/.changeset/proud-zebras-remember.md
+++ b/.changeset/proud-zebras-remember.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Add the `cf` object to the `request` object that is passed to API routes.
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -492,6 +492,16 @@ function fixFunctionContents(contents: string): string {
 		';let{originalRequest:$1=$2}=$2;',
 	);
 
+	// This adds the `cf` object back to the `request` object used in API routes.
+	contents = contents.replace(
+		/(ip:(\w+)\.request\.ip)/gm,
+		'$1,cf:$2.request.cf',
+	);
+	contents = contents.replace(
+		/(ip:\w+\((\w+)\.headers,\w+\.Ip\))/gm,
+		'$1,cf:$2.cf',
+	);
+
 	return contents;
 }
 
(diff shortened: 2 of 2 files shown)


dario-piotrowicz:
Thanks for the PR @james-elicx 🙂 

Your changes however conflict with what the team planned on doing, we were thinking of implementing a helper function for `env`, `ctx` and `cf` (and we're laying the ground work for it), similar to what's been done in https://github.com/cloudflare/next-on-pages/pull/450

Implementing our utility would also not require brittle code regex replacements (which we've used in the past, but only when absolutely necessary).

Moreover people would not have to manually declare the `cf` type (which is not terrible but not the best either).

I'll bring this to the team and see what people think 👍 

In the meantime, I 'm wondering have you considered how this could be used in the `next-dev` submodule? and does your suggested approach only work on API routes/middlewares? or for components as well?

james-elicx:
> Your changes however conflict with what the team planned on doing, we were thinking of implementing a helper function for `env`, `ctx` and `cf` (and we're laying the ground work for it), similar to what's been done in #450
> 
> Implementing our utility would also not require brittle code regex replacements (which we've used in the past, but only when absolutely necessary).

Well, you already know my objections to most of that helper functions discussion we had 😅.

`env` is already accessible via the standard `process.env`.
`cf` should just be made to be accessible on the request object as that is the standard way of using it for cloudflare stuff.
`waitUntil` and `passThroughOnException` are already accessible in middleware but yeah, they probably need a utility to be accessible in api route handlers.

> In the meantime, I 'm wondering have you considered how this could be used in the `next-dev` submodule?

I didn't consider dev mode because the Next.js geo location info does not get populated in dev mode either.

> and does your suggested approach only work on API routes/middlewares? or for components as well?

It only works for API routes and middleware, which are the only places you're supposed to be allowed to access the request object. This is just restoring normal behaviour to the request object that you get when deploying other stuff to Cloudflare. The request object is not available to components, so components won't be able to interact with it.

dario-piotrowicz:
@james-elicx I did check with the team and we agreed that we'd prefer to continue with the utility strategy I mentioned above

Compared to your proposed solution the utility would:
 - not require brittle regex code replacement
 - have the `cf` object already typed without users needing to do that manually
 - work with `next dev`
 - work with any server side code (pages `getServerSideProps`, RSC, etc...)
 - use imports which seem aligned with the Next.js DX (e.g. importing `headers` from `next/headers`)

Hopefully this works for you, please let me know 🙂 
 
In either case if you want we can keep this PR open to compare it with the utility when I open the PR for it (in the next few days) 👍

PR #659, opened 2024-02-06
Get request context

dario-piotrowicz:
TODO:
 - [ ] add an example to this PR's description
 - [ ] add the proper documentation for this
 - [ ] wait for `getBindingsProxy`'s `cf` and `ctx`
 - [ ] improve the `Env` typing


pages-e2e/features/pagesGetRequestContext/assets/wrangler.toml
```diff
@@ -0,0 +1,6 @@
+[vars]
+MY_TOML_VAR = "my var from wrangler.toml"
+
+[[kv_namespaces]]
+binding = "MY_TOML_KV"
+id = "???"
```
petebacondarwin, line 6: ??? ?

pages-e2e/package.json
```diff
@@ -4,7 +4,7 @@
 	"scripts": {
 		"//test:e2e 1": "we start with npm i because this fixes an existing bug in the pages-e2e-test-runner-cli",
 		"//test:e2e 2": "which fails to detect test suites in the files after any run of the test-runner cli",
-		"test:e2e": "npm i ; npx @cfpreview/pages-e2e-test-runner-cli --environment=local"
+		"test:e2e": "npm i ; npx @cfpreview/pages-e2e-test-runner-cli  --install-wrangler=3.17.0 --environment=local"
```
petebacondarwin, line 7: Is this basically moved from the GH Action above?
Is there no later version that we can use?

petebacondarwin (approved):

pages-e2e/package.json
dario-piotrowicz: Just not to get out of scope (/delay the PR) here I figured I wouldn't try to increase this to the latest in this PR

If it's ok I will look into it separately 🙂 

PS: I did briefly try with `latest`/`^3.28.0` but it seems to not work properly, I'll need to investigate that and potentially open an issue in this repo and/or workers-sdk

packages/next-on-pages/src/buildApplication/generateGlobalJs.ts
```diff
@@ -11,10 +11,11 @@ export function generateGlobalJs(): string {
 		})
 		.catch(() => null);
 
-		const __ENV_ALS_PROMISE__ = import('node:async_hooks').then(({ AsyncLocalStorage }) => {
+		const __ALSes_PROMISE__ = import('node:async_hooks').then(({ AsyncLocalStorage }) => {
```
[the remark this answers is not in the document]
dario-piotrowicz: nono it is! 🙂 

https://github.com/cloudflare/next-on-pages/blob/963e1993d5c07e8316f1da196c9da3189dad1e66/packages/next-on-pages/templates/_worker.js/index.ts#L27

Basically this global js code gets added at the top of the worker template js and then we use the defined variable there

pages-e2e/package.json
petebacondarwin: No problem. Would be good to find out why it is not working with later versions. I think we introduced the new dev server proxy in 3.19?

pages-e2e/features/pagesGetRequestContext/assets/wrangler.toml
petebacondarwin: I was just wondering if this was on purpose... like you were not sure what to put here.

packages/next-on-pages/src/api/getRequestContext.ts
```diff
@@ -0,0 +1,40 @@
...
+				| undefined;
+		}
+	)[cloudflareRequestContextSymbol];
+
+	if (!cloudflareRequestContext) {
+		throw new Error('Error: failed to retrieve the Cloudflare request context');
+	}
+
+	return {
+		...cloudflareRequestContext,
```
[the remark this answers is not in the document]
petebacondarwin: It's fine. Just interested in why.

petebacondarwin (approved):

pages-e2e/package.json
dario-piotrowicz: ok thanks, I'll have a look 😄

pages-e2e/features/pagesGetRequestContext/assets/wrangler.toml
dario-piotrowicz: ah ok, well let's put `x`es here just to make it clearer and consistent with [what we do in c3/wrangler](https://github.com/cloudflare/workers-sdk/blob/7412ce7e6674c37b4263319bb255064c5702321f/packages/create-cloudflare/templates/hello-world/js/wrangler.toml#L15) 🙂

packages/next-on-pages/src/api/getRequestContext.ts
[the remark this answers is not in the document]
petebacondarwin: Hmm interesting.. `getPlatformProxy()` sounds reasonable.

paramaggarwal:
How to use this inside App Router `page.tsx` file?

dario-piotrowicz:
@paramaggarwal you can import the utility from `@cloudflare/next-on-pages` and call it from your component

If should work fine as long as the component is using the edge runtime

You can see an example of that here: [get-request-context/page.jsx](https://github.com/dario-piotrowicz/next-on-pages/blob/3846730c4a0d12cda128d691160968d0a131a6f1/pages-e2e/features/appGetRequestContext/assets/app/get-request-context/page.jsx)

paramaggarwal:
Thanks for the example @dario-piotrowicz. I tried the same and am still not able to get it to work. Here are some learnings from my investigation:

1. The setup with `wrangler types`, updating `env.d.ts` to use it with `interface CloudflareEnv extends Env {}` is done
2. When I create a new app with `create-cloudflare` the `next.config.mjs` gets `await setupDevPlatform()` and not just `setupDevPlatform()` [as seen here](https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/templates/next/c3.ts#L51).
3. This was causing my `npx next dev` command to hang and the Next.js running console log was not showing up. Then I removed the `await` as per the other places in the docs and this let my Next app run, but my `getRequestContext` failed with error `failed to retrieve context` [from code here](https://github.com/dario-piotrowicz/next-on-pages/blob/main/packages/next-on-pages/src/api/getRequestContext.ts#L46)
4. Then I added logs locally to the flow and can see that this API call is not returning and an error is also not being thrown for me in [this line of code](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/dev-registry.ts#L173)

I've still not managed to fix my issue, but I think my local wrangler setup is not in sync with the remote cloudflare worker/pages instance because I had renamed the remote one and I think that broke something. So this is an issue on my side.

But this highlights two fixes:

1. Recommend developers to keep the `await` in `await setupDevPlatform()` so that if there are any errors they can be thrown on app startup itself.
2. The `await fetch(${DEV_REGISTRY_HOST}/workers)` seems to be hanging indefinitely for me - not sure what is the issue there.

dario-piotrowicz:
Hi @paramaggarwal thanks for the message 🙂 

Let me reply bit by bit...

> EDIT: I was able to resolve it by forcing the local worker registry to restart. ([reference](https://github.com/cloudflare/workers-sdk/issues/1920#issuecomment-1949886227))

I see, glad to hear that things are better 😄 , is it all good now then? or do you need further assitance? 🙂 

> 2. When I create a new app with `create-cloudflare` the `next.config.mjs` gets `await setupDevPlatform()` and not just `setupDevPlatform()` [as seen here](https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/templates/next/c3.ts#L51).

The `await` that create cloudflare adds is correct, I can see that I forgot to add that in our [`next-dev` readme](https://github.com/cloudflare/next-on-pages/blob/3846730c4a0d12cda128d691160968d0a131a6f1/internal-packages/next-dev/README.md?plain=1#L57), that's my bad, I'll look into adding it (although things should also work without the `await`)

> 3. This was causing my `npx next dev` command to hang and the Next.js running console log was not showing up. Then I removed the `await` as per the other places in the docs and this let my Next app run, but my `getRequestContext` failed with error `failed to retrieve context` [from code here](https://github.com/dario-piotrowicz/next-on-pages/blob/main/packages/next-on-pages/src/api/getRequestContext.ts#L46)

This makes it sound to me like indeed the `setupDevPlatform` was facing some issues, and by remove the `await` you effectively simply made the issue being ignored still without the dev platform getting set up (hence why `getRequestContext` would fail).

> I've still not managed to fix my issue, but I think my local wrangler setup is not in sync with the remote cloudflare worker/pages instance because I had renamed the remote one and I think that broke something. So this is an issue on my side.

`next-dev` as well as `wrangler pages dev` don't have anything to do with the remote Cloudflare worker/pages instance, so having it out of sync with your local setup should not really be an issue at all

> 1. Recommend developers to keep the `await` in `await setupDevPlatform()` so that if there are any errors they can be thrown on app startup itself.

👍 
> 2. The `await fetch(${DEV_REGISTRY_HOST}/workers)` seems to be hanging indefinitely for me - not sure what is the 
issue there.

Seems like a `wrangler` issue, anyways one that you've solved right? 🙂

paramaggarwal:
All good now, thanks a lot for the proactive help. 👍

dario-piotrowicz:
> One small idea. I can foresee a situation where someone did not add the setupDevPlatform() line in nextjs config but is then calling the getRequestContext() method. So we could update the failure log to suggest adding setupDevPlatform() to the nextjs config.

That's a good idea, I can add that, thanks 🙂👍

PR #661, opened 2024-02-13
address `getRequestContext` feedback

dario-piotrowicz:
Applying some changes to `getRequestContext` based on feedback received here: https://github.com/cloudflare/next-on-pages/discussions/596#discussioncomment-8443400
___

> [!NOTE]
> no changeset is needed since `getRequestContext` hasn't yet been released

### DIFF
--- a/packages/next-on-pages/src/api/getRequestContext.ts
+++ b/packages/next-on-pages/src/api/getRequestContext.ts
@@ -36,3 +36,15 @@ export function getRequestContext<
 
 	return cloudflareRequestContext;
 }
+
+export function getOptionalRequestContext<
+	CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+	Context = ExecutionContext,
+>(): RequestContext<CfProperties, Context> | null {
+	try {
+		const cloudflareRequestContext = getRequestContext<CfProperties, Context>();
+		return cloudflareRequestContext;
+	} catch {
+		return null;
+	}
+}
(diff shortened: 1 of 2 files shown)


packages/next-on-pages/src/api/getRequestContext.ts
```diff
@@ -34,7 +34,17 @@ export function getRequestContext<
...
+	return cloudflareRequestContext;
+}
+
+export function getOptionalRequestContext<
+	CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+	Context = ExecutionContext,
+>(): RequestContext<CfProperties, Context> | null {
+	try {
+		const cloudflareRequestContext = getRequestContext<CfProperties, Context>();
+		return cloudflareRequestContext;
```
petebacondarwin, line 46:
```suggestion
		return getRequestContext<CfProperties, Context>();
```

petebacondarwin (approved):

petebacondarwin (commented):
Actually if you invert the relationship between these two functions then it can be much simpler: the call to `getOptionalRequestContext()` would simply map `undefined` to `null` (or just leave as `undefined`) and then `getRequestContext()` would call `getOptionalRequestContext()` and throw if the value is null/undefined.

dario-piotrowicz:
> Actually if you invert the relationship between these two functions then it can be much simpler: the call to `getOptionalRequestContext()` would simply map `undefined` to `null` (or just leave as `undefined`) and then `getRequestContext()` would call `getOptionalRequestContext()` and throw if the value is null/undefined.

ah yeah... good point 🙂👍

packages/next-on-pages/src/api/getRequestContext.ts
dario-piotrowicz: I've anyways applied the other suggestion so this doesn't apply anymore 🙂👍

petebacondarwin (approved):

PR #662, opened 2024-02-13
Update README.md to add 'declare global' to env.d.ts

OpenSource03:
Add declare global to internal-packages/next-dev/README.md so people can better understand what they need to do. Just declaring interface is not sufficient and got me confused as I'm not used to working with env.d.ts files, and I'm sure there are other people who would find themselves confused.

### DIFF
--- a/internal-packages/next-dev/README.md
+++ b/internal-packages/next-dev/README.md
@@ -62,12 +62,13 @@ Next (optional but highly recommended) we create a [TypeScript declaration file]
 
 ```ts
 // file: env.d.ts
-
-interface CloudflareEnv {
-	MY_KV_1: KVNamespace;
-	MY_KV_2: KVNamespace;
-	MY_R2: R2Bucket;
-	MY_DO: DurableObjectNamespace;
+declare global {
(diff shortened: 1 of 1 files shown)


dario-piotrowicz:
Thank you very much for the PR @OpenSource03 🙂 

Declaring the interface in a declaration file in the __global scope__ (as indicated in the readme) should work, you can see it working by creating a new Next.js application using this C3 prerelease (https://github.com/cloudflare/workers-sdk/pull/5009):
```sh
npx https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/7893611368/npm-package-create-cloudflare-5009 --no-auto-update --framework next
```

I just tested there and it all looks good to me.

The `declare global` becomes necessary if your declaration file is in a __module scope__ (e.g. if the file has any `import`s/`export`s) then in that case you would need to use `declare global`

So I think that the current readme is correct... but I guess maybe we could clarify better the above? 🤔

dario-piotrowicz:
Closing this PR due to the inactivity + my earlier comment 👆

PR #668, opened 2024-02-17
add missing `await` to `setupDevPlatform` call in next-dev readme

dario-piotrowicz:


internal-packages/next-dev/README.md
```diff
@@ -54,7 +54,8 @@ export default nextConfig;
 // (note: this check is recommended but completely optional)
 if (process.env.NODE_ENV === 'development') {
 	// we simply need to call the utility
-	setupDevPlatform();
+	// (`await`ing the call is not mandatory but recommended)
```
dario-piotrowicz, line 57: Note: this comment might look strange, I wanted to add it because top level awaits are not allowed in cjs files, so if someone were to have a `next.config.(js/cjs)` then they could not `await` the function call, so hopefully this can avoid some confusion (without being too verbose/descriptive)

please let me know what you think 🙂

paramaggarwal: We should add why we recommend it to make this fully complete. It's ok to be verbose on the comments as it's supposed to be documentation anyway. 

Something like:

> The await only helps verify that the setup has succeeded, and is not mandatory. If using a .js/cjs config format that you can safely remove this top level await.

dario-piotrowicz: cool, thanks 🙂 

I've updated the comment please let me know what you think

petebacondarwin (approved):

internal-packages/next-dev/README.md
```diff
@@ -54,7 +54,11 @@ export default nextConfig;
 // (note: this check is recommended but completely optional)
 if (process.env.NODE_ENV === 'development') {
 	// we simply need to call the utility
-	setupDevPlatform();
+	// (`await`ing the call helps making sure that the setup has succeeded,
+	//  it is not mandatory but recommended. In config files using the
+	//  cjs format, where top level awaits are not supported you can
+	//  safely simply call the function and omit the `await`)
+	await setupDevPlatform();
```
petebacondarwin, line 61: In files that don't support top-level await, an alternative to avoid the unhandled rejection could be:

```
setupDevPlatform().catch(e => console.error(e));
```

dario-piotrowicz: yes, thanks for the input 🙂

do you think that it would be valuable to add in the comment? (or even in the code?)

I'm just afraid to get too verbose here....

what do you think?

petebacondarwin:
```ts
	// `await`ing the call helps making sure that the setup has succeeded,
	//  If you cannot use top level awaits you could use the following to avoid an unhandled rejection.
	//  `setupDevPlatform().catch(e => console.error(e));`
	await setupDevPlatform();
```

PR #682, opened 2024-02-26
Improve the `getRequestContext`/`getOptionalRequestContext`'s behavior in the Node.js runtime

dario-piotrowicz:
Currently if users run `getRequestContext` or `getOptionalRequestContext` in the Node.js runtime they get a generic error saying that the request context could not be found, these changes improve such situation by making it much clearer to the user that the error lies in the runtime being used.

These changes make it so that when either function is used:

- if we can detect for a high level of certainty that the code is running in the Node.js runtime throw an error clearly stating that the wrong runtime is being used
- if we are not completely sure what runtime is being used (and the request context could not be retrieved) add a hint in the error message prompting the developer to double check that they are using the correct runtime (only applies to `getRequestContext`)
- if the (correct) edge runtime is being used we don't include this information at all.

### DIFF
--- a/.changeset/brave-jeans-wait.md
+++ b/.changeset/brave-jeans-wait.md
@@ -1,0 +1,19 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Improve the `getRequestContext`/`getOptionalRequestContext`'s behavior in the Node.js runtime
+
+Currently if users run `getRequestContext` or `getOptionalRequestContext` in the Node.js runtime
+they get a generic error saying that the request context could not be found, these changes
+improve such situation by making it much clearer to the user that the error lies in the runtime
+being used.
+
+These changes make it so that when either function is used:
+
+- if we can detect for a high level of certainty that the code is running in the Node.js runtime
+  throw an error clearly stating that the wrong runtime is being used
+- if we are not completely sure what runtime is being used (and the request context could not be
+  retrieved) add a hint in the error message prompting the developer to double check that they
+  are using the correct runtime (only applies to `getRequestContext`)
+- if the (correct) edge runtime is being used we don't include this information at all.
--- a/packages/next-on-pages/src/api/getRequestContext.ts
+++ b/packages/next-on-pages/src/api/getRequestContext.ts
@@ -19,6 +19,12 @@ const cloudflareRequestContextSymbol = Symbol.for(
 	'__cloudflare-request-context__',
 );
 
+const notEdgeRuntimeErrorMessage = dedent`
+\`getRequestContext\` and \`getOptionalRequestContext\` can only be run inside the edge runtime,
+so please make sure to have included \`export const runtime = 'edge'\`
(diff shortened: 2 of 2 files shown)


packages/next-on-pages/src/api/getRequestContext.ts
```diff
@@ -66,3 +86,36 @@ export function getRequestContext<
...
 	return cloudflareRequestContext;
 }
+
+/**
+ * detects whether the current code is running inside the (local) edge runtime or not
+ *
+ * @returns a boolean indicating whether the code is running inside the edge runtime if that can be accurately detected,
+ * 			'maybe' if the code is likely being run inside the edge runtime but we're not 100% sure of that
+ */
+function isInsideEdgeRuntime(): boolean | 'maybe' {
```
petebacondarwin, line 96:
```suggestion
function inferRuntime(): 'edge' | 'node' | 'unsure' {
```

packages/next-on-pages/src/api/getRequestContext.ts
```diff
@@ -66,3 +86,36 @@ export function getRequestContext<
...
+			);
+			if (!notSupportedInEdgeError) {
+				return true;
+			}
+		}
+		// an error was thrown... likely we are in the edge runtime, but there's no real guarantee
+		return 'maybe';
+	}
+
+	// for the time being we can safely assume that if `process.cwd()` doesn't error
```
petebacondarwin, line 118: Or that they have included some polyfill?
Can we check something else like `process.release.name`?

https://nodejs.org/docs/latest/api/process.html#processrelease

petebacondarwin:
Could we create some tests for this?

dario-piotrowicz:
> Could we create some tests for this?

Currently we don't have any e2e test for local development behaviors (maybe we should 😕, the problem is that it would require an e2e testing process completely different from what we have now) so that's why I haven't added tests for this

But I guess we could have unit tests for `getRequestContext` 👍 

Good point! thanks I'll look into adding them 🙂

packages/next-on-pages/src/api/getRequestContext.ts
dario-piotrowicz: `process.release.name` sounds like a great idea! 😃 

I didn't know about it (nor noticed it when I was scanning `process` to see what we could use from that)

I'll have a look thanks! 😄

dario-piotrowicz: works like a charm! 🤩👍

petebacondarwin (approved):

.changeset/brave-jeans-wait.md
```diff
@@ -0,0 +1,10 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Make `getRequestContext` and `getOptionalRequestContext` throw is used inside the Node.js runtime
```
dario-piotrowicz, line 5:
```suggestion
Make `getRequestContext` and `getOptionalRequestContext` throw if used inside the Node.js runtime
```

PR #688, opened 2024-03-02
fix: originalRequest bug appearing again

james-elicx:
A user in Discord was running into this Next.js bug again. Turns out their Vercel functions were generated in an unminified form when using Next.js 14.1.1, so extending our regex for the unminified version.

image


dario-piotrowicz (approved):
The change looks good to me 🙂

(well as good as a string replacement hack can be I guess 😅 🙈)

Before merging, would it be possible to see the app that caused the issue so that I could manually check this? also I am wondering if the code not being minified itself could be an issue 😕 🤔

james-elicx:
> The change looks good to me 🙂
> 
> (well as good as a string replacement hack can be I guess 😅 🙈)
> 
> Before merging, would it be possible to see the app that caused the issue so that I could manually check this? also I am wondering if the code not being minified itself could be an issue 😕 🤔

Sure, it was https://github.com/Headpat-Community/web. If you change it to next@14.1.1, the Vercel functions won't be minified. I have a feeling Next.js have completely broken their bundling, because 14.1.1 also results in significantly larger _worker.js bundle sizes.

dario-piotrowicz:
@james-elicx thanks! ❤️ 

yeah nothing seems out of the ordinary there (like in the config file) 😓 :+1:

(and yeah it seems like your fix is what's needed to get the app to work)

Ok, let's continue with this and hope that we can remove it soon 😓

PR #690, opened 2024-03-05
fix `process.env` not giving access to bindings during local development (in `next-dev`)

dario-piotrowicz:
This PR fixes the fact that `process.env` doesn't contain the proper bindings during dev mode

You can see a reproduction of the issue here: https://github.com/dario-piotrowicz/next-on-pages-local-process.env-issue-repro

This is a regression introduced in https://github.com/cloudflare/next-on-pages/pull/659

There I removed the `bindingsProxyHasBeenSetSymbol` flag used to see if `process.env` was properly patched and just relied on the existence of `runtimeContext[cloudflareRequestContextSymbol]` instead.

Unfortunately this doesn't work since Next.js itself creates/updates the `process.env` object thus we cannot/should not use external flags to see if it was patched but the flag has to be part of the `process.env` object itself. So that if Next.js has replaced the object we can see that and re-patch it.

### DIFF
--- a/internal-packages/next-dev/src/shared.ts
+++ b/internal-packages/next-dev/src/shared.ts
@@ -35,7 +35,7 @@ export function monkeyPatchVmModule({ env, cf, ctx, caches }: PlatformProxy) {
 
 	const originalRunInContext = vmModule.runInContext.bind(vmModule);
 
-	vmModule.runInContext = (...args: [string, RuntimeContext, ...[unknown]]) => {
+	vmModule.runInContext = (...args: [string, RuntimeContext, ...unknown[]]) => {
 		const runtimeContext = args[1];
 
 		runtimeContext[cloudflareRequestContextSymbol] ??= {
--- a/internal-packages/next-dev/src/shared.ts
+++ b/internal-packages/next-dev/src/shared.ts
@@ -59,7 +59,7 @@ function monkeyPatchProcessEnv(
 		runtimeContext.process?.env &&
 		!runtimeContext.process.env[processEnvIsPatched]
 	) {
-		if (runtimeContext.process?.env) {
+		if (runtimeContext.process.env) {
 			for (const [name, binding] of Object.entries(env)) {
 				runtimeContext.process.env[name] = binding;
 			}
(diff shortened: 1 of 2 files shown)


internal-packages/next-dev/src/shared.ts
```diff
@@ -21,56 +35,64 @@ export function monkeyPatchVmModule({ env, cf, ctx, caches }: PlatformProxy) {
...
-				[cloudflareRequestContextSymbol]?: {
-					env: unknown;
-					ctx: unknown;
-					cf: unknown;
-				};
-			},
-			...[unknown],
-		]
-	) => {
+	vmModule.runInContext = (...args: [string, RuntimeContext, ...[unknown]]) => {
```
petebacondarwin, line 38: Isn't `...[unknown]` the same thing as `unknown`?
So you could just do

```suggestion
	vmModule.runInContext = (...args: [string, RuntimeContext, unknown]) => {
```

petebacondarwin: https://www.typescriptlang.org/play?#code/C4TwDgpgBAZgdgRigXigCgHRYIYCcDmAzgFxQDahwuAlnPgDRRwCuAtgEYS6PNwDWcAPYB3OAF0AlCgB8UAG6DqAEwDcAKFCRYcAEwp0WDHiKkKVWgyZtO3KIbK8BI8ZJnzFqtWoA2EYLARSeCRUNClkWQBvAF91X38YHSDdfRgEdSA

petebacondarwin: Or did you really mean:

```suggestion
	vmModule.runInContext = (...args: [string, RuntimeContext, ...unknown[]]) => {
```

internal-packages/next-dev/src/shared.ts
```diff
@@ -21,56 +35,64 @@ export function monkeyPatchVmModule({ env, cf, ctx, caches }: PlatformProxy) {
...
 
+function monkeyPatchProcessEnv(
+	runtimeContext: RuntimeContext,
+	env: Record<string, unknown>,
+) {
+	if (
+		runtimeContext.process?.env &&
+		!runtimeContext.process.env[processEnvIsPatched]
+	) {
+		if (runtimeContext.process?.env) {
```
petebacondarwin, line 62: This must always be true in here.

petebacondarwin (approved):
A couple of nits.

internal-packages/next-dev/src/shared.ts
dario-piotrowicz: yeah you're totally right, thanks for noticing it 😄 🙏

PR #716, opened 2024-03-20
increase the number of e2e attempts as well as the timeouts

dario-piotrowicz:


dario-piotrowicz:
no longer necessary/applicable

PR #730, opened 2024-04-05
normalize multi segments paths so that paths are split over multiple query params

dario-piotrowicz:


dario-piotrowicz:
Unlikely to be the proper solution 😓

PR #734, opened 2024-04-07
mark `cloudflare:*` as external for esbuild

james-elicx:


n3oney:
This fixes my issue with not being able to use postgres.js!

packages/next-on-pages/src/buildApplication/processVercelFunctions/build.ts
```diff
@@ -83,24 +83,33 @@ type RelativePathOpts = {
...
 	setup(build) {
-		build.onResolve({ filter: /^node:/ }, ({ kind, path }) => {
-			// this plugin converts `require("node:*")` calls, those are the only ones that
-			// need updating (esm imports to "node:*" are totally valid), so here we tag with the
-			// node-buffer namespace only imports that are require calls
+		build.onResolve({ filter: /^(node|cloudflare):/ }, ({ kind, path }) => {
+			/**
+			 * This plugin converts `require("prefix:*")` calls, those are the only ones that need
+			 * updating (esm imports to "prefix:*" are totally valid), so here we tag with the
+			 * built-in-modules namespace only imports that are require calls.
```
dario-piotrowicz, line 100: Nit: instead of `prefix` I'd try to use something that makes it more apparent that it's a variable value

```suggestion
			 * This plugin converts `require("<PREFIX>:*")` calls, those are the only ones that need
			 * updating (esm imports to "<PREFIX>:*" are totally valid), so here we tag with the
			 * built-in-modules namespace only imports that are require calls.
```

packages/next-on-pages/src/buildApplication/processVercelFunctions/build.ts
```diff
@@ -83,24 +83,33 @@ type RelativePathOpts = {
...
+				? { path, namespace: 'built-in-modules' }
 				: undefined;
 		});
 
-		// we convert the imports we tagged with the node-built-in-modules namespace so that instead of `require("node:*")`
-		// they import from `export * from "node:*";`
+		/**
+		 * We convert the imports we tagged with the built-in-modules namespace so that instead of
+		 * `require("prefix:*")` they import from `export * from "prefix:*";`
+		 */
```
dario-piotrowicz, line 110: Nit: instead of `prefix` I'd try to use something that makes it more apparent that it's a variable value

```suggestion
		/**
		 * We convert the imports we tagged with the built-in-modules namespace so that instead of
		 * `require("<PREFIX>:*")` they import from `export * from "<PREFIX>:*";`
		 */
```

dario-piotrowicz (approved):
LGTM, thanks @james-elicx 😄

Issue #746, opened 2024-04-18
[🐛 Bug]: "The following routes were not configured to run with the Edge Runtime for static pages" after upgrading to Next 14.2.2


Adriel-M:
### next-on-pages environment related information

```
bunx @cloudflare/next-on-pages --info                                                                                             ✘ 1 main ✱

System:
	Platform: darwin
	Arch: arm64
	Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:19:22 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T8112
	CPU: (8) arm64 Apple M2
	Memory: 16 GB
	Shell: /bin/zsh
Package Manager Used: bun (1.1.3)

Relevant Packages:
	@cloudflare/next-on-pages: 1.11.0
	vercel: N/A
	next: N/A
```

### Description

After upgrading to Next.js 14.2.2, I can't deploy anymore and have been getting these error messages after running `bunx @cloudflare/next-on-pages` both locally and when deploying.

```
⚡️ Invalid prerender config for /about.action
⚡️ Invalid prerender config for /index.action
⚡️ Invalid prerender config for /posts/[slug].action
⚡️ Invalid prerender config for /posts/page/[page].action
⚡️ Invalid prerender config for /posts.action
⚡️ Invalid prerender config for /projects.action
⚡️ Invalid prerender config for /tags/[tag]/page/[page].action
⚡️ Invalid prerender config for /tags/[tag].action

⚡️ ERROR: Failed to produce a Cloudflare Pages build from the project.
⚡️
⚡️ 	The following routes were not configured to run with the Edge Runtime:
⚡️ 	  - /about.action
⚡️ 	  - /index.action
⚡️ 	  - /posts/[slug].action
⚡️ 	  - /posts/page/[page].action
⚡️ 	  - /posts.action
⚡️ 	  - /projects.action
⚡️ 	  - /tags/[tag]/page/[page].action
⚡️ 	  - /tags/[tag].action
⚡️
```

These should be static pages:

```
▲  Route (app)                                                                   Size     First Load JS
▲  ┌ ○ /                                                                         193 B            94 kB
▲  ├ ○ /_not-found                                                               145 B          87.2 kB
▲  ├ ○ /about                                                                    1.31 kB         100 kB
▲  ├ ƒ /og                                                                       0 B                0 B
▲  ├ ○ /posts                                                                    193 B            94 kB
[...]

PR #749, opened 2024-04-20
Exclude action.func folders when processing prerenderFunctions

Adriel-M:
Fixes: #746 

Excludes this folder when processing `prerenderFunctions` since the `.vc-config.json` located in this folder contains `operationType == isr` and collected then validated later on. The validation looks for an associated `.prerender-config.json` in the root functions folder but doesn't have it for this new `.action.func` folder.

This folder was introduced in the 14.2.2 release specifically in this pr https://github.com/vercel/next.js/pull/64592

At the moment it doesn't have any functionality but might need to get revisited in the future.


dario-piotrowicz:
@Adriel-M thanks so very much for this PR! ❤️

It looks great and works as intended 🙂 

However the code is inconsistent with the current logic we have in which we first collect all the functions and then ignore the invalid ones we can, this also results us in logging such ignored functions in `.vercel/output/static/_worker.js/nop-build-log.json` (which might or not be useful in this case 😅).

Also this PR's logic is being tested (thanks a lot for adding the test btw! ❤️) under `processPrerenderFunctions` although the logic isn't part of that function 😕.

I think we should ideally amend the above points before merging the PR, I've drafted here what I think the proper tweaks should be: https://github.com/cloudflare/next-on-pages/commit/bc60685e09f85280133118c619b07869b130b205

Please have a look and let me know if you'd be willing to make these changes 🙏
(if not please let me know and I can push the necessary changes to your branch and merge)

.changeset/funny-icons-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Exclude .action.func folder introduced in next.js 14.2.2
```
dario-piotrowicz, line 5:
```suggestion
Ignore the `.action.func` folders introduced in next.js 14.2.2
```

Adriel-M:
@dario-piotrowicz cherry-picked your commit + populated the TODO comment. Let me know if it's good enough!

dario-piotrowicz (approved):

dario-piotrowicz:
@Adriel-M all good (sorry I just updated the comment for consistency sake), thanks a lot for the PR again! ❤️ 🙏

Issue #740, opened 2024-04-10
[🐛 Bug]: nextjs local preview succeeded, deployment succeeded, but no deployment was shown


hehehai:
### next-on-pages environment related information

System:
        Platform: darwin
        Arch: x64
        Version: Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:27 PDT 2023; root:xnu-10002.41.9~6/RELEASE_X86_64
        CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
        Memory: 32 GB
        Shell: /bin/zsh
Package Manager Used: npm (10.1.0)

Relevant Packages:
        @cloudflare/next-on-pages: 1.11.0
        vercel: N/A
        next: N/A

### Description

developed the nextjs project, using pages deployment, the first deployment was successful, but after my subsequent modifications, the deployment was successful, but accessing pages failed

<img width="1185" alt="image" src="https://github.com/cloudflare/next-on-pages/assets/12692552/720a5326-575d-4480-a3ce-09b1405645bf">

### Reproduction

https://github.com/hehehai/at4-kit

```toml
name = "at4-start"
compatibility_date = "2024-04-05"
compatibility_flags = ["nodejs_compat"]

[vars]
NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY = "pk_test_xx"
CLERK_SECRET_KEY = "sk_test_xx"

[[d1_databases]]
binding = "DB" # i.e. available in your Worker on env.DB
database_name = "prisma-demo-db"
database_id = "xx"
```

```json
"pages:build": "npx @cloudflare/next-on-pages",
"preview": "npm run pages:build && wrangler pages dev .vercel/output/static",
"deploy": "npm run pages:build && wrangler pages deploy .vercel/output/static",
"build-cf-types": "wrangler types --env-interface CloudflareEnv env.d.ts"
```

### Pages Deployment Method

None

### Pages Deployment ID

at4-start

### Would you like to help?

- [ ] Would you like to help fixing this bug?

hehehai:
<img width="1334" alt="image" src="https://github.com/cloudflare/next-on-pages/assets/12692552/63d9ffba-0963-48cd-afc9-ef2e5e5761eb">
<img width="577" alt="image" src="https://github.com/cloudflare/next-on-pages/assets/12692552/dea16b39-ad40-4702-bbb4-16614ed6003a">

liuhenry:
[...]

PR #754, opened 2024-04-22
fix: runBuild should exit with 1 on unexpected errors (when watch mode is disabled)

juanferreras:
~(Maybe) partially addresses: #740.~ (EDIT: OP seems to have a different issue, the build does go through without errors even though the worker is somehow malformed).

I've encountered a similar issue on my own, we had an import of a file with multiple locale-specific characters, and it was causing the `.vercel/output/functions/<page-name>.func/index.js` to contain a blurb like
```
Ş:"Scedil",ß:"szlig",𝔱:"tfr",𝓉:"tscr",𝕥:"topf",𝒯:"Tscr",𝔗:"Tfr",𝕋:"Topf",ť:"tcaron",Ť:"Tcaron",ţ:"tcedil",Ţ:"Tcedil","™":"trade",ŧ:"tstrok",Ŧ:"Tstrok",𝓊:"uscr",𝕦:"uopf",𝔲:"ufr",𝕌:"Uopf",𝔘:"Ufr",𝒰:"Uscr",ú:"uacute",Ú:"Uacute",ù:"ugrave",Ù:"Ugrave",ŭ:"ubreve",Ŭ:"Ubreve",û:"ucirc",Û:"Ucirc",ů:"uring",Ů:"Uring",ü:"uuml",Ü:"Uuml",ű:"udblac",Ű:"Udblac",ũ:"utilde",Ũ:"Utilde",ų:"uogon",Ų:"Uogon",ū:"umacr",Ū:"Umacr",𝔳:"vfr",𝕧:"vopf",𝓋:"vscr",𝔙:"Vfr",𝕍:"Vopf",𝒱:"Vscr",𝕨:"wopf",𝓌:"wscr"
```
which caused the build logs to output
```
▲  Collected static files (public/, static/, .next/static): 59.326ms
▲  Build Completed in .vercel/output [4m]
⚡️ Completed `pnpm dlx vercel build`.
⚡️ Unexpected error: Unexpected character '�' (8036:87743)
```

I wasn't able to come up with a simple repro on this specific issue (and it looks more like something our project was doing wrong), but it's fairly simple to test that if `acorn's parse()` throws inside `getFunctionIdentifiers()`, the deploy fails and the exit code is 0 (which means CI or a command that does `&& pnpm run:deploy` would carry on with a broken build).
CleanShot 2024-04-22 at 09 16 05@2x

In our case, it was creating the same defect as OP has on #740, a deployment successfully goes out but all paths return 404 (even homepage) as something goes wrong building the worker's code.

**I'm assuming there might be N different reasons why a build can fail at this stage**, but if they're all going through the unexpected error path, I'd assume we'll always want to signal exit 1 to prevent deployments to continue broken.

### DIFF
--- a/.changeset/loud-owls-remember.md
+++ b/.changeset/loud-owls-remember.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+runBuild should exit with 1 on unexpected errors (when watch mode is disabled)
--- a/.changeset/loud-owls-remember.md
+++ b/.changeset/loud-owls-remember.md
@@ -2,4 +2,4 @@
 '@cloudflare/next-on-pages': patch
 ---
 
-runBuild should exit with 1 on unexpected errors (when watch mode is disabled)
+exit with exit status 1 on build unexpected errors (when watch mode is disabled)
--- a/packages/next-on-pages/src/index.ts
+++ b/packages/next-on-pages/src/index.ts
@@ -65,6 +65,9 @@ function runBuild(options: CliOptions) {
 		const errorMessage =
 			error instanceof Error ? error.message : JSON.stringify(error);
 		cliError(`Unexpected error: ${errorMessage}`);
+		if (!options.watch) {
+			process.exit(1);
+		}
 	});
 }
 


.changeset/loud-owls-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+runBuild should exit with 1 on unexpected errors (when watch mode is disabled)
```
dario-piotrowicz, line 5:
```suggestion
exit with exit status 1 on build unexpected errors (when watch mode is disabled)
```

PR #756, opened 2024-04-23
Cache loading and update optimization

ibobo:
Parallelizes operations while loading the cache tags manifest, and locally caches them for up to a second to avoid refresh on each cache request. On cache update, uses `ctx.waitUntil` instead of blocking the response while update is in progress.

I don't have real-life test numbers, but on our website where some pages require a 3 API request waterfall to render, I easily was seeing 400ms "wasted" by writing data to the cache. These changes solved my issue.

### DIFF
--- a/packages/next-on-pages/templates/cache/adaptor.ts
+++ b/packages/next-on-pages/templates/cache/adaptor.ts
@@ -13,6 +13,10 @@ export class CacheAdaptor {
 	public tagsManifest: TagsManifest | undefined;
 	/** The key used for the tags manifest in the cache. */
 	public tagsManifestKey = 'tags-manifest';
+	/** Promise that resolves when tags manifest is loaded */
+	public tagsManifestPromise: Promise<void> | undefined;
+	/** Records the last successful load time of the tags manifest */
+	public lastTagsManifestLoad: number | undefined;
 
 	/**
 	 * @param ctx The incremental cache context from Next.js. NOTE: This is not currently utilised in NOP.
@@ -52,7 +56,7 @@ export class CacheAdaptor {
 		};
 
 		// Update the cache entry.
-		await this.update(key, JSON.stringify(newEntry));
+		const updateOp = this.update(key, JSON.stringify(newEntry));
 
 		switch (newEntry.value?.kind) {
 			case 'FETCH': {
@@ -70,6 +74,8 @@ export class CacheAdaptor {
 				);
 			}
 		}
+
+		await updateOp;
 	}
 
 	/**
@@ -84,7 +90,12 @@ export class CacheAdaptor {
 		{ softTags }: { softTags?: string[] },
 	): Promise<CacheHandlerValue | null> {
 		// Get entry from the cache.
-		const entry = await this.retrieve(key);
+		const entryPromise = this.retrieve(key);
+
+		// And start loading the tags manifest.
+		const tagsManifestLoad = this.loadTagsManifest();
+
+		const entry = await entryPromise;
 		if (!entry) return null;
 
 		let data: CacheHandlerValue;
@@ -97,8 +108,8 @@ export class CacheAdaptor {
 
 		switch (data.value?.kind) {
 			case 'FETCH': {
-				// Load the tags manifest.
-				await this.loadTagsManifest();
+				// Await for the tags manifest to end loading.
+				await tagsManifestLoad;
 
 				// Check if the cache entry is stale or fresh based on the tags.
 				const tags = getTagsFromEntry(data);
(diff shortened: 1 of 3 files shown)


packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -70,6 +74,8 @@ export class CacheAdaptor {
 				);
 			}
 		}
+
+		await updateOp;
```
dario-piotrowicz, line 78:
```suggestion
        // Make sure the cache has been updated before returning
		await updateOp;
```

packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -84,7 +90,12 @@ export class CacheAdaptor {
 		{ softTags }: { softTags?: string[] },
 	): Promise<CacheHandlerValue | null> {
 		// Get entry from the cache.
-		const entry = await this.retrieve(key);
+		const entryPromise = this.retrieve(key);
+
+		// And start loading the tags manifest.
```
dario-piotrowicz, line 95: Nit.
```suggestion
		// Start loading the tags manifest.
```

packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -140,8 +151,35 @@ export class CacheAdaptor {
...
+	 *
+	 * @param force Whether to force a reload of the tags manifest.
+	 */
+	public async loadTagsManifest(force = false): Promise<void> {
+		// Refresh tags manifest after 1 second or if not loaded.
+		const shouldLoad =
+			force ||
+			!this.tagsManifest ||
+			!this.lastTagsManifestLoad ||
+			Date.now() - this.lastTagsManifestLoad > 1000;
```
dario-piotrowicz, line 163: Did you come up with the 1 second logic here? or is it something that Next.js/Vercel does?

dario-piotrowicz (approved):
Hi @ibobo 👋 

Thanks so very much for the PR, it looks awesome! ❤️‍🔥 

I'm happy to go with it but I've left a few very minor comments, please have a look 🙂 

In particular I'd be very keen to understand the reasoning behind the 1 second refresh tags manifest logic as we usually try to follow closely what Next.js/Vercel does without introducing our own logic unless necessary 🙂

dario-piotrowicz:
PS: @ibobo so sorry that it took a while to review your PR! 🙇 🙇 🙇 (usually I do it much faster, I've been occupied on other projects in the last few weeks)

packages/next-on-pages/templates/cache/adaptor.ts
ibobo: No, it isn't something I'm aware Next.js or Vercel does. I thought it could be a reasonable tradeoff, but I'm open for better ideas. The reason behind this is just to avoid re-loading the tags for each and every get, so, as an example, caching them just for the duration of the current request would be equally fine.

ibobo:
> PS: @ibobo so sorry that it took a while to review your PR! 🙇 🙇 🙇 (usually I do it much faster, I've been occupied on other projects in the last few weeks)

Not a problem!

packages/next-on-pages/templates/cache/adaptor.ts
ibobo: Yeah, that's ok for me! 😄 

I think I'll have time to make this change this week. Thanks! 💯

dario-piotrowicz: Awesome! thanks so very much! ❤️

ibobo:
Hi @dario-piotrowicz, I had the time to check on the 1 second logic. I saw that a new Adaptor is instantiated for each cache load, I thought the adapter lasted for a whole request... so the 1 second logic is also useless, and I removed it as requested.

So, this PR becomes part 1 of 2 😄 it's valuable on its own, since it parallelizes some loads and uses `waitUntil` for the saving; in part 2 I want to make cache adaptor shared for a whole request. I will open an issue to discuss the matter.

Thanks!

dario-piotrowicz:
> I saw that a new Adaptor is instantiated for each cache load, I thought the adapter lasted for a whole request... 

Dang! you're right! yeah sorry that slipped my mind! 😓 

> So, this PR becomes part 1 of 2 😄 it's valuable on its own, since it parallelizes some loads and uses `waitUntil` for the saving; 

Yes definitely 😄 💪 

> in part 2 I want to make cache adaptor shared for a whole request. I will open an issue to discuss the matter.

yes, recreating the adaptor each time doesn't seem correct to me... thanks a lot! 😄

cc. @james-elicx in case you're interested since you implemented the cache adaptor 🙂 (and I'd be curious to hear if the current implementation in which the adapter is created each time could have been intentional or just an oversight)

james-elicx:
> > in part 2 I want to make cache adaptor shared for a whole request. I will open an issue to discuss the matter.
> 
> yes, recreating the adaptor each time doesn't seem correct to me... thanks a lot! 😄
> 
> cc. @james-elicx in case you're interested since you implemented the cache adaptor 🙂 (and I'd be curious to hear if the current implementation in which the adapter is created each time could have been intentional or just an oversight)

Since it's done in the patched fetch which exists outside the ALS context, I'm not sure how you would make it shared for the duration of a request without it existing across multiple requests as part of the global fetch patch.

I mean, I don't think there's necessarily anything wrong with it existing across multiple requests, but for some reason I have a memory of trying to avoid having things that lived across multiple requests.

dario-piotrowicz:
> Since it's done in the patched fetch which exists outside the ALS context, I'm not sure how you would make it shared for the duration of a request without it existing across multiple requests as part of the global fetch patch.
>
> I mean, I don't think there's necessarily anything wrong with it existing across multiple requests, but for some reason I have a memory of trying to avoid having things that lived across multiple requests.

Ah no yeah that is a good point! 🤔 

I'd really avoid having it shared across multiple requests... 

But to isolate it we could also simply implement a separate ALS here for this, couldn't we?

PR #765, opened 2024-05-02
add commit shas to cache names and keys

dario-piotrowicz:


packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -203,7 +203,13 @@ export class CacheAdaptor {
...
 	 */
 	public buildCacheKey(key: string) {
-		return `https://${SUSPENSE_CACHE_URL}/entry/${key}`;
+		// Note: we use the pages commit sha if present so that caches from
+		//       a previous deployment are not reused for new ones
+		return `https://${SUSPENSE_CACHE_URL}${
+			process.env.CF_PAGES_COMMIT_SHA
+				? `_${process.env.CF_PAGES_COMMIT_SHA}`
+				: ''
+		}/entry/${key}`;
```
WalshyDev, line 212: maybe better put in path? So it'd be `https://internal_suspense_cache_hostname.local/95c51fb350642275383d075267f379df2b31046d/entry/chunk-abc.js';

dario-piotrowicz: sure, I'll do that 🙂👍

(I don't think it makes any difference... but I also don't really have a preference here 🙂)

dario-piotrowicz: updated 🙂👍

packages/next-on-pages/templates/cache/adaptor.ts
```diff
@@ -203,7 +203,13 @@ export class CacheAdaptor {
 	 * @returns The fully-formed cache key for the suspense cache.
 	 */
 	public buildCacheKey(key: string) {
-		return `https://${SUSPENSE_CACHE_URL}/entry/${key}`;
+		// Note: we use the pages commit sha if present so that caches from
+		//       a previous deployment are not reused for new ones
```
james-elicx, line 207: Next.js docs state that the cache should be shared across deployments.
 
image

dario-piotrowicz: oh.... thanks I totally missed that 😓

dario-piotrowicz:
Closing as the current behaviour seems correct at per @james-elicx comment

PR #770, opened 2024-05-04
fix fetch patch hanging promise by moving symbol to globalThis

### LINKED ISSUE #769: [🐛 Bug]: Patching fetch clobbers Next.js patching fetch leading to recursive wrapping

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000
        CPU: (10) arm64 Apple M1 Pro
        Memory: 32 GB
        Shell: /bin/zsh
Package Manager Used: npm (10.5.2)

Relevant Packages:
        @cloudflare/next-on-pages: 1.11.2
        vercel: N/A
        next: 14.2.3

### Description

Basically, both Next.js (possibly just from 14.2 onwards) and next-on-pages, attempt to patch fetch – and both check for properties on the `globalThis.fetch` function itself... which is the very thing that both are patching and replacing – so each think that it hasn't been patched and patch it again.

This happens every request, so the patched function just keeps getting wrapped and wrapped, deeper and deeper.

next-on-pages patches here:

https://github.com/cloudflare/next-on-pages/blob/f150546a8c4a7a997d1f6f0a60d9c303f8ddd768/packages/next-on-pages/templates/_worker.js/utils/fetch.ts#L8-L14

Next.js patches here:

https://github.com/vercel/next.js/blob/9ec37c120aba9d5f16e00566a10cba68858e363d/packages/next/src/server/lib/patch-fetch.ts#L779

This ended up showing up as many `Error: The script will never generate a response` errors along with:

```
A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.
```

(Didn't go down the full rabbit hole of what's getting in the way – just that `fetch` calls seem to stall – and either calling the original global fetch, or manually fixing this monkey-patch behaviour, both seemed to fix it)

### Reproduction

I just had a simple page that did this:

```ts
export const runtime = 'edge';

export default async function LinkPage() {
  const id = crypto.randomUUID();

  console.log(`Begin fetch: ${id}`);

  // Harder to reproduce when fetch is cached? So make url random to prevent caching
  await (await fetch(`https://www.cloudflare.com/?cb=${id}`)).arrayBuffer();

  console.log(`End fetch: ${id}`);

  return <div>OK</div>;
}
```

Calling the page directly would sometimes work (I guess the patching hadn't fully wrapped by then), but if I had other pages with the same code that I `<Link>`ed to, then clicking those links would cause the error (and only the `Begin fetch` line would show up – the `End fetch` never made it)

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

I'd fix this by setting a property anywhere but on the function you're monkey-patching. Eg, just a symbol on `globalThis` itself?

Also probably worth raising an issue with Next.js too – they also shouldn't be checking for patch existence on the function itself

### Would you like to help?

james-elicx:
fixes #769

See the issue for context.

### DIFF
--- a/.changeset/itchy-cheetahs-design.md
+++ b/.changeset/itchy-cheetahs-design.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix hanging promise caused by fetch patch symbol
--- a/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/fetch.ts
@@ -5,13 +5,13 @@ import { handleSuspenseCacheRequest } from './cache';
  * to work
  */
 export function patchFetch(): void {
-	const alreadyPatched = (globalThis.fetch as Fetch)[patchFlagSymbol];
+	const alreadyPatched = (globalThis as GlobalWithPatchSymbol)[patchFlagSymbol];
 
 	if (alreadyPatched) return;
 
 	applyPatch();
 
-	(globalThis.fetch as Fetch)[patchFlagSymbol] = true;
+	(globalThis as GlobalWithPatchSymbol)[patchFlagSymbol] = true;
 }
 
 function applyPatch() {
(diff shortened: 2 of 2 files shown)


dario-piotrowicz (approved):
LGTM, thanks a bunch for the fix @james-elicx 😄

mhart:
Yesss, LGTM! :shipit:

PR #771, opened 2024-05-04
Remove the `cache` property from calls to `new Request(...)` in React

### LINKED ISSUE #719: [🐛 Bug]: The 'cache' field on 'RequestInitializerDict' is not implemented.

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 23.4.0: Wed Feb 21 21:44:43 PST 2024; root:xnu-10063.101.15~2/RELEASE_ARM64_T6000
        CPU: (8) arm64 Apple M1 Pro
        Memory: 16 GB
        Shell: /bin/zsh

Relevant Packages:
        @cloudflare/next-on-pages: 1.10.0
        vercel: N/A
        next: N/A

### Description

I deployed Next.js 14 [E-commerce](https://github.com/vercel/commerce/) app on pages and everything is working fine most of the time but I occasionally get this error:
```
{
  cause: 'unknown',
  status: 500,
  message: "The 'cache' field on 'RequestInitializerDict' is not implemented.",
  query: 'something'
  }
```

### Reproduction

- Go to product page (Or any page that uses `cache: 'no-store'` property)
- Refresh the page for couple of time
- The error happens

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
In the React server code, there is a ternary where they create a new Request object and forward all properties on `RequestInit` to it. This won't work in workerd due to the `cache` property being present. Therefore, we need to overwrite their ternary with a way for us to strip the `cache` property from the `RequestInit` object.

https://github.com/vercel/next.js/blob/9ec37c12/packages/next/src/compiled/react/cjs/react.react-server.production.js#L87

Additionally, once this patch has been added, it looks like the error can then happen in our patched fetch. Therefore, also adding similar logic there to strip out the `cache` property from the `RequestInit` object.

fixes #719

### DIFF
--- a/.changeset/odd-buses-lay.md
+++ b/.changeset/odd-buses-lay.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Remove `cache` from React's `new Request(...)` cache.
--- a/.changeset/tame-islands-arrive.md
+++ b/.changeset/tame-islands-arrive.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Strip the `cache` property from our patched fetch.
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -503,13 +503,13 @@ function fixFunctionContents(contents: string): string {
 	// https://github.com/vercel/next.js/blob/9ec37c12/packages/next/src/compiled/react/cjs/react.react-server.production.js#L87
 	contents = contents.replace(
 		/([\w$]+) instanceof URL\?new Request\([\w$]+,([\w$]+)\):[\w$]+;/gm,
-		`(() => {
-			if ($1 instanceof URL) {
-				const { cache, ...init } = $2 ?? {};
-				return new Request($1, init);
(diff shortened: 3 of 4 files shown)


dario-piotrowicz:
Thanks a lot for the PR @james-elicx 😄

But it seems like the e2es are failing, and the failures do seem legit:
 - https://github.com/cloudflare/next-on-pages/actions/runs/8953245001/job/24609146706#step:4:768
 - https://github.com/cloudflare/next-on-pages/actions/runs/8953245001/job/24609146706#step:4:787
 
Could you please have a look? 🙏

james-elicx:
argh, i was hoping a rerun might fix them. welp. that's irritating

dario-piotrowicz:
> argh, i was hoping a rerun might fix them. welp. that's irritating

yeah no, I already did try rerunning them 🥲

Also usually they kind of either all run or all (or most) fail, when just a few tests fail like now I think that does indicate a real regression/issue 😅

ehabsakran:
What's the status of this PR? I'm having this issue with Cloudflare pages and next-firebase-auth-edge library.

james-elicx:
> What's the status of this PR? I'm having this issue with Cloudflare pages and next-firebase-auth-edge library.

i think it's probably fine to merge personally.

the area of caution is we haven't done any tests about if this would cause a regression for how the fetch cache in nextjs works. i feel like it would probably be okay though.

PR #789, opened 2024-05-19
Fix city header encoding (RFC3986)

### LINKED ISSUE #788: [🐛 Bug]: x-vercel-ip-city header must be encoded (RFC3986)

_No response_

### Description

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:19:22 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T8112
        CPU: (8) arm64 Apple M2
        Memory: 16 GB
        Shell: /bin/zsh
Package Manager Used: npm (10.2.4)

Relevant Packages:
        vercel: N/A
        next: 14.2.3

### Reproduction

When accessing from a location where there are non ASCII characters in the city name I get the following error:

```
✘ [ERROR] Problematic header name or value: "Münsingen" (raw bytes: "\x4d\xfc\x6e\x73\x69\x6e\x67\x65\x6e"). This string contains 8-bit characters in the range 0x80 - 0xFF. As a quirk to support Unicode, we encode header strings in UTF-8, meaning the actual header name/value on the wire will be "\x4d\xc3\xbc\x6e\x73\x69\x6e\x67\x65\x6e". Consider encoding this string in ASCII for compatibility with browser implementations of the Fetch specifications.
```

The city name for the location of the requester's public IP address must be encoded according to [RFC3986](https://tools.ietf.org/html/rfc3986).  as described in the [Vercel Docs](https://vercel.com/docs/edge-network/headers#x-vercel-ip-city).

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

marcojakob:
Fixes #788

### DIFF
--- a/.changeset/city-header-encoding.md
+++ b/.changeset/city-header-encoding.md
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/next-on-pages": minor
+'@cloudflare/next-on-pages': minor
 ---
 
 Fix: The city name for the location of the requester's public IP address must be encoded according to RFC3986.
--- a/.changeset/city-header-encoding.md
+++ b/.changeset/city-header-encoding.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/next-on-pages": minor
+---
+
+Fix: The city name for the location of the requester's public IP address must be encoded according to RFC3986.
--- a/package-lock.json
+++ b/package-lock.json
@@ -14611,7 +14611,7 @@
 			}
 		},
 		"packages/eslint-plugin-next-on-pages": {
-			"version": "1.11.3",
+			"version": "1.11.0",
 			"license": "MIT",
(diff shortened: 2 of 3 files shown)


dario-piotrowicz (commented):
@marcojakob thanks so very much for the fix, it looks great to me! ❤️ 

Could you just add a path changeset to the PR? (via `npm run changeset`)

dario-piotrowicz:
@marcojakob also the formatting please 😅🙏 

(or if you prefer I can do that for you and push on your branch 🙂)

dario-piotrowicz (approved):
Once again @marcojakob thanks so very much for the nice fix ❤️

PR #811, opened 2024-06-23
make sure headers set in middlewares are accessibly by layouts and pages

### LINKED ISSUE #797: [🐛 Bug]: Failed to read header in page / layout after setting it in the middleware

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 23.5.0: Wed May  1 20:14:38 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6020
        CPU: (10) arm64 Apple M2 Pro
        Memory: 16 GB
        Shell: /bin/zsh

Relevant Packages:
        vercel: 34.2.4
        next: 14.1.0

### Description

Cloudflare next-on-pages fails to read header in next js pages/layout after setting it in the middleware.

### Reproduction

I have created a minimal reproducible example in this repo - [clmnin/cf-middleware-set-header](https://github.com/clmnin/cf-middleware-set-header)

## How to set it up

```sh
git clone https://github.com/clmnin/cf-middleware-set-header.git
cd cf-middleware-set-header
npm i
```

### Run NextJS Dev to see what are the expected outputs
```sh
npm run dev
# console log "token-value" for the middleware, page and layout
# ▲ Next.js 14.1.0
# - Local:        http://localhost:3000
# 1. middleware: token-value
# 2. page: token-value
# 3. layout: token-value
```

This is the expected output for a next js app. The page and layout routes get the response passed by the middleware.

### Run Cloudflare pages preview using wrangler
```sh
npm run preview
# fails to console log "token-value" for the middleware, page and layout
# [wrangler:inf] Ready on http://localhost:8788
# 1. middleware: token-value
# 2. page: null
# 3. layout: null
```

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

dario-piotrowicz:
fixes #797


pages-e2e/vitest.config.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="vitest" />
```
petebacondarwin, line 1: Why is this needed? There are no ambient types being used here.

petebacondarwin (approved):

pages-e2e/vitest.config.ts
dario-piotrowicz: it's needed because it augments the UserConfig type:
Screenshot 2024-06-26 at 10 20 49

Without it typescript is not happy about the `test` field:
Screenshot 2024-06-26 at 10 20 30

petebacondarwin: Is that not solved by adding it to a tsconfig.json? Not blocking this, of course. Just interested.

dario-piotrowicz: yeah I feel like that'd be an alternative solution, I don't feel like one is necessarily better than the other, I can give that a quick try 🙂 👍

dario-piotrowicz: `tsconfig` updated instead 🙂👍

dario-piotrowicz: actually just including this file in the `tsconfig` was enough to fix the issue 😅 

I guess typescript is picking up the vitest `d.ts` files by itself already 🤔

PR #819, opened 2024-06-29
introduce in-memory mutex to avoid race condition issues in parallel workers runs

### LINKED ISSUE #805: [🐛 Bug]: React contexts break when loading pages in parallel

```
System:
	Platform: darwin
	Arch: arm64
	Version: Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000
	CPU: (10) arm64 Apple M1 Pro
	Memory: 32 GB
	Shell: /bin/zsh
Package Manager Used: npm (10.7.0)

Relevant Packages:
	vercel: 34.2.7
	next: 14.2.4
```

### Description

When accessing any next pages that use contexts in _parallel_ we get 500 `Internal Server Error` with:

```
✘ [ERROR] TypeError: Cannot read properties of null (reading 'useContext')
```

This has been a very tricky bug to hunt down, but we experience this when running wrangler locally, in our end-to-end test suites, and we have metrics that show us that this affects our users in production.

The pages must be loaded in quick succession. There must be _several_ pages that are visited, not the same page.

To reproduce this error, in the reproduction repo run:

1. Run `npm run preview` in one tab.
2. In the next tab `npx playwright test --repeat-each=5`
3. Observe `TypeError: Cannot read properties of null (reading 'useContext')` in the logs + 500 internal server error

### Reproduction

https://github.com/ostenbom/reproduce-next-parallel-context

### Pages Deployment Method

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

dario-piotrowicz:
fixes #805

### DIFF
--- a/packages/next-on-pages/templates/in-memory-mutex.ts
+++ b/packages/next-on-pages/templates/in-memory-mutex.ts
@@ -1,0 +1,58 @@
+/**
+ * What is this file for?
+ *
+ * workerd isolates share the same modules registry meaning that it can happen for the same isolate to run
+ * multiple instances of the same worker and those will share the modules' state.
+ *
+ * This can be problematic here as React uses modules state to set and get some data
+ * (in particular the ReactSharedInternals: https://github.com/facebook/react/blob/2e72ea8401/packages/shared/ReactSharedInternals.js)
+ * and if different workers end up sharing the same modules state this can lead to race conditions and unexpected bugs
+ * (for example: https://github.com/cloudflare/next-on-pages/issues/805).
+ *
+ * So this module here is an auxiliary module that we use to see if workers happen to be sharing the same modules state,
+ * and if that's the case this also allows us not to run the workers in parallel (to avoid the above mentioned issue).
+ *
+ * Things to note:
+ *  - this is a hopefully temporary workaround and should be addressed by workerd instead
+ *  - this does make request handling for next-on-pages workers potentially slower, but only when
+ *    the workers happen to share the same modules state (meaning that they run in the same isolate at around
+ *    the same time), so although not ideal is still better than running them in an unpredictable state
+ *  - this addresses the fact that workers don't end up running in parallel and generating race conditions because
+ *    of the shared modules state, but it does not address the fact that the modules state can be left dirty from a worker's
+ *    run and that that might effect the next worker's run, hopefully this is not an issue (i.e. React doesn't rely on modules
+ *    to necessarily be clean before a run). I think this last point can only be addressed by workerd itself.
+ */
+
+let queue: string[] = [];
+
+/**
+ * This function allows the acquisition of the next-on-pages in-memory mutex, it has
+ * to be run before the worker logic runs and it has to run the returned done/cleanup
+ * function (to free the mutex for the next worker run)
+ *
+ * @param uuid the unique id which identities a worker run
+ * @returns a done/cleanup function if the worker is allowed to run, null otherwise
+ */
+export function run(uuid: string) {
+	if (!queue.includes(uuid)) {
+		// add worker to queue
+		queue.push(uuid);
+	}
+
+	if (queue[0] !== uuid) {
+		// there's a different worker in the queue
+		return null;
+	}
+
+	// this worker is the first in the queue so it can run
+	return () => {
+		const doneId = queue.shift();
+		if (doneId !== uuid) {
+			// eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


ostenbom:
@dario-piotrowicz  We haven't managed to get this pre-release in production for us because there seems to be a deadlock somewhere (on a small proportion of requests). Really struggling to create a reproducible case since the worker isn't flushing log lines until the request completes and the request never does complete 😅

dario-piotrowicz:
> @dario-piotrowicz We haven't managed to get this pre-release in production for us because there seems to be a deadlock somewhere (on a small proportion of requests). Really struggling to create a reproducible case since the worker isn't flushing log lines until the request completes and the request never does complete 😅

oh... ok that's not good 😓 

let me have a think about it 😓

dario-piotrowicz:
@ostenbom I've added a `finally` block for releasing the mutex, I hope this should be enough to solve your issue, could you please check if you still get the issue on your end and let me know? 🙏 

(if this doesn't fix it I have another idea in mind, but I hope that this should be enough)

ostenbom:
@dario-piotrowicz 

Here are the very strange conditions under which I can get it to deadlock 🤪 :

- [Added a middleware to do a `NextResponse.rewrite()`](https://github.com/ostenbom/reproduce-next-parallel-context/commit/6faaaef07eeb2c6787b5659ca43bd3300aad0eba)
- Deploy to cf pages
- `curl -I https://my-parallel-app.pages.dev/ -H 'x-forwarded-host: 019c8a6b.my-parallel-app.pages.dev'`
- Deadlock, eventually we see 2 cancelled requests in the logs to the two different domains of the pages worker

Screenshot 2024-07-04 at 17 31 43

The last `finally` commit didn't do the trick unfortunately. Could it have to do with the behaviour of `NextResponse.rewrite()`?

ostenbom:
@dario-piotrowicz anything more you can do here? Or is this a not-so-expected use case? Appreciate your help here! ⭐

packages/next-on-pages/templates/in-memory-mutex.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * What is this file for?
+ *
+ * workerd isolates share the same modules registry meaning that it can happen for the same isolate to run
+ * multiple instances of the same worker and those will share the modules' state.
```
kentonv, line 5: An isolate by definition runs only one "instance" of the Worker. However, one Worker instance may handle many requests, possibly at the same time, possibly from different users. Global state will be shared across all these requests.

Basically all servers work this way, including Node.js. We actually would like to change this in Workers, but it requires building new fundamental technology.

ostenbom:
Sorry to keep pinging you @dario-piotrowicz but this is affecting hundreds of our users every day, and is our main source of instability on `next-on-pages`, is there anything we can do to fix this?

dario-piotrowicz:
@ostenbom sorry, I meant to reply to your older ping sooner 😓 

I don't think that this is a not-so-expected use case and I do believe we need to fix it asap...

I simply hadn't had time to dedicate to this (as I've been working on other unrelated projects)

I don't think that there's anything specific you can do on your end 😓 

anyways I'm really sorry that this is affecting many of your users, I will try my best to prioritize this work and I promise to either give you a solution or some concrete info by the end of this week/begging of the next (monday at the very latest) 🙏

dario-piotrowicz:
Hi @ostenbom 🙂 

I've tried a different solution for this in https://github.com/cloudflare/next-on-pages/pull/836
(since the in-memory mutex was just a first attempt which would always at best be a sub-optimal solution)

I've been testing this with your [reproduction](https://github.com/ostenbom/reproduce-next-parallel-context) and it seems to be working smoothly 😃 (hopefully I am not getting false positives for some reason 🤔)

I still want to polish it a bit and analyze more this approach, but please if you want try the [PR's prerelease](https://github.com/cloudflare/next-on-pages/pull/836#issuecomment-2245163482) and let me know if it works for you 🙂

ostenbom:
Thanks @dario-piotrowicz! This one looks good! Both passing the reproduction and our internal tests, plus not deadlocking! 😍

dario-piotrowicz:
@ostenbom Awesome! thanks for confirming! 😄 

I'll close this PR then and let's just move to the other one, I need to do some light polishing and I think it should be good to go 🙂

PR #836, opened 2024-07-23
implement route scoping solution

### LINKED ISSUE #805: [🐛 Bug]: React contexts break when loading pages in parallel

dario-piotrowicz:
currently routes all share the same global scope, this can be problematic and cause race conditions and failures, like the one presented in #805 that happens when various requests are handled in parallel (and the global scope ends up in an invalid state).

So, this PR solves the issue by wrapping each js file (functions and chunks included) in functions that override the `self`, `globalThis` and `global` symbols with proxies that try to make (as reasonably well as possible) each route's global scope as isolated as possible.
____

fixes #805

### DIFF
--- a/.changeset/thirty-birds-build.md
+++ b/.changeset/thirty-birds-build.md
@@ -1,0 +1,32 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+fix: implement route specific global scoping strategy
+
+currently routes all share the same global scope, this can be problematic and cause
+race conditions and failures
+
+One example of this is the following code that is present in route function files:
+
+```ts
+self.webpackChunk_N_E = ...
+```
+
+and
+
+```ts
+self.webpackChunk_N_E.push(...)
+```
+
+this indicates that an in-memory global collection of the webpack chunks is shared by all routes,
+this combined with the fact that chunks can have their own module state this can easily cause routes
+to conflict with each other at runtime.
+
+So, in order to solve the above issue wrap every route function in a function wrapper which
+accepts as parameters, thus overrides, the `self`, `globalThis` and `global` symbols. The symbols
+are then to be resolved with proxies that redirect setters to route-scoped in-memory maps and
+getters to the above mentioned map's values and fallback to the original symbol values otherwise
+(i.e. `globalThis` will be overridden by a proxy that, when setting values, sets them in a separate
+location and, when getting values, gets them from said location if present there or from the real
+`globalThis` otherwise)
--- a/.changeset/thirty-birds-build.md
+++ b/.changeset/thirty-birds-build.md
@@ -20,12 +20,10 @@ self.webpackChunk_N_E.push(...)
 ```
 
 this indicates that an in-memory global collection of the webpack chunks is shared by all routes,
-this combined with the fact that chunks can have their own module state this can easily cause routes
-to conflict with each other at runtime.
+this combined with the fact that chunks can have their own module state this can easily cause routes to conflict with each other at runtime.
 
-So, in order to solve the above issue wrap every route function in a function wrapper which
-accepts as parameters, thus overrides, the `self`, `globalThis` and `global` symbols. The symbols
-are then to be resolved with proxies that redirect setters to route-scoped in-memory maps and
+So, in order to solve the above issue, all route functions are wrapped in a function which accepts as parameters, thus overrides, the `self`, `globalThis` and `global` symbols. The symbols
+will be resolved with proxies that redirect setters to route-scoped in-memory maps and
 getters to the above mentioned map's values and fallback to the original symbol values otherwise
 (i.e. `globalThis` will be overridden by a proxy that, when setting values, sets them in a separate
 location and, when getting values, gets them from said location if present there or from the real
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -350,7 +350,7 @@ async function prependWasmImportsToCodeBlocks(
 					relativeTo: nopDistDir,
 				});
 
-				let functionImports = '';
+				const functionImports: string[] = [];
 
 				for (const identifier of wasmImports) {
 					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
@@ -358,11 +358,14 @@ async function prependWasmImportsToCodeBlocks(
 					const wasmImportPath = normalizePath(
 						join(relativeImportPath, newDest as string),
 					);
-					functionImports += `import ${identifier} from "${wasmImportPath}";\n`;
+					functionImports.push(`import ${identifier} from "${wasmImportPath}"`);
 				}
 
 				const oldContents = await readFile(filePath);
-				await writeFile(filePath, `${functionImports}${oldContents}`);
+				await writeFile(
+					filePath,
+					`${functionImports.join(';')};${oldContents}`,
+				);
 			},
 		),
 	);
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -233,7 +233,7 @@ async function buildFunctionFile(
 	groupedImports.forEach((keys, path) => {
 		if (path.endsWith('.wasm')) {
 			// we don't need/want to apply any code transformation for wasm imports
-			functionImports += `import ${keys} from "${path}"`;
+			functionImports += `import ${keys} from "${path}";`;
 			return;
 		}
 
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -250,7 +250,7 @@ async function buildFunctionFile(
 
 	fnInfo.outputPath = relative(workerJsDir, newFnPath);
 
-	const finalFileContents = iffefyFunctionFile(
+	const finalFileContents = iifefyFunctionFile(
 		fileContents,
 		functionImports,
 		fnInfo,
@@ -283,7 +283,7 @@ type BuildFunctionFileOpts = {
  * @param chunksExportsMap a map containing getters and chunks identifiers being used by the function
  * @returns the updated/iifefied file content
  */
-function iffefyFunctionFile(
+function iifefyFunctionFile(
 	fileContents: string,
 	functionImports: string[],
 	fnInfo: FunctionInfo,
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -282,14 +282,14 @@ type BuildFunctionFileOpts = {
  *
  * @param fileContents the function file's contents
  * @param functionImports the imports that need to be added to the file
- * @param fnInfo the function's information
+ * @param functionInfo the function's information
  * @param chunksExportsMap a map containing getters and chunks identifiers being used by the function
  * @returns the updated/iifefied file content
  */
 function iifefyFunctionFile(
 	fileContents: string,
 	functionImports: string[],
-	fnInfo: FunctionInfo,
+	functionInfo: FunctionInfo,
 	chunksExportsMap: Map<string, Set<string>>,
 ): string {
 	const wrappedContent = `
@@ -307,7 +307,7 @@ function iifefyFunctionFile(
 	`;
 
 	const proxyCall = `const proxy = globalThis.__nextOnPagesRoutesIsolation.getProxyFor('${
-		fnInfo.route?.path ?? ''
+		functionInfo.route?.path ?? ''
 	}');`;
 
 	const chunksExtraction = [...chunksExportsMap.entries()].flatMap(
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -216,7 +216,7 @@ async function buildFunctionFile(
 	{ importsToPrepend }: { importsToPrepend: NewImportInfo[] },
 	{ workerJsDir, nopDistDir }: ProcessVercelFunctionsOpts,
 ): Promise<{ buildPromise: Promise<void> }> {
-	let functionImports = '';
+	const functionImports: string[] = [];
 
 	// Group the identifier imports by the keys for each path.
 	const groupedImports = importsToPrepend.reduce((acc, { key, path }) => {
@@ -242,14 +242,16 @@ async function buildFunctionFile(
 			// if we're dealing with a wasm file there is a single default export to deal with
 			const defaultExport = exports;
 			// we don't need/want to apply any code transformation for wasm imports
-			functionImports += `import ${defaultExport} from "${path}";`;
+			functionImports.push(`import ${defaultExport} from "${path}"`);
 			return;
 		}
 
 		const namedExportsId = `getNamedExports_${chunkMapIdx++}`;
 		const exportKeys = exports.split(',');
 		chunksExportsMap.set(namedExportsId, new Set(exportKeys));
-		functionImports += `import { getNamedExports as ${namedExportsId} } from '${importPath}';\n`;
+		functionImports.push(
+			`import { getNamedExports as ${namedExportsId} } from '${importPath}'`,
+		);
 	});
 
 	fnInfo.outputPath = relative(workerJsDir, newFnPath);
@@ -289,7 +291,7 @@ type BuildFunctionFileOpts = {
  */
 function iffefyFunctionFile(
 	fileContents: string,
-	functionImports: string,
+	functionImports: string[],
 	fnInfo: FunctionInfo,
 	chunksExportsMap: Map<string, Set<string>>,
 ): string {
@@ -319,9 +321,12 @@ function iffefyFunctionFile(
 		},
 	);
 
-	return [functionImports, proxyCall, ...chunksExtraction, wrappedContent].join(
-		';',
-	);
+	return [
+		...functionImports,
+		proxyCall,
+		...chunksExtraction,
+		wrappedContent,
+	].join(';');
 }
 
 /**
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -246,7 +246,7 @@ async function buildFunctionFile(
 		);
 
 		const namedExportsId = `getNamedExports_${chunkMapIdx++}`;
-		chunksExportsMap.set(namedExportsId, new Set([...keys.split(',')]));
+		chunksExportsMap.set(namedExportsId, new Set(keys.split(',')));
 		functionImports += `import { getNamedExports as ${namedExportsId} } from '${importPath}';\n`;
 	});
 
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -690,4 +690,4 @@ async function processBundledAssets(
  *
  * This is the name of the object used for such exports.
  */
-const namedExportsObjectName = '__next_on_pages__named_exports_object__';
+const namedExportsObjectName = '__nextOnPagesNamedExportsObject';
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -690,4 +690,4 @@ async function processBundledAssets(
  *
  * This is the name of the object used for such exports.
  */
-const namedExportsObjectName = '__next-on-pages-named_exports_object__';
+const namedExportsObjectName = '__next_on_pages__named_exports_object__';
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -190,13 +190,7 @@ async function functionifyFileContent(path: string) {
 	const originalFileContents = await readFile(path, 'utf8');
 	return `
 		const namedExports = {};
-		export const getNamedExports = ((self, globalThis, global) => {
-			${originalFileContents.replace(
-				/export\s+const\s+(\S+)\s*=/g,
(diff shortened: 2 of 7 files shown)


.changeset/thirty-birds-build.md
```diff
@@ -0,0 +1,32 @@
...
+
+```ts
+self.webpackChunk_N_E.push(...)
+```
+
+this indicates that an in-memory global collection of the webpack chunks is shared by all routes,
+this combined with the fact that chunks can have their own module state this can easily cause routes
+to conflict with each other at runtime.
+
+So, in order to solve the above issue wrap every route function in a function wrapper which
```
vicb, line 26: Looks like some words are missing in this sentence

> So, in order to solve the above issue wrap every route function in a function wrapper which

Suggest

>  So, in order to solve the above issue, all route functions are wrapped in a function which

.changeset/thirty-birds-build.md
```diff
@@ -0,0 +1,32 @@
...
+accepts as parameters, thus overrides, the `self`, `globalThis` and `global` symbols. The symbols
+are then to be resolved with proxies that redirect setters to route-scoped in-memory maps and
```
vicb, line 28: suggest: are then to be -> will be resolved

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -168,6 +176,32 @@ async function processFunctionIdentifiers(
...
+ * (which are used as the function's parameter names)
+ *
+ * @param path the path of the ESM file
+ * @returns the converted file content
+ */
+async function functionifyFileContent(path: string) {
+	let fileContents = await readFile(path, 'utf8');
+	fileContents = `const namedExports = {};${fileContents}`;
+	fileContents = fileContents.replace(
+		/export const (\S+) =/g,
```
vicb, line 193: isn't `/export const (\S+) =/g` is a bit too strict?

what about:

```
export\s+const\s+(\S+)\s*=
```

dario-piotrowicz: sure, good call 🙂👍

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -193,7 +227,16 @@ async function buildFunctionFile(
 		return acc;
 	}, new Map<string, string>());
 
+	let chunkMapIdx = 0;
+	const chunksExportsMap = new Map<string, Set<string>>();
+
 	groupedImports.forEach((keys, path) => {
+		if (path.endsWith('.wasm')) {
+			// we don't need/want to apply any code transformation for wasm imports
+			functionImports += `import ${keys} from "${path}"`;
```
vicb, line 236: maybe add `;`

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -202,12 +245,19 @@ async function buildFunctionFile(
 			join(relativeImportPath, addLeadingSlash(path)),
 		);
 
-		functionImports += `import { ${keys} } from '${importPath}';\n`;
+		const namedExportsId = `getNamedExports_${chunkMapIdx++}`;
+		chunksExportsMap.set(namedExportsId, new Set([...keys.split(',')]));
```
vicb, line 249: Suggest: replace with `.filter`to be more explicit about the intention of `[...]`

.changeset/thirty-birds-build.md
dario-piotrowicz: well, words are not missing, maybe the sentence is just a bit unclear to follow 😅 

I tend to use the imperative tense for both the commit title and description (as I believe that to be a best practice on writing good commits), so sometimes the result can sound a bit forced 😕😓 

I'll try to improve the wording, while still keep it imperative 🙂 (unless you prefer descriptions not to be in imperative)

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -168,6 +176,32 @@ async function processFunctionIdentifiers(
...
+		/export const (\S+) =/g,
+		'namedExports["$1"] =',
+	);
+	fileContents = `
+		export const getNamedExports = ((self, globalThis, global) => {
+			${fileContents};
+			return namedExports;
+		});
+	`;
+	return fileContents;
```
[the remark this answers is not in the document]
dario-piotrowicz: I'll give it a try 🙂👍

vicb (commented):
I have done a first pass of review with my very limited knowedlge.

I'll do a second pass of review tomorrow.

One thing that would be nice is a high level description of what the PR do - 1 or 2 sentences to give context, describes the problem that is solved (the content is currently empty).

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: sorry I don't follow, how are you suggesting to use `.filter`? 🙂

vicb: I thought `[...keys.split(',')]` would filter out the empty strings but it doesn't so you can actually replace it with `keys.split(',')` only

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -168,6 +176,32 @@ async function processFunctionIdentifiers(
...
+ *
+ * @param path the path of the ESM file
+ * @returns the converted file content
+ */
+async function functionifyFileContent(path: string) {
+	let fileContents = await readFile(path, 'utf8');
+	fileContents = `const namedExports = {};${fileContents}`;
+	fileContents = fileContents.replace(
+		/export const (\S+) =/g,
+		'namedExports["$1"] =',
```
james-elicx, line 194: i'm not sure this will work quite as well as we're hoping.

I can't seem to find any function files that have `export const ... =` in them in either of the projects i've looked in, so im not sure where this would be updating things?

It also looks like this would overwrite the next.js error messages and add them to the named exports object which probably isn't ideal.

image

.changeset/thirty-birds-build.md
dario-piotrowicz: I couldn't think of an imperative wording that would be much more clear than the current version 😕 

So I'm fine with what you suggested 👍

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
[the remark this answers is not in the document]
dario-piotrowicz: looks quite better, thanks a bunch for the suggestion! 🚀

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: oh yeah.... why did I was spreading 😕

must have been because of some refactoring or something! 😓 

good catch! Thanks! ❤️

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: > I can't seem to find any function files that have export const ... = in them in either of the projects i've looked in, so im not sure where this would be updating things? Unless it's only certain use cases this would apply to?

mh... I'll have a look 👀 

___

I'm not sure I understand the comment about the Next.js error messages... could you clarify what you mean?

dario-piotrowicz: Are you sure the regex replacement doesn't do anything? I just created a brand new next-on-pages app using C3 and tried to run this branch's build with the regex replacement removed and I immediately get these build errors:
Screenshot 2024-07-31 at 23 36 05

To be fair these are exports next-on-pages itself adds... so maybe you only checked in the Vercel output files?

pages-e2e/fixtures/app14.0.0/main.fixture
```diff
@@ -4,7 +4,6 @@
 		"appRouting",
 		"appConfigsTrailingSlashTrue",
 		"appWasm",
-		"appServerActions",
```
dario-piotrowicz, line 7: Note: I've removed this feature from the app14.0.0 fixture... this is definitely not ideal but I couldn't figure out why this was failing in the e2es

If you try to run the app14.0.0 fixture with the feature included you will see that the application does work as intended (sort of... it seems like server actions do present some problems... but those are pre-existing problems that also main presents)

Also this feature is included in the appLatest fixture, so we are indeed still testing this functionality...

If we really believe it worth it I can spend more time trying to see if there's a way to keep the feature in app14.0.0

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -193,7 +225,16 @@ async function buildFunctionFile(
 		return acc;
 	}, new Map<string, string>());
 
+	let chunkMapIdx = 0;
+	const chunksExportsMap = new Map<string, Set<string>>();
+
 	groupedImports.forEach((keys, path) => {
+		if (path.endsWith('.wasm')) {
+			// we don't need/want to apply any code transformation for wasm imports
+			functionImports += `import ${keys} from "${path}";`;
```
vicb, line 234: Should this be `import {${keys}} from "${path}";`?

vicb: Also should the ìmportPath`be used?

nit: it looks like `relativeImportPath`can be moved out of the loop?

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -225,6 +273,53 @@ type BuildFunctionFileOpts = {
...
+				// the default export needs to become the return value of the iife, which is then re-exported as default
+				.replace(/export default /g, 'return ')}
+		})(proxy, proxy, proxy);
+	`;
+
+	const proxyCall = `const proxy = globalThis.__nextOnPagesRoutesIsolation.getProxyFor('${
+		fnInfo.route?.path ?? ''
+	}');`;
+
+	const chunksExtraction = [...chunksExportsMap].flatMap(
```
vicb, line 307: nit: replace with `chunksExportsMap.entries()`

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -225,6 +273,53 @@ type BuildFunctionFileOpts = {
...
+		([getNamedExportsId, keys]) => {
+			return [
+				`const exportsOf${getNamedExportsId} = ${getNamedExportsId}(proxy, proxy, proxy);`,
+				...[...keys].map(
```
vicb, line 311: nit: `keys.entries.map(...)`

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
james-elicx: > could you clarify what you mean?

the regex would try and add an entry to the object and replace part of the error message

> To be fair these look like exports that next-on-pages itself adds/creates... so maybe you only checked in the Vercel output files?

i haven't run a build with this, but i'm looking at a build of cloudy from about a month ago, and i can't find `export const` anywhere in the function files (vercel or next-on-pages).

vercel functions:
image

_worker.js dir:
image

vicb (approved):
Added a few minor comments.

I definitely need to play more with the project to understand all the implications.

One thing I would love to see is added unit tests. They would help making sure the transformations are what we expect and they could be a good starting point to write a non regression test if bugs get reported.

I'll approve for now and will discuss with you if I think about any other stuff later in the day.

Thanks

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: Screenshot 2024-08-01 at 10 18 00

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: Screenshot 2024-08-01 at 10 19 02

I can do something like this if you find the spreading too difficult to read?
Screenshot 2024-08-01 at 10 19 20

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
vicb: My idea was to avoid creating the whole array upfront but then I guess `[... chunksExportsMap.entries()]` (adding the explicit entries)

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
vicb: Same comment as above, maybe only add `.entries()` to make the code explicit.

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: > Should this be `import {${keys}} from "${path}";`?

Screenshot 2024-08-01 at 10 37 20

I think this works as intended, the fact is that wasm have only a single default export (I think!)

I should definitely at least update the code to make this more explicit! 👍

dario-piotrowicz: > Also should the ìmportPath`be used?

> nit: it looks like relativeImportPathcan be moved out of the loop?

Yes and yes 🙂 , awesome, thanks a lot for catching those! 🙏

dario-piotrowicz: Regarding the import string, I've updated the code to make it more clear (https://github.com/cloudflare/next-on-pages/pull/836/commits/8a3686154991c7adeb91dbb30a33a39c661a7313) please let me know if you find this better 🙂

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -193,21 +225,41 @@ async function buildFunctionFile(
...
 		const importPath = normalizePath(
 			join(relativeImportPath, addLeadingSlash(path)),
 		);
 
-		functionImports += `import { ${keys} } from '${importPath}';\n`;
+		if (path.endsWith('.wasm')) {
+			// if we're dealing with a wasm file there is a single default export to deal with
+			const defaultExport = exports;
+			// we don't need/want to apply any code transformation for wasm imports
+			functionImports += `import ${defaultExport} from "${path}";`;
```
vicb, line 245: suggest: what about making this a `string[]`

it's generally not recommended to += strings (not memory/cpu efficient)
you can `.join('\n')` after the loop to make that consistent.

Is it expected that the export map is not updated form wasm?

dario-piotrowicz: Good point, updated thanks 🙂👍

> Is it expected that the export map is not updated for wasm?

do you mean the `chunksExportsMap`?
if so yes, that map is only for the chunks, where we apply the code transformation (that wraps the chunks in functions that we then call), wasm files are not part of this

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
james-elicx: the suspense cache isn't one - that's just a variable we use internally. but there's plenty of references to global, self, etc. in these function files but none of them would go through the proxy

dario-piotrowicz: > the suspense cache isn't one

yeah no, my bad, we pass the cache files trhough the functions build pipeline

> but there's plenty of references to global, self, etc. in these function files but none of them would go through the proxy

what do you mean? could you clarify? why would they not go through the proxy?

james-elicx: actually nevermind. i think im just misunderstanding how the named exports and the function wrapping the file is used

dario-piotrowicz: Note: We do get `export const` in at this step, especially since we've added the lines outselves earlier:
https://github.com/cloudflare/next-on-pages/blob/0a07f340b868bb887e465b86557efdae793c348f/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts#L386

The fact that we don't get them in the final output, before this PR, I believe is because those get converted by esbuild to exports-as: 
Screenshot 2024-08-01 at 13 16 11

I am really wondering if we could not add the `export const`s ourselves.... but by debugging this I noticed that there are indeed a few `export const`s that we're not adding... so I might need to keep things as they are unfortunately 😕

dario-piotrowicz: > actually nevermind. i think im just misunderstanding how the named exports and the function wrapping the file is used

no come on, if you have any concern I am happy to hear them 🙂🙏 

shall we move this async and discuss things?

dario-piotrowicz: @james-elicx regex replacement removed! 🚀 https://github.com/cloudflare/next-on-pages/pull/836/commits/b13338aebe83572a3e34a9cf9221e765afaf0547

The less of these we have the better! thanks a lot for pointing this and making me improve the solution! ❤️

dario-piotrowicz: 🚀 

Screenshot 2024-08-01 at 17 29 15

dario-piotrowicz: Resolved as discussed async 😄

pages-e2e/features/customEntrypoint/assets/custom-entrypoint.ts
```diff
@@ -1,3 +1,4 @@
+//@ts-nocheck
```
dario-piotrowicz, line 1: For context, this is required since in the e2es we call next-on-pages with `npx` and we don't install it as a dependency (we did try that but the Pages e2es framework wasn't handling that correctly, so we had to pivot to just `npx` 😓)

___

This was actually working without the ts-nocheck comment when it was introduced: https://github.com/cloudflare/next-on-pages/pull/828
My best guess as to why it used to work but it doesn't anymore is that the Vercel CLI would ignore ts files that it didn't recognize but latest versions of the CLI (in the e2e tests we use always the latest Vercel CLI release) do try to compile all the ts files in the project's directory, generating the build issue here (since, legitimately, the next-on-pages dependency could not be solved)

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -193,21 +222,43 @@ async function buildFunctionFile(
...
 		const importPath = normalizePath(
 			join(relativeImportPath, addLeadingSlash(path)),
 		);
 
-		functionImports += `import { ${keys} } from '${importPath}';\n`;
+		if (path.endsWith('.wasm')) {
+			// if we're dealing with a wasm file there is a single default export to deal with
+			const defaultExport = exports;
+			// we don't need/want to apply any code transformation for wasm imports
+			functionImports.push(`import ${defaultExport} from "${path}"`);
```
vicb, line 242: super nit: consistent quotes here and on l250 - and `;` never harms ;)

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -225,6 +276,60 @@ type BuildFunctionFileOpts = {
...
+			];
+		},
+	);
+
+	return [
+		...functionImports,
+		proxyCall,
+		...chunksExtraction,
+		wrappedContent,
+	].join(';');
```
vicb, line 330: IMO `;` should be in the source and `\n` used to join the lines. - minifying should not be the concern of this, right?

vicb: By joining with `;` you assume that whoever will update the code later will know not to add `;`

vicb (approved):

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: Yeah good point 🙂 :+1:

Updated 👍

packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
dario-piotrowicz: yeah I noticed the quotes too! 😅 👍 

Updated both the quotes and `;`

PR #838, opened 2024-07-28
Fix `x-vercel-ip-country-region` header's value

### LINKED ISSUE #839: [🐛 Bug]: `x-vercel-ip-country` not filled correctly.

System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:49 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6020
        CPU: (12) arm64 Apple M2 Max
        Memory: 64 GB
        Shell: /bin/zsh
Package Manager Used: npm (10.8.1)

Relevant Packages:
        @cloudflare/next-on-pages: 1.12.1
        vercel: N/A
        next: N/A

### Description

According to the doc, [`x-vercel-ip-country-region`](https://vercel.com/docs/edge-network/headers#x-vercel-ip-country-region) should be an [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) code.

But for now we just fill the field with `request.cf.region`, which are full name of the region.

ref #838

### Reproduction

_No response_

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

- [x] Would you like to help fixing this bug?

longfangsong:
Fix #839 

~~Just like what @marcojakob observed on city names in #788, we should also encode region and country names for the location of the requester's public IP address.~~

`request.cf.region` is the full name of the region the user is in, however according to the doc, [x-vercel-ip-country-region](https://vercel.com/docs/edge-network/headers#x-vercel-ip-country-region) should be an [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) code, and I believe `request.cf.regionCode` is the correct value to be used here.

So that problems like what I met:

```
✘ [ERROR] Problematic header name or value: "Västra Götaland County" (raw bytes: "\x56\xe4\x73\x74\x72\x61\x20\x47\xf6\x74\x61\x6c\x61\x6e\x64\x20\x43\x6f\x75\x6e\x74\x79"). This string contains 8-bit characters in the range 0x80 - 0xFF. As a quirk to support Unicode, we encode header strings in UTF-8, meaning the actual header name/value on the wire will be "\x56\xc3\xa4\x73\x74\x72\x61\x20\x47\xc3\xb6\x74\x61\x6c\x61\x6e\x64\x20\x43\x6f\x75\x6e\x74\x79". Consider encoding this string in ASCII for compatibility with browser implementations of the Fetch specifications.
```

won't occur anymore.

### DIFF
--- a/.changeset/curly-weeks-sin.md
+++ b/.changeset/curly-weeks-sin.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Use `request.cf.regionCode` for `x-vercel-ip-country-region` for keeping behavior consistent with documentation
--- a/packages/next-on-pages/templates/_worker.js/utils/request.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/request.ts
@@ -17,7 +17,7 @@ export function adjustRequestForVercel(request: Request): Request {
 		adjustedHeaders.set('x-vercel-ip-country', request.cf.country as string);
 		adjustedHeaders.set(
 			'x-vercel-ip-country-region',
-			request.cf.region as string,
+			request.cf.regionCode as string,
 		);
 		adjustedHeaders.set('x-vercel-ip-latitude', request.cf.latitude as string);
 		adjustedHeaders.set(


dario-piotrowicz (approved):
Thanks a bunch @longfangsong 😄 

This looks good to me, although could you please format the code (with `npm run prettier:fix`)? 🙏

longfangsong:
> PS: not a huge deal... but I wonder why the [vercel docs](https://vercel.com/docs/edge-network/headers) don't specify that these headers should be encoded (as they do for [x-vercel-ip-city](https://vercel.com/docs/edge-network/headers#x-vercel-ip-city)) 🤔

Good catch, according to that doc, [`x-vercel-ip-country`](https://vercel.com/docs/edge-network/headers#x-vercel-ip-country) should be an [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) country code, and [`x-vercel-ip-country-region`](https://vercel.com/docs/edge-network/headers#x-vercel-ip-country-region) should be an [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) code.

However all information we can get from `request.cf` is the full name of the country/region, and unfortunately I don't know how to map these names to ISO codes (eg, in most libraries like [`iso-3166`](https://github.com/wooorm/iso-3166/blob/HEAD/2.js) it is `Västra Götalands län` instead of `Västra Götaland County`).

(May I know where these values in `request.cf` come from?)

For now I think it's up to you to merge this pr or not (as long as it suppress the error message about non-ascii header), but I may open an issue about the real problem since I don't know how to fix it for now.

longfangsong:
Update: I found `regionCode` in `cf` and `cf-ipcountry` in request headers, these seems to be what we need here, so I may fix the root problem now.

dario-piotrowicz (approved):
Once again, thanks so much @longfangsong 🫶 🚀

PR #846, opened 2024-08-02
replace regex replacemend logic in `iifefyFunctionFile` and handle `_ENTRIES` declarations

dario-piotrowicz:
This PR replaces the regex replacements introduced in the `iifefyFunctionFile` function in https://github.com/cloudflare/next-on-pages/pull/836 with with more proper AST code modding

As part of this I also noticed that we do not want to update `_ENTRIES` identifiers if they are actually declared in the file (as it happens in older Vercel CLI releases) so I've addressed this too.

> [!NOTE]
> This PR does not include a changeset since the changes in https://github.com/cloudflare/next-on-pages/pull/836 have not yet been released and this PR simply improves that solution

### DIFF
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
@@ -23,6 +23,8 @@ import {
 } from '../../utils';
 import { copyAssetFile } from './prerenderFunctions';
 import { cliError } from '../../cli';
+import * as recast from 'recast';
+import * as recastParser from 'recast/parsers/babel';
 
 /**
  * Dedupes edge functions that were found in the build output.
@@ -253,7 +255,7 @@ async function buildFunctionFile(
 
 	fnInfo.outputPath = relative(workerJsDir, newFnPath);
 
-	const finalFileContents = iifefyFunctionFile(
+	const finalFileContents = await iifefyFunctionFile(
 		fileContents,
 		functionImports,
 		fnInfo,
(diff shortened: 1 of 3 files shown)


packages/next-on-pages/src/buildApplication/processVercelFunctions/dedupeEdgeFunctions.ts
```diff
@@ -330,6 +325,79 @@ function iifefyFunctionFile(
...
 }
 
+/**
+ * iifefyFunctionFile converts a function file's content into an iife (so that we can apply global proxies to
+ * it), this function is used to prepare the file's content so that it is ready to be embedded into an iife
+ *
+ * @param fileContents the original file contents
+ * @returns the file contents updated and ready to be embedded into an iife
+ */
+function prepareFunctionFileContentForIifeConversion(fileContents: string) {
```
james-elicx, line 335: i would be curious to know what the impact on build-time is with this change - are you able to get some numbers for that?

james-elicx: also, we already use acorn for parsing asts - would that help here as well?

dario-piotrowicz: > also, we already use acorn for parsing asts - would that help here as well?

I don't think so since there all we care about is getting (and analyzing) the AST while here we want to actually traverse and manipulate it

dario-piotrowicz: > i would be curious to know what the impact on build-time is with this change - are you able to get some numbers for that?

I should be able to get some before vs after numbers if you want 🤔 

In any case this runs only on function files (not all js files) so... I hope it's not to costly 😓
Even if this was slower I think that a slower but more correct and robust solution would be preferable than a quick but hacky and brittle one, right?

dario-piotrowicz: > i would be curious to know what the impact on build-time is with this change - are you able to get some numbers for that?

Oh wow! it's actually pretty bad! 😖 (thanks so much for the useful comment! I didn't imagine the impact could be this big! ❤️)

I've created a brand new Next.js app with C3 and run the build script, with the latest next-on-pages release it takes around 0.27s but with this PR's prerelease it takes around 2.95s...

I wanted to have a look at a more complex app so I gave the [app-playground-edge app](https://github.com/dario-piotrowicz/next-apps-for-testing/tree/master/apps/app-playground-edge) a try as well, in that case the build time goes from around 1.75s to around 4.35s... so a bit better but still pretty bad 😕

So maybe it might be best to stick with regex replacement for now... 😓 😖 

@vicb what do you think?

james-elicx: I thought that might happen 😅. It was a big problem with how we originally did things before the big CLI rewrite. Back then, I changed it from two passes over the AST to one pass which helped speed things up quite a bit. So yeah... Umm... It'll slow it down a lot.

I wonder if it might be possible to try and do some of this during the original parsing of the AST? I'm not sure how much would need changing to make it work, but it'd be an interesting idea to look into

dario-piotrowicz: > I thought that might happen 😅. It was a big problem with how we originally did things before the big CLI rewrite. Back then, I changed it from two passes over the AST to one pass which helped speed things up quite a bit. So yeah... Umm... It'll slow it down a lot.


> I wonder if it might be possible to try and do some of this during the original parsing of the AST? I'm not sure how much would need changing to make it work, but it'd be an interesting idea to look into

vicb:
I see no reason to switch to AST for now as your experiment shows it's much slower.

But it sounds like the `handle _ENTRIES declarations` should be fixed.

Thanks!

dario-piotrowicz:
Closing in favour of https://github.com/cloudflare/next-on-pages/pull/852

PR #867, opened 2024-08-27
Add CLI option to ignore invalid functions

### LINKED ISSUE #845: [⚡ Feature]: Allow split deployments with latent nodejs runtime functions
### Description

I'm currently running a next.js app on vercel. The next.js app contains a mix of edge and nodejs APIs, however the nodejs endpoints do not run on vercel (due to a body size limit), and are deployed on a separate server. Vercel will proxy to this server with the help of a conditional rewrite defined in the next.config.js

```js
{
  rewrites: async () => ({
    beforeFiles: process.env.PROXY_NODE_URL
      ? [
          "/api/my-node-js-endpoint",
          // ...
        ].map((source) => ({
          source,
          destination: `${process.env.PROXY_NODE_URL}${source}`,
        }))
      : undefined,
    })
}
```

This allows the best of both worlds: the entire API can be defined in next.js and ran locally, and then divvied up for a production deployment.

I want to try out cloudflare's Next on pages but am receiving the following error:

> ⚡️ ERROR: Failed to produce a Cloudflare Pages build from the project.
> ⚡️ 
> ⚡️      The following routes were not configured to run with the Edge Runtime:
> ⚡️        - /api/my-node-js-endpoint

As far as I can tell, the invalid function check is a heuristic that is eagerly throwing an error for functions that can never be invoked. 

My request:
- Allow the build system to ignore invalid functions based on a CLI argument (`--force` 🤔 ?)
- Or add heuristics that won't consider a function invalid if a rewrite will always take precedence

### Additional Information

_No response_

### Would you like to help?

- [X] Would you like to help implement this feature?

nickbabcock:
The presence of invalid functions shouldn't force a build failure as a deployment can still be valid (and even correct as outlined in #845) with invalid functions.

This commit adds a `--force` cli option (discussion welcomed!) that allows users to opt to ignore invalid functions (and skip more checks in the future?).

Closes #845

### DIFF
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
@@ -14,7 +14,7 @@ import { isUsingAppRouter, isUsingPagesRouter } from '../getVercelConfig';
 type InvalidFunctionsOpts = Pick<
 	ProcessVercelFunctionsOpts,
 	'functionsDir' | 'vercelConfig'
->;
+> & { ignoreInvalidFunctions?: boolean };
 
 /**
  * Checks if there are any invalid functions from the Vercel build output.
@@ -46,7 +46,10 @@ export async function checkInvalidFunctions(
 	await tryToFixInvalidFuncsWithValidIndexAlternative(collectedFunctions);
 	await tryToFixInvalidDynamicISRFuncs(collectedFunctions);
 
-	if (collectedFunctions.invalidFunctions.size > 0) {
+	if (
+		collectedFunctions.invalidFunctions.size > 0 &&
+		!opts.ignoreInvalidFunctions
+	) {
 		await printInvalidFunctionsErrorMessage(
 			collectedFunctions.invalidFunctions,
 		);
--- a/.changeset/hot-forks-try.md
+++ b/.changeset/hot-forks-try.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': minor
+---
+
+Add `force` CLI option to ignore invalid functions
--- a/packages/next-on-pages/src/buildApplication/buildApplication.ts
+++ b/packages/next-on-pages/src/buildApplication/buildApplication.ts
@@ -30,6 +30,7 @@ export async function buildApplication({
 	watch,
 	outdir: outputDir,
 	customEntrypoint,
+	force,
 }: Pick<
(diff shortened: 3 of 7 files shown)


packages/next-on-pages/src/buildApplication/processVercelFunctions/invalidFunctions.ts
```diff
@@ -46,7 +46,10 @@ export async function checkInvalidFunctions(
 	await tryToFixInvalidFuncsWithValidIndexAlternative(collectedFunctions);
 	await tryToFixInvalidDynamicISRFuncs(collectedFunctions);
 
-	if (collectedFunctions.invalidFunctions.size > 0) {
+	if (
+		collectedFunctions.invalidFunctions.size > 0 &&
+		!opts.ignoreInvalidFunctions
```
james-elicx, line 51: my concern with this approach is that people may find and use `--force` because they get this error, without actually accounting for the routes properly... In your case, you want to host these routes elsewhere, but I imagine there'll be people who don't realise they need to actually specify these routes should execute somewhere else, and expect it to just work by _forcing_ a build.

nickbabcock: Same. I don't enjoy the thought of forcing a build.

I worry that the alternative of trying to better evaluate the next.config to determine if the function is actually invalid may be fraught with difficulty (I've not looked too closely, so I'll defer to your judgement). Is this what you had in mind?

Would printing a warning instead of exiting be sufficient? So the errors aren't swept under the rug.

From a connotation perspective, I find `force` to be ok, but I'm happy to rename the argument as well

dario-piotrowicz:
Thank you very much for the PR (and sorry for the late reply here 🙇). 

Earlier this year we released our [Cloudflare adapter for OpenNext](https://opennext.js.org/cloudflare), and we have been continuously investing effort in its improvement since then. The Cloudflare adapter for OpenNext is currently stable and recommended for deployment of [Next.js](http://next.js/) applications to Cloudflare. As such, we will no longer be maintaining next-on-pages. We recently deprecated the package on npm and we plan to archive this repo on Monday Sep 29, 2025. In preparation for archival, we are closing all open issues and PRs. 

If you have concerns, please feel free to reply here before the repo is archived (after that, it will no longer be possible to comment) or to reach out to our team on [Discord](https://discord.com/invite/cloudflaredev).
 
Thank you so much for using and contributing to next-on-pages 🧡

PR #886, opened 2024-10-01
fix: webpack chunks not deduping with sentry

### LINKED ISSUE #860: [🐛 Bug]: Sentry bloats up bundle size when building with npx @cloudflare/next-on-pages

System:
        Platform: darwin
        CPU: arm64 Apple M3 Pro
        Memory: 18 GB
        Shell: /bin/zsh

Relevant Packages:
        @cloudflare/next-on-pages: 1.13.2
        next: 14.2.5
        @sentry/nextjs: 8.26.0

### Description

When installing Sentry in any Next.js project, it massively bloats the bundle size, going way beyond the limit of 25MB for normal projects. For example, in my project, it increases the bundle size almost by a factor of 10, reaching 70MB. This, of course, is unreasonable, and I suspect that it has something to do with the deduplication that Cloudflare performs during the build, but I'm not really sure.

I think this issue is somehow related to [this GitHub issue](https://github.com/cloudflare/next-on-pages/issues/678), which is caused by the vercel/otel package.

Overall, it prevents the use of Cloudflare Pages for Next.js in production at some scale, as there seems to be no working option to use an observability tool. We are overall really satisfied with hosting on Cloudflare, but this might force us to switch soon.

### Reproduction

Steps to reproduce:

1. Clone a next.js project (I will use https://github.com/vercel/ai-chatbot as example)
2. Create a build using `npx @cloudflare/next-on-pages`
3. Check the bundle size in the folder `.vercel/output/static/_worker.js` (for the above example I arrive at 4.1 MB)
4. Install Sentry and jump through the setup (`npx @sentry/wizard@latest -i nextjs`)
5. Build and check the bundle size again (for the above example I arrive at 15.8 MB)

### Pages Deployment Method

None

### Pages Deployment ID

_No response_

### Additional Information

_No response_

### Would you like to help?

james-elicx:
When using sentry, you get a SequenceExpression with the CallExpression inside it. We need to find that CallExpression so we can use it for deduping.

fixes #860

image

### DIFF
--- a/.changeset/nasty-kids-tickle.md
+++ b/.changeset/nasty-kids-tickle.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Fix the Webpack chunk deduplication when Sentry is used, as it changes the AST node structure for Webpack chunks.
--- a/packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
+++ b/packages/next-on-pages/src/buildApplication/processVercelFunctions/ast.ts
@@ -269,32 +269,39 @@ function extractManifest(
 function extractWebpack(
 	statement: AST.StatementKind,
 ): RawIdentifier<'webpack'>[] {
+	if (statement.type !== 'ExpressionStatement') return [];
+
+	const expr =
+		statement.expression.type === 'SequenceExpression'
+			? statement.expression.expressions.find(
+					expr => expr.type === 'CallExpression',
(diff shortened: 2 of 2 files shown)


dario-piotrowicz (approved):
The change looks good to me, thanks a lot @james-elicx 🫶 

Ideally I think the code should be more robust and deal with all the `SequenceExpression`'s `expressions`, but that doesn't need to be a blocker for this PR

Issue #894, opened 2024-10-17
[🐛 Bug]: Can't setup a websocket in an endpoint


sassanh:
### next-on-pages environment related information

```
System:
        Platform: darwin
        Arch: arm64
        Version: Darwin Kernel Version 24.1.0: Mon Sep 30 00:10:31 PDT 2024; root:xnu-11215.40.63~39/RELEASE_ARM64_T6000
        CPU: (10) arm64 Apple M1 Pro
        Memory: 32 GB
        Shell: /bin/bash
Package Manager Used: npm (10.8.2)

Relevant Packages:
        @cloudflare/next-on-pages: 1.13.5
        vercel: N/A
        next: 14.2.15
```

### Description

I can't setup websocket with the `@cloudflare/next-on-pages` after version `0.10.0`

@dario-piotrowicz kindly provided this sample repository for running websocket inside next-on-pages projects:
https://github.com/dario-piotrowicz/next-app-websocket-pages.git

It was provided [here](https://github.com/cloudflare/next-on-pages/issues/131)

But it doesn't work with next-on-pages 0.10.0 or later locally and it doesn't work at all when deployed.

### Reproduction

1. Clone this repository: https://github.com/dario-piotrowicz/next-app-websocket-pages.git
1. Change dependencies in package.json like this:

    ```json
      "dependencies": {
        "@types/node": "22.7.6",
        "@types/react": "18.3.11",
        "@types/react-dom": "18.3.1",
        "next": "14.0.0",
        "react": "18.3.1",
        "react-dom": "18.3.1",
        "typescript": "5.6.3"
      },
      "devDependencies": {
        "@cloudflare/workers-types": "^4.20241011.0",
        "@cloudflare/next-on-pages": "==0.10.0",
        "vercel": "^28.0.0"
      }
    ```

1. Run `npx @cloudflare/next-on-pages`
1. Run `wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat`
1. Try to connect to the websocket, for example with `websocat ws://localhost:8788/api/socket`
1. Now try the same with `@cloudflare/next-on-pages==0.9.0` and it works.
1. Try deploying both and none of them works when deployed with error:

```
{
"stack": " at ke (bundledWorker-0.8185137090843237.mjs:702:50)",
"name": "RangeError",
"message": "Responses may only be constructed with status codes in the range 200 to 599, inclusive.",
"timestamp": 1729141252885
}
```

### Pages Deployment Method

None

[...]

PR #902, opened 2024-11-02
avoid re-instantiating `Response` when not required

sassanh:
The problem is described here: https://github.com/cloudflare/next-on-pages/issues/894
You may ignore the rest and just read the last comment

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/utils/routing.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/routing.ts
@@ -131,6 +131,9 @@ export async function runOrFetchBuildOutputItem(
 		return new Response('Internal Server Error', { status: 500 });
 	}
 
+	if (resp.status === 101) {
+		return resp;
+	}
 	return createMutableResponse(resp);
 }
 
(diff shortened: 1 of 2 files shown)


packages/next-on-pages/templates/_worker.js/utils/routing.ts
```diff
@@ -131,7 +131,7 @@ export async function runOrFetchBuildOutputItem(
 		return new Response('Internal Server Error', { status: 500 });
 	}
 
-	return createMutableResponse(resp);
+	return resp;
```
sassanh, line 134: `generateResponse` uses this function, but it already does what `createMutableResponse` does when needed.

It is also used here `packages/next-on-pages/templates/_worker.js/routes-matcher.ts:239`, but since my app is not using middlewares I couldn't test it, `npm run test:unit` passes though.
If it is needed there, maybe we can do `createMutableResponse` locally for it?

Hopefully my changes are not fundamentally against your design, I just went with common sense based on my very limited knowledge on the codebase.

sassanh: I changed it to
```typescript
	if (resp.status === 101) {
		return resp;
	}
	return createMutableResponse(resp);
```

again this is an attempt to allow websocket connections without knowing much about the codebase, I don't know if it is only status code 101 we should put this exception for, or if it is the correct signal for the exception.

PR #940, opened 2025-02-19
Extract module name

penalosa:
Newer versions of esbuild try to resolve dynamic imports by [crawling the filesystem for matching modules](https://esbuild.github.io/api/#glob). This doesn't work with the code `next-on-pages` generates, because the modules don't necessarily exist on the filesystem.

This PR extracts a `moduleName` variable so that esbuild doesn't try to look for matching modules on the filesystem.


dario-piotrowicz (approved):
Looks good to me 🙂

PR #946, opened 2025-02-27
_Really_ don't inline dynamic imports

penalosa:
Take 2 of https://github.com/cloudflare/next-on-pages/pull/940, which was still inlining dynamic imports due to esbuild's minification algorithm removing the intermediate variables.

### DIFF
--- a/packages/next-on-pages/templates/_worker.js/utils/cache.ts
+++ b/packages/next-on-pages/templates/_worker.js/utils/cache.ts
@@ -1,5 +1,6 @@
 import type { CacheAdaptor, IncrementalCacheValue } from '../../cache';
 import { SUSPENSE_CACHE_URL } from '../../cache';
+import { doImport } from './doImport';
 
 // https://github.com/vercel/next.js/blob/48a566bc/packages/next/src/server/lib/incremental-cache/fetch-cache.ts#L19
 const CACHE_TAGS_HEADER = 'x-vercel-cache-tags';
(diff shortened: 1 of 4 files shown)


packages/next-on-pages/templates/_worker.js/utils/cache.ts
```diff
@@ -108,6 +108,10 @@ export async function getSuspenseCacheAdaptor(): Promise<CacheAdaptor> {
 	return getInternalCacheAdaptor('cache-api');
 }
 
+async function doImport(m: string) {
```
dario-piotrowicz, line 111: I'd add a comment clarifying what this function is for

packages/next-on-pages/templates/_worker.js/utils/fetch.ts
```diff
@@ -32,6 +32,10 @@ function applyPatch() {
 	};
 }
 
+async function doImport(m: string) {
+	return import(m);
+}
```
dario-piotrowicz, line 37: I'd probably have a single `doImport` function (maybe in its own file under `utils/`) so that we can also add comments to it without repetition, etc..

dario-piotrowicz (approved):
Makes sense to me 🙂 , I just left a few nits/suggestions

dario-piotrowicz (approved):
Looks great, thanks for the updates @penalosa 🫶

PR #963, opened 2025-04-01
fix internal usage of `global` instead of `globalThis`

dario-piotrowicz:


james-elicx (approved):

dario-piotrowicz:
thanks @james-elicx 🫶

PR #983, opened 2025-08-29
Ignore invalid favicon.ico.rsc nodejs functions being generated by the Vercel CLI (since 46.1.0)

dario-piotrowicz:


james-elicx (approved):
code looks fine

jamesopstad (approved):

.changeset/twelve-avocados-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/next-on-pages': patch
+---
+
+Ignore invalid favicon.ico.rsc nodejs functions being generated by the Vercel CLI (since 46.1.0)
```
vicb, line 5: nit: do we need to say that those are "invalid"? 

I feel like if we add "invalid" in this PR we should explain why they are invalid (I'm interested to learn more either way).

If we just remove "invalid" then we don't need to explain why and just say that we ignore those functions.

vicb: Oh whatever this has been merged already.

Still interested in why those "functions" are invalid if you see this message...

james-elicx: We use invalid to refer to functions that do not expect to receive and are not able to use, so in this example, a Node.js function would be considered invalid by our CLI as we unable to use them and do not expect them to be in the output. However, there are some cases though where we can later figure out that these functions, while technically invalid, are safe for us to ignore, so we move them from being considered invalid to being ignored, e.g. for cases like this where the Vercel CLI has generated a function that we don't need to use, or one that we replace with a prerendered file instead.

